Issue #3, opened 2016-01-14
Create backend Dockerfile


kardan:
https://github.com/akvo/akvo-provisioning/issues/246

kardan:
While improving the docs in the backend/README.md I found out that Django management commands don't work so doing things like creating a new Django app in the project does not work. Commands like:
```shell
docker-compose run web python manage.py start app users
```
I was expecting to work but they just return without any feedback.

PR #5, opened 2016-01-18
#3 backend dockerfile

kardan:
Created a docker-compose setup with some initial docs in backend/README.md.


backend/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+db:
+  image: postgres:9.5
+web:
+  build: .
+  command: python manage.py runserver 0.0.0.0:8000
+  volumes:
+    - .:/opt/akvo-dash
```
orifito: maybe I'm missing something... but why do we need the current host path mounted at /opt/akvo-dash?

kardan: My thinking was that we need to have access to the Python project code while it's being developed. Would we not need to build a new container on each code change otherwise?

orifito: Yeah this is a very common approach to have a truly development environment. Let's standardise "/app" :smiley:

orifito:
Love the use of env var files :clap: :smile:

Issue #7, opened 2016-01-23
Build high level component structure


jonase:
We will attempt to follow a fairly idiomatic redux + react approach when building the client app. My current understanding in how to achieve this is as follows:

* `Component` -- Plain react component, often stateless, unaware of redux.
* `Container` -- React component that listens for changes in the redux store. Selects the data it needs from the store (via `connect`). Binds action creator functions to `dispatch` and sends data and the bound action creators to child components as props.
* There is a known fundamental trade-off between the simplicity of declarative stateless react components and the management and passing of props and callbacks down the render tree. It is possible to have nested containers but this should be avoided until the props management becomes too cumbersome.

The component structure attempts to follow the layout overview by @gabemart:

```
/* Routes */

<Route path="/" component={App}>
	<Route component={Library}>
		<Route component={LibraryAll} />
		<Route component={LibraryDatasets} />
		<Route component={LibraryVisualisations} />
		<Route component={LibraryDashboards} />
	</Route>
	<Route path="Activity">
	</Route>
	<Route path="Admin">
	</Route>
</Route>

/* "Static" (i.e. non-modal) components */

<App>
	<Nav>
		<WorkspaceNav>
			<NavLink to="Library" collapsed="false"/>
			<OrganisationMenu />
			<CollectionsList>
				<AddCollection>
				<CollectionsListItem />
				<CollectionsListItem />
				<CollectionsListItem />
				<CollectionsListItem />
			</CollectionsList>
			<NavLink to="Activity" />
			<NavWorkspaceSwitch />
		</WorkspaceNav>
		/* Or, if we're in the admin section: */
		<AdminNav />
	</Nav>
	{this.props.children}
</App>

<Library>
	<LibraryHeader>
		<LocationIndicator>
			/* By default, just displays "Library".
			**
			** If we are in an edit view, displays the name of the entity we
			** are editing and some limited related info (e.g. "last saveed {x} mins ago").
			**
			** When collections are implemented, will show the name of the collection we
			** are currently *in* if we are viewing one.
			*/
		</LocationIndocator>
		<LibraryCreateButton>
			/* Displays a menu to create a new dataset, visualisation or dashboard
			** and routes us to the correct child-page of the Library.
			*/
		</LibraryCreatebutton>
		<LibrarySearch />
		<LibraryTabList>
			<LibraryHeaderTab tab="All" />
			<LibraryHeaderTab tab="Datasets" />
			<LibraryHeaderTab tab="Visualisations" />
[...]

PR #14, opened 2016-01-29
Add LibraryListing functionality

gabemart:
#7


client/src/components/library/LibraryDisplayMenu.jsx
```diff
@@ -26,6 +33,8 @@ export default class LibraryDisplayMenu extends Component {
 LibraryDisplayMenu.propTypes = {
   sortOrder: PropTypes.oneOf(['last_modified', 'created', 'name']).isRequired,
   onChangeSortOrder: PropTypes.func,
+  reverseSort: PropTypes.bool.isRequired,
+  onChangeReverseSort: PropTypes.func,
```
jonase: I suppose we should add `isRequired` to the `func` props as well?

client/src/components/library/LibraryListing.jsx
```diff
@@ -1,7 +1,143 @@
...
+      let dateA = a.created;
+      let dateB = b.created;
+
+      if (this.props.sortOrder === 'last_modified') {
+        dateA = a.modified || dateA;
+        dateB = b.modified || dateB;
+      }
+
+      if (this.props.reverseSort) {
+        output = new Date(a.created) - new Date(b.created);
```
jonase: Should we be using `dateA` & `dateB` here?

client/src/components/library/LibraryListingGroup.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default class LibraryListingGroup extends Component {
+  render() {
+    const listGroupDate = new Date(Date.parse(this.props.listGroup.listGroupDate));
+    const listGroupTitle = listGroupDate.toUTCString(); // TODO: format this
+
+    return (
+      <div className="LibraryListingGroup">
+        {this.props.isSortDateType &&
+          <h3>{listGroupTitle}</h3>
```
jonase: If sorted & grouped by name, do we want to show `A` for the "a" grouping etc.?

jonase:
I agree we could try to refactor the `LibraryListing.render()` method. Would it be possible to create three functions:

* filterEntities(library, filterBy, searchString),
* sortEntities(entities, sortOrder, isReversed),
* groupEntities(entities,  sortOrder)

and in the `LibraryListing.render` basically only do

```javascript
const entities = filterEntities(library, filterBy, searchString);
const sortedEntities = sortEntities(entities, sortOrder, isReversed);
const groupedEntities = groupEntities(sortedEntities, sortOrder);

return (
  <div className={`LibraryListing ${this.props.displayMode}`}>
    <ul>
      ...
);
```

All of this assumes that we have fetched all the entities from the server which is fine for now but something we probably have to revisit later.

jonase:
Another thing I thought about. It's not really the job of LibraryListing to `sort` & `filter` the entities. This could be done in the `LibraryListingGroup` as well.

* `LibraryListing` groups the entities
* `LibraryListingGroup` sorts & filters the entities

This (or similar) approach could perhaps simplify the implementation?

gabemart:
>Another thing I thought about. It's not really the job of LibraryListing to sort the entities. This could be done in the LibraryListingGroup as well.

I agree it makes sense for each `LibraryListingGroup` to sort it's own internal entities. However, I think the LibraryListing will still have to sort the collection of `LibraryListingGroups` itself

jonase:
:+1:

PR #16, opened 2016-01-31
[#7] Create collection modal

jonase:
@gabemart see also https://github.com/akvo/akvo-dash/issues/15


gabemart:
@jonase Are we sure it makes sense to implement the `CreateCollectionModal` as dependent on the component state of `WorkspaceNav`? 

The mockups show a couple of other options for triggering this modal, e.g. in the `LibraryHeader` and `LibraryListing`. If the display of the `CreateCollectionModal` was dependent on root state rather than component state, we might be able to add this functionality very cheaply?

gabemart:
Also, we might want to think about creating a custom `customModal` component that encapsulates the third party 'Modal' component we are using and implements some standard elements - it seems likely that every modal we create will have at least some common parts, like a close button, cancel button, title, etc. Using a custom wrapper for the third-party component would let us define and style these common elements in one  place.

I think it might be premature to do this now as it's not a priority, but perhaps it's something we can think about.

jonase:
@gabemart I think I will move both the create-dataset and create-collection modal higher up in the render tree and handle the show/hide via redux actions. Something like `{ action: 'SHOW_MODAL', modal: 'create-dataset' }` and `{ action: 'HIDE_MODAL' }`. This will probably mean that we'll drop the `url` for creating a dataset which is fine as it's causing some problems already.. and I don't think modals need to be backed by urls anyway.

gabemart:
code checked :+1:

Issue #20, opened 2016-02-04
Setup backend database


kardan:
- [x] Create simple provisioning script to get a db
- [x] Make sure migrations work properly
- [x] Create a way to clean up
- [x] Update README to cover db

PR #23, opened 2016-02-05
#20 db setup

kardan:
One snag that I struggled with a bit is that for the migrations to work out of the box the system needs to be running. That since the migration happens in a component.


backend/provision/setup-up.sh
```diff
@@ -0,0 +1,34 @@
...
+CLI_ERR_MSG="Postgres CLI tools not available (psql). Using Postgres.app, look at http://postgresapp.com/documentation/cli-tools.html. Aborting."
+hash psql 2>/dev/null || { echo >&2 $CLI_ERR_MSG ; exit 1; }
+
+# Postgres
+psql -c "CREATE ROLE dash WITH PASSWORD 'password' CREATEDB LOGIN;"
+
+psql -c "
+CREATE DATABASE dash
+OWNER dash
+ENCODING 'UTF8' LC_COLLATE = 'en_US.UTF-8' LC_CTYPE = 'en_US.UTF-8'
```
iperdomo: Let's inherit `encoding` and other parameters from `template0`

iperdomo:
:+1:

Issue #33, opened 2016-02-11
HTTP api


jonase:
This issue will try to specify the HTTP api interface between the client dashboard and the backend.

# TOC
* [datasets](#datasets)
* [library](#library)
* [visualisations](#visualisations)

# <a name="datasets"></a> /datasets

### Create a new dataset from a data source 

```
POST /api/datasets
{
  "name": "My Dataset"
  "source" {
    "kind": "LINK"
    "url": "http://..."
  }
}

returns:

{
  "id": "de305d54-75b4-431b-adb2-eb6b9e546014"
  "status": "PENDING"
  "name": "My Dataset"
}
```

Since the dataset is fetched & prepared asynchronously on the backend the dataset will not be immediately available, and therefor in a `PENDING` state.

### Get a single dataset

```
GET /api/datasets/:id

Returns one of
{
  "id": "de305d54-75b4-431b-adb2-eb6b9e546014"
  "status": "PENDING"
  "name": "My Dataset"
}

{
  "id": "de305d54-75b4-431b-adb2-eb6b9e546014"
  "status": "FAILURE"
  "reason": "Could not find dataset on remote server (404)"
  "name": "My Dataset"
}

{
  "id": "de305d54-75b4-431b-adb2-eb6b9e546014"
  "status": "OK"
  "name": "My Dataset"
  "columns": [
    {
      "title": "Column 1",
      "type": "NUMBER",
      "values": [1, 2, 3]
    },
    {
      "title": "Column 2",
      "type": "STRING",
      "values" ["foo", "bar", "baz"]
    },
    {
      "title": "Column 3",
      "type": "DATE",
      "format": "dd-mm-yyyy",
      "values": [1454036337, 1454036370, 1454036402]
    }
  ],
}
```

### Get data source information

Once the dataset has been imported the data source is seldom needed. You can still get the data source definition with

``` 
GET /datasets/:id/source
{
  "kind": "LINK"
  "url": "http://..."
}
```

###  Get imports 
List only pending by default?
``` 
GET /datasets/:id/imports
[{
  "id": "..",
  "status: "OK"
 } ,{
  "id": "..",
  "status: "PENDING"
}]
```

### Cancel pending dataset
```
// TODO
```

### Delete dataset
```
DELETE /datasets/:id

Set the dataset to disabled and return one of alts
{
  "id": "de305d54-75b4-431b-adb2-eb6b9e546014",
  "status":  "DELETED",
  "enabled": false
}

{
[...]

PR #43, opened 2016-02-19
#33 library endpoint

kardan:
- TravisCI
- Postgres (jsonb / timestamps in unix format)
- 404 check on datasets/:id


jonase:
:+1:

Issue #47, opened 2016-02-23
Extend dataset endpoint with PUT  & DELETE


kardan:
Follow how visualisations work.

kardan:
Since DELETE is tricky*, that's not implemented.

*It's tricky since visualisations & in the longer term dashboards depend on datasets. See a comment here: https://github.com/akvo/akvo-dash/blob/fb27d7787ff9e34c08735d7f5fbdb3f3b75060b6/backend/src/org/akvo/dash/endpoint/dataset.clj#L116

kardan:
Closing this since I created https://github.com/akvo/akvo-dash/issues/69

PR #55, opened 2016-02-26
[#47] Dataset can now be updated.

kardan:


jonase:
:+1:  Looks good. We can investigate the json middleware after this initial "release"

Issue #63, opened 2016-02-29
Add styles based on preliminary visual designs 


gabemart:
Apply some basic visual styling beyond the "wireframe" styling that currently exists. 

The styles here are based on preliminary visual designs, and will change substantially over time.

PR #64, opened 2016-02-29
[#63] Apply initial visual styles

gabemart:
#63


client/src/components/WorkspaceNav.jsx
```diff
@@ -23,32 +23,75 @@ const getCollapsedStatus = (pathname) => {
...
     };
   }
 
-  getOnClick(isCollapsible) {
-    let onClick = null;
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.location.pathname !== this.props.location.pathname) {
+      this.setState({
+        isManuallyExpanded: false,
+        isManuallyCollapsed: false,
```
jonase: why do we need both `isManuallyExpanded` and `isManuallyCollapsed`? Are they not mutually exclusive?

client/index.html
```diff
@@ -8,5 +8,6 @@
 	<body>
 		<div id="root"></div>
 		<script type="text/javascript" src="/bundle.js"></script>
+    <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,700' rel='stylesheet' type='text/css'>
```
jonase: there seems to be some mixed tabs/spaces here. Perhaps move the whole file to spaces?

client/src/components/library/LibraryCreateButton.jsx
```diff
@@ -1,18 +1,30 @@
...
+    const options = [
+      { value: 'dataset', label: 'Dataset' },
+      { value: 'visualisation', label: 'Visualisation' },
+      { value: 'dashboard', label: 'Dashboard' },
+      { value: 'collection', label: 'Collection' },
+    ];
+    const onChange = (value) => {
+      onCreate(value);
+      // Force a rerender so menu still shows placeholder
+      this.forceUpdate();
```
jonase: note-to-self/investigate @jonase: It seems that we need to `forceUpdate()` because react-select doesn't do controlled components.

client/src/components/visualisation/configMenu/TwoAxisConfigMenu.jsx
```diff
@@ -13,6 +14,32 @@ const getDatasetArray = datasetObject => {
...
+const getDatasetOptions = (datasetArray) => {
+  const output = [];
+
+  datasetArray.forEach(option => {
+    output.push({
+      value: option.id, label: option.name,
+    });
+  });
+
+  return output;
```
jonase: would it be simpler to do

```javascript
return datasetArray.map(option => ({
  value: option.id,
  label: option.name,
}));
```

client/src/components/visualisation/configMenu/TwoAxisConfigMenu.jsx
```diff
@@ -13,6 +14,32 @@ const getDatasetArray = datasetObject => {
...
+
+  if (array) {
+    array.forEach((entry, index) => {
+      output.push({
+        value: index, label: entry.title,
+      });
+    });
+  }
+
+  return output;
```
jonase: same here. Perhaps consider `map` instead of `forEach+push`?

client/src/components/workspace-nav/CollectionListItem.jsx
```diff
@@ -18,4 +24,5 @@ export default class CollectionListItem extends Component {
 
 CollectionListItem.propTypes = {
   collection: PropTypes.object,
+  pathname: PropTypes.string,
```
jonase: This is more of a general comment which we can look out for in the future. Above we call  `this.props.pathname.indexOf(...)` without checking if `pathname` is `null/undefined`. This is fine, but should we in that case require that `pathname` is a string: `PropTypes.string.isRequired`.

gabemart:
Thank you for your feedback @jonase - I agree in each case, and have made the suggested changes except where noted

client/src/components/visualisation/configMenu/TwoAxisConfigMenu.jsx
gabemart: I agree with this - but these config menus need to be refactored in any case to extract some common code. When I refactor, I will use `map` as you suggest

jonase:
:+1:

PR #73, opened 2016-03-17
Update deps

jonase:


gabemart:
Code checked :+1:

Didn't find any problems. My only comment would be, should we consider declaring explicit versions of our lint presets dependencies, to prevent them being auto-updated in the build process?

There are both pros and cons to this approach. If we neglect manually updating the version numbers, we risk some very unpleasant refactoring when we eventually get around to it. On the other hand, it would prevent the build process failing at unpredictable intervals.

What are your thoughts?

jonase:
@gabemart I've been thinking about the same thing.. We could also pin *all* dependencies and only do manual updates. This is what I'm used to from Java land but it seems like people put a lot of trust in [semver](http://semver.org/) in JS.

gabemart:
@jonase and I have agreed to leave dependencies as-is for now

Issue #58, opened 2016-02-28
Keep package.json dependencies up-to-date


jonase:
To the best of our ability we should try to keep up with the rapidly changing Javascript framework and library dependencies. The best way to achieve this is to regularly check for outdated packages and set aside some time to resolve breaking changes. A good approach has been to run [ncu](https://github.com/tjunnone/npm-check-updates) about once a week.

jonase:
with the 2016-06-30 milestone we should try to update to React v.15

PR #77, opened 2016-03-23
[#58] Update client deps

jonase:
* Remove custom no-jsx-bind rule as arrow functions are now allowed in render function


gabemart:
Code checked :+1:

Issue #83, opened 2016-04-05
Integrate Keycloak at backend


kardan:
# Context
At the moment we only authorise users on the client using Keycloak. However the client pass the token to the backend but the backed does not no anything about tokens. Hence we can't validate that the token is a valid token.

# Problem or idea
The backend needs to know about JWT tokens. 

# Solution or next step
Integrate the JWT middleware from akvo-commons. Then the backend should be able to integrate with Keycloak and validate tokens.

kardan:
Work done in https://github.com/akvo/akvo-dash/issues/74 (https://github.com/akvo/akvo-dash/blob/c341877d78ac4ead1136854683316dd250595222/backend/src/org/akvo/dash/middleware.clj)

kardan:
Seems we can pass additional headers in options https://github.com/tus/tus-js-client#tusdefaultoptions

PR #93, opened 2016-04-19
[#83] Enforce valid JWT token on /files upload

kardan:
Before this resumable uploads was tunnelled on the auth middleware. Now
the client sends the token and we validate requests.

### DIFF
--- a/backend/test/org/akvo/dash/middleware_test.clj
+++ b/backend/test/org/akvo/dash/middleware_test.clj
@@ -1,5 +1,6 @@
 (ns org.akvo.dash.middleware-test
   (:require [org.akvo.dash.middleware :as m]
+            [org.akvo.dash.system :as system]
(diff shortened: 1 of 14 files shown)


backend/test/org/akvo/dash/middleware_test.clj
```diff
@@ -77,3 +78,27 @@
...
+
+
+    (testing "Invalid token should not yeild jwt-claims."
+      (let [response (jwt-middleware (assoc-in (immutant-request :get "/")
+                                               [:headers "authorization"]
+                                               "invalid-token"
+                                               ))]
+        (is (not (contains? response :jwt-claims)))))
+
+    #_(testing "Valid token should yeild jwt-claims."
```
iperdomo: Let's remove commented code?

kardan: Removed

PR #99, opened 2016-04-21
Feature/84 85 import

jonase:


backend/src/org/akvo/dash/import.clj
```diff
@@ -1,74 +1,75 @@
...
-  {:type (last (str/split url #"\."))
-   :data (:body (client/get url))})
-
-(defmethod content "DATA_FILE" [{{:strs [fileName url]} :spec}]
-  {:type (last (str/split fileName #"\."))
-   :data (slurp (str "/tmp/akvo/dash/resumed/"
-                     (last (str/split url #"\/"))
-                     "/file"))})
-
+         (do
```
kardan: We only do one fn call, remove do?

backend/src/org/akvo/dash/import/csv.clj
```diff
@@ -58,11 +58,21 @@
...
+  (format "DROP TABLE IF EXISTS %s CASCADE" table-name))
+
 (defn get-headers
   "Returns the first line CSV a file"
   [path separator]
   (with-open [r (io/reader path)]
-    (first (csv/read-csv r :separator separator))))
+    (vec (first (csv/read-csv r :separator separator)))))
 
 (defn get-num-cols
```
kardan: Can we trust the CSV?

Issue #80, opened 2016-03-30
Enhancements on CSV file upload screen


iperdomo:
- [x] Include a progress bar of the upload process
- [ ] Include a section with the CSV settings
  - [ ] File encoding
  - [ ] Skip [n] lines
  - [ ] Text qualifier / Text delimiter

jonase:
@iperdomo 

* With https://github.com/akvo/akvo-dash/issues/81 the user wouldn't need to specify the encoding since we try to solve it by using apache tika
* Should we add something like "Treat first row as column names"? If not we might want to add a transformation like "Extract first row as column names" and change the current default behaviour.
* Right now we have `Data file` and `Link` data sources. Should we unify these two into a single "CSV" data source where you either upload a file or provide a link. One problem with this would be that a link can potentially support periodic updates while a file upload can't. 
* If we treat Link and Upload as two different data sources perhaps the user needs to specify file type (only csv for now) and only show csv specific options when the file type has been selected. Or should we have more specific data sources like "CSV Link" and "CSV Upload"?

iperdomo:
> With #81 the user wouldn't need to specify the encoding since we try to solve it by using apache tika

Agree. I think we should detect it

> Should we add something like "Treat first row as column names"? If not we might want to add a transformation like "Extract first row as column names" and change the current default behaviour.

Let's agree on this based on @Kiarii input

> Right now we have Data file and Link data sources. Should we unify these two into a single "CSV" data source where you either upload a file or provide a link. One problem with this would be that a link can potentially support periodic updates while a file upload can't.

I think we want to keep them separate but _re-label_ the titles to "CSV [...]"

> If we treat Link and Upload as two different data sources perhaps the user needs to specify file type (only csv for now) and only show csv specific options when the file type has been selected. Or should we have more specific data sources like "CSV Link" and "CSV Upload"?

For now we should restrict to only CSV. In the case of the `file` input, we can use the `accept` attribute. https://www.w3.org/TR/html-markup/input.file.html#input.file.attrs.accept

In the case of the link we need to download the file and check the mime type with Tika, if is `text/csv` we process it.

iperdomo:
I made a sample test downloading a CSV file from http://api.worldbank.org/v2/en/indicator/BN.CAB.XOKA.CD?downloadformat=csv and using Tika, detect the mime type.

````
(def tika (Tika.))                                                                                                                                              
[...]

PR #103, opened 2016-04-25
[#80] Add upload progress bar for file upload

gabemart:
#80


jonase:
@gabemart the progress bar is really nice! I wonder if we should leave it visible at 100% when the upload is finished instead of hide it?

gabemart:
@jonase I have tested out your suggestion and I like it 👍

client/src/components/modals/createDataset/DataFileDataSourceSettings.jsx
```diff
@@ -35,17 +42,23 @@ export default class DataFileDataSourceSettings extends Component {
...
+
   uploadFile(file) {
     const onChange = this.props.onChange;
+    const handleProgress = this.handleProgress;
     const upload = new tus.Upload(file, {
       headers: { Authorization: `Bearer ${keycloak.token}` },
       endpoint: '/api/files',
       onError(error) {
         console.error(`Failed because: ${error}`);
+        handleProgress(null);
```
jonase: In `handleProgress` we `parseFloat` the `percentage` argument and `parseFloat(null)` evaluates to `NaN` which I don't think is desired here, especially since we explicitly check for `null` in `getProgressBarBool`.

client/src/components/modals/createDataset/DataFileDataSourceSettings.jsx
```diff
@@ -17,6 +18,12 @@ export default class DataFileDataSourceSettings extends Component {
     this.handleDragEnter = this.handleDragEnter.bind(this);
     this.handleDragOver = this.handleDragOver.bind(this);
     this.handleDrop = this.handleDrop.bind(this);
+    this.handleProgress = this.handleProgress.bind(this);
+    this.state = { uploadProgressPercentage: null };
+  }
+
+  getShowProgressBarBool() {
```
jonase: I would suggest renaming this method to remove the type suffix. Perhaps simply `showProgressBar`? or `isProgressBarVisible`? I don't think we do much data type suffix/prefix based naming currently and I like it that way.

gabemart:
@jonase I agree with both your comments and have made some changes.

* I renamed the function to `isProgressBarVisible()` as suggested

* After revisiting error-handling to correct the problem with parsing null, I added some optional text props to the progress bar component so it will display to the user if there is an error and show a "success" message if the upload completes

jonase:
👍

Issue #78, opened 2016-03-29
Review the testing story


kardan:
At the moment we have a few tests, we need to review how to deal with tests. We need to make sure we are not afraid to move fast!

This requires to both write unit tests, and also functional tests. To enable functional tests we need to:
- [ ] Enable test using db locally
- [ ] Enable db at CI server

Review if we can add lint / version tools to CI.

zzgvh:
I'm reopening since there's more travis-db stuff to be done. Created a new branch though.

iperdomo:
@zzgvh the idea is to have smaller issues with concrete descriptions, instead of a "container" of stuff.

PR #105, opened 2016-04-27
#78 testing the db

zzgvh:
There's now a working set of migrations, up and down, and the first of the tests in functional_test.clj works. The others depend on keycloak and tenant subdomains if I understand things right, so I'm leaving them commented out for now.


backend/test/org/akvo/dash/fixtures.clj
```diff
@@ -1,21 +1,43 @@
...
-  [f]
-  (f))
+(def dev-config
+  {:app {:middleware [wrap-stacktrace]}})
+
+(def config
+  (meta-merge config/defaults
+              config/environ
+              dev-config))
+;(defn system-fixture
```
kardan: This can be removed.

backend/test/org/akvo/dash/fixtures.clj
```diff
@@ -1,21 +1,43 @@
 (ns org.akvo.dash.fixtures
```
kardan: Seems like the requires can be cleaned up a bit.We import stuff we don't use.

backend/test/org/akvo/dash/fixtures.clj
```diff
@@ -1,21 +1,43 @@
...
-;;     (-> system :ragtime ragtime/reload (ragtime/rollback "001-activity"))
-;;     (finally (stop))))
+(defn system-fixture
+  "Starts the system and migrates, no setup or tear down."
+  [f]
+  (let [conn {:connection-uri (-> config :db :uri)}]
+    (try
+      (go)
+      (migrate/migrate conn)
+      (insert-tenant conn {:db_uri "jdbc:postgresql://localhost/test_dash_tenant_1?user=dash&password=password"
```
kardan: @zzgvh I know you talked about refactoring this two inserts to a function, but for me I think this is all fine.

kardan:
Overall looks good. Cleanup the org.akvo.dash.fixtures require from non used stuff & I'll merge.

kardan:
One thing that I'm unsure about for the future is how to handle rollback and seeds. Maybe we should not rollback the tenants table. Or we could move what's in seed.sh to Clojure land. But have not looked / thought about what's best there.

PR #112, opened 2016-04-29
[#111] tear-down.sh takes force (-f) argument.

kardan:
Tear-down now accepts an -f to force client connections to be closed.


zzgvh:
Yep works fine!

PR #117, opened 2016-05-03
#78 more testing the db

zzgvh:
Get travis to run PG 9.5


zzgvh:
One of the main drawbacks as of now is that the deployments will take longer, since ```sudo: required``` results in travis building a full ubuntu VM rather than a container, if I understand thing right.

kardan:
Comment from @iperdomo to use 
```bash
sed -i 's/md5$/trust/g' /etc/postgresql/9.5/main/pg_hba.conf
```
instead of echo to not get duplicate rows.

PR #124, opened 2016-05-04
[#102] Maps visualisation and UI refactor

gabemart:
#102

This adds a "map" type visualisation with a leaflet map.

This PR also standardises the visualisation editor UI to reflect that each visualisation should only be able to draw data from one dataset


client/src/components/visualisation/ConfigMenu.jsx
```diff
@@ -0,0 +1,253 @@
+import React, { PropTypes } from 'react';
+import DashSelect from '../common/DashSelect';
+import ColumnMenu from './configMenu/ColumnMenu';
+import LabelColumnMenu from './configMenu/LabelColumnMenu';
+import LabelInput from './configMenu/LabelInput';
+
+
+const getDatasetArray = datasetObject => {
+  const datasetArray = [];
+  const sortFunction = (a, b) => a.name.toLowerCase() > b.name.toLowerCase();
```
iperdomo: AFAIK the sort needs to return `{-1, 0, 1}` (less than zero, zero or more than zero), but this expression returns `{true,false}` ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

client/src/components/charts/DashMap.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default function DashMap({ visualisation, datasets }) {
+  const chartData = chart.getChartData(visualisation, datasets);
+
+  return (
+    <div className="DashMap dashChart">
+      <Map
+        center={[0, 0]}
+        zoom={2}
+        scrollWheelZoom={false}
```
iperdomo: Do we want to enable this feature? I think it easier to zoom on a particular area if this is enabled? An example of other maps that use scroll to zoom:

> When the user scrolls with the mouse over the map, map zooms in and out (as it currently does).

https://github.com/akvo/akvo-flow/issues/1475

I would propose: `scrollWheelZoom='center'`

client/src/utilities/chart.js
```diff
@@ -1,29 +1,47 @@
...
+  let dataValues = [];
   let output = [];
 
   switch (visualisation.visualisationType) {
+    case 'map':
+
+      dataX.forEach((entry, index) => {
+        const label = nameDataX ? nameDataX.values[index] : null;
+        const newPositionObject = {
+          position: [parseFloat(dataX[index]), parseFloat(dataY[index])],
```
jonase: @gabemart Should this be the other way around? `[Lat, Lon] = [dataY, dataX]`

client/src/components/charts/DashMap.jsx
gabemart: This is perhaps a matter of personal taste, but I think scroll-to-zoom has very bad usability on embedded maps unless the map is the only feature on the page - otherwise, a user scrolls halfway down the page, then suddenly the embedded map hooks into the scroll events and the map starts zooming instead of the page, in a manner that's unintuitive and surprising.

Given that both the map creation dialog and the dashboards themselves are complex pages with several elements, and the user may reasonably expect to scroll up and down the page in the usual fashion, I am a strong advocate for disabling scroll-to-zoom.

client/src/components/visualisation/ConfigMenu.jsx
gabemart: Good point - the current function does sort the datasetObjects correctly via coercion, but will be inefficient with a lot of objects. Will change.

client/src/components/visualisation/ConfigMenu.jsx
```diff
@@ -0,0 +1,253 @@
...
+        value: index, label: entry.title,
+      });
+    });
+  }
+  return output;
+};
+
+const getSubTitle = subtitle => (
+  <h3 className="subtitle">{subtitle}</h3>
+);
```
jonase: I wonder if it would be better to turn this into a component, something like 
```jsx
const SubTitle = ({ children }) => (
  <h3 className="subtitle">{children}</h3>
);

SubTitle.propTypes = {
  children: PropTypes.node.isRequired,
};
```
so the call-sites can be written as `<SubTitle>X-Axis</SubTitle>` instead of `{getSubTitle('X-Axis')}`?

Note that "functional" components doesn't have to be in its own file so we can still treat it as just a helper component.

client/src/components/visualisation/ConfigMenu.jsx
```diff
@@ -0,0 +1,253 @@
...
+  Object.keys(datasetObject).forEach(key => {
+    datasetArray.push(datasetObject[key]);
+  });
+
+  datasetArray.sort(sortFunction);
+
+  return datasetArray;
+};
+
+const getDatasetOptions = datasetArray => {
```
jonase: Would it be clearer to use `map` instead of `forEach` here? Something like
```js
const getDatasetOptions = datasetArray => (
  datasetArray.map(option => ({
    value: option.id,
    label: option.name,
  }))
);
```

client/src/components/visualisation/ConfigMenu.jsx
```diff
@@ -0,0 +1,253 @@
...
+  datasetArray.forEach(option => {
+    output.push({
+      value: option.id, label: option.name,
+    });
+  });
+
+  return output;
+};
+
+const getDashSelectOptionsFromColumnArray = array => {
```
jonase: Same here. `map` instead of `forEach`?

client/src/components/charts/DashMap.jsx
Kiarii: I think Gabe makes a valid point; scrolling for map-only pages would work well while for embed maps, it may pose some usability issues: the user would have to clearly understand that scrolling on different areas does different things, while it may be clear in isolation that scrolling on a map zooms it and scrolling on a page takes the user up/down that page, together, it adds a cognitive task of first identifying where the mouse is and then determining if scrolling there would result to the desired scroll...

client/src/components/visualisation/ConfigMenu.jsx
gabemart: Yes! Nice. You correctly intuited that the only reason I didn't make this a component in the first instance is that it is too trivial for a separate file. This syntax is very succinct for a small component 👍

jonase:
👍

PR #129, opened 2016-05-10
[#80] CSV enhancements

gabemart:
#80


client/src/actions/dataset.js
```diff
@@ -139,10 +146,10 @@ export function saveDatasetSettings(id, { name }) {
 }
 
 // Only name for now.
-export function defineDatasetSettings({ name }) {
+export function defineDatasetSettings({ name, hasColumnHeaders }) {
```
jonase: @gabemart `hasColumnHeaders` is not needed here anymore

client/src/components/modals/DatasetSettings.jsx
```diff
@@ -9,11 +9,11 @@ class DatasetSettings extends Component {
     super(props);
     const { dataset } = this.props;
     this.state = { name: dataset.name };
-    this.handleChangeName = this.handleChangeName.bind(this);
+    this.handleChangeSettings = this.handleChangeSettings.bind(this);
   }
 
-  handleChangeName(newName) {
-    this.setState({ name: newName });
+  handleChangeSettings({ name, hasColumnHeaders }) {
```
jonase: Same here. I don't think we use `hasColumnHeaders` from this component anymore.

PR #134, opened 2016-05-11
Issue/130 tardis testing

zzgvh:
I'd like comments about the coding itself, as well as ideas on how to test the Tardis even more.

### DIFF
--- a/backend/test/org/akvo/dash/tardis_test.clj
+++ b/backend/test/org/akvo/dash/tardis_test.clj
@@ -1,0 +1,72 @@
+(ns org.akvo.dash.tardis-test
+  (:require
+    [akvo.commons.psql-util :refer [pgobj->val]]
+    [cheshire.core :as json]
+    [clj-time.core :as time]
+    [clj-time.coerce :as coerce]
+    [clojure.test :refer :all]
+    [hugsql.core :as hugsql]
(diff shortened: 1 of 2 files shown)


backend/test/org/akvo/dash/tardis_test.clj
```diff
@@ -34,39 +35,36 @@
...
+        (let [test-delete (fn [db_uri]
+                            (delete-dataset-by-id db_uri dataset)
+                            (is (=
+                                  0
+                                  ((count-history-dataset-by-ts db_uri {:now (sql-now)}) :count)))
+                            (is (=
+                                  2
+                                  ((count-all-history-datasets db_uri) :count))))]
+          (for-all-tenants test-delete)))
       )))
```
kardan: These trailing parens can be moved prior line.

kardan:
Minor style opinions which was covered in a call. But otherwise works fine when executed via
```bash
lein test :all
```
But there is an issue when testing via a REPL & editor and the system-fixture which always thinks it's running non test mode and try to use the normal db, not the test ones the provisioning script created. 

To be followed up.

kardan:
@zzgvh ping, you worked more on this?

iperdomo:
@kardan i think we should close the PR ? We're not actively working on this. Not sure if the code is in _good state_ to be merged?

PR #152, opened 2016-05-16
Issue/137 library view

jonase:


client/src/reducers/datasets.js
```diff
@@ -19,6 +19,37 @@ function saveDatasetSettings(state, dataset) {
   });
 }
 
+function importDatasetPending(state, { importId, name }) {
+  const now = Date.now();
```
gabemart: Perhaps we should grab the date in the action creator instead, to keep the reducers free of calls to non-pure functions? Same goes for line 36

jonase: Yes. I'll change this

kardan:
Looks good to me.

gabemart:
@jonase 👍 this PR looks good - could you please check my last commit (which gets rid of the SVG that was throwing the console warning) and then merge if all is well?

Issue #167, opened 2016-05-23
Move dataset-column definition as property in dataset_version


iperdomo:
Current implementation uses `dataset_column` table to hold metadata about the columns (type, title, sort-direction, [others]). 

It would make easier to create a new `dataset_version` if this metadata is a property (JSON encoded), instead of a table with a FK to `dataset`.

PR #168, opened 2016-05-24
[#167] Move column specification from dataset_column table to jsonb p…

jonase:
…ayload on dataset_version


backend/src/org/akvo/dash/endpoint/dataset.clj
```diff
@@ -11,18 +11,19 @@
...
+            table-name)))
 
 (defn find-dataset [conn id]
   (when-let [dataset (dataset-by-id conn {:id id})]
-    (let [columns (sort-by :column-order (dataset-columns-by-dataset-id conn {:id id}))
-          data (rest (jdbc/query conn [(select-data-sql (:table-name dataset) columns)] :as-arrays? true))
+    (let [columns (remove #(get % "hidden") (:columns dataset))
+          data (rest (jdbc/query conn [(select-data-sql (:table-name dataset)
+                                                        columns)]
+                                 :as-arrays? true))
```
iperdomo: This code indentation/format seems _different_ ?

jonase: Do you mean that

```clojure
[(select-data-sql (:table-name dataset)
                  columns)]
```
should be on a new line?

Issue #175, opened 2016-05-25
Refactor endpoints to use let-routes


kardan:
Using let-routes like (https://github.com/akvo/akvo-lumen/blob/295de07a9276a0ff29fc9b2a88c235e4e25a3ac9/backend/src/org/akvo/dash/endpoint/dashboard.clj#L14) would remove some duplicate code from endpoints.

kardan:
PR merged

PR #176, opened 2016-05-25
[#175] Refactored endpoints

kardan:
- Introduced let-routes for more elegant tenant-conn handling.
- Since the endpoints “format” had evolved over time I also made a
general review and polish of the endpoints to make them look fairly the
same.
- Removed the unused activity endpoint.


backend/src/org/akvo/dash/endpoint/dataset.clj
```diff
@@ -36,27 +36,24 @@
...
+            (not-found {:id id})))
 
-      (context "/transformations" []
+        (DELETE "/" _
+          (delete-dataset-by-id tenant-conn {:id id})
+          (response {:id id}))
 
-        (GET "/" []
-          (response {:fns []}))))))
+        (context "/transformations" []
```
jonase: In general, I think we should remove/don't add these kinds of stubs and only add things that we're working on. Before the transformations work is "done" we can't know if we need/want such an endpoint.

backend/src/org/akvo/dash/endpoint/import.clj
```diff
@@ -7,15 +7,14 @@
...
-  ""
-  [{tm :tenant-manager}]
+(defn endpoint [{:keys [tenant-manager]}]
   (context "/api/imports" {:keys [params tenant] :as request}
+    (let-routes [tenant-conn (connection tenant-manager tenant)]
 
-    (GET "/" []
-      (response/response (all-imports (connection tm tenant))))
+      (GET "/" _
+        (response/response (all-imports tenant-conn)))
```
jonase: I don't see why we should add endpoints that we're not using from the client.

backend/src/org/akvo/dash/endpoint/share.clj
```diff
@@ -35,14 +35,14 @@
...
+(defn endpoint [{:keys [tenant-manager]}]
   (context "/api/shares" {:keys [params tenant] :as request}
+    (let-routes [tenant-conn (connection tenant-manager tenant)]
 
-    (GET "/" []
-      (response {:index 0
-                 :items (collection (connection tenant-manager tenant))}))
+      (GET "/" _
+        (response {:index 0
+                   :items (collection tenant-conn)}))
```
jonase: Same here, I don't think we're using this, so do we need it?

Issue #164, opened 2016-05-21
Dashboard creation client prototype


gabemart:
A prototype for the Dashboard editor consisting of:

* [x] A canvas
* [x] To which entities can be added
* [x] On which entities can be dragged to new positions
* [x] On which entities can be resized
* [x] And have the entity contents display correctly after resizing
* [x] On which entities can be removed

An entity may be:

* [x] A visualisation (of any type)
* [x] A text element

gabemart:
closed via https://github.com/akvo/akvo-lumen/pull/181

PR #181, opened 2016-05-28
[#164] Dashboard editor

gabemart:
#164 

The actions to save a dashboard still needed to be added, but the editor itself is in a working, coherent state.


client/src/components/charts/DashMap.jsx
```diff
@@ -6,14 +6,19 @@ require('../../../node_modules/leaflet/dist/leaflet.css');
...
-export default function DashMap({ visualisation, datasets }) {
+export default function DashMap({ visualisation, datasets, width, height }) {
   const chartData = chart.getChartData(visualisation, datasets);
   return (
     <div className="DashMap dashChart">
       <Map
         center={[0, 0]}
         zoom={2}
         scrollWheelZoom={false}
+        key={width}
```
jonase: is a key needed here?

client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -0,0 +1,262 @@
...
+import DashboardCanvasItem from './DashboardCanvasItem';
+
+require('../../styles/DashboardEditor.scss');
+require('../../../node_modules/react-grid-layout/css/styles.css');
+require('../../../node_modules/react-resizable/css/styles.css');
+
+const getArrayFromObject = (object) => {
+  const arr = [];
+
+  Object.keys(object).forEach((key) => {
```
jonase: s/forEach/map ?

client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -0,0 +1,262 @@
...
+  });
+
+  const newIdInt = highestIdInt + 1;
+
+  return `${itemType}-${newIdInt}`;
+};
+
+const getFirstBlankRowGroup = (layout, height) => {
+  /* Function to find the first collection of blank rows big enough for the
+  /* default height of the entity about to beinserted. */
```
jonase: beinserted => be inserted

client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -0,0 +1,262 @@
...
+  layout.forEach(item => {
+    for (let row = item.y; row < (item.y + item.h); row++) {
+      occupiedRows[row] = true;
+      if (row > lastRow) lastRow = row;
+    }
+  });
+
+  /* Loop through every row from 0 to the last occupied. If we encounter a blank
+  /* row i, check the next sequential rows until we have enough blank rows to
+  /* fit our height. If we do, return row i. */
```
jonase: This comment style looks weird to me :). I look at `/*` and `*/` as "parens" that need to match. I would do

```
/* 
 * Loop through every row from 0 to the last occupied. If we encounter a blank
 * row i, check the next sequential rows until we have enough blank rows to
 * fit our height. If we do, return row i. 
 */
```

or

```
/* Loop through every row from 0 to the last occupied. If we encounter a blank
   row i, check the next sequential rows until we have enough blank rows to
   fit our height. If we do, return row i. */

```

client/src/containers/Dashboard.jsx
```diff
@@ -0,0 +1,118 @@
...
+            const spec = update(this.state.spec, { $merge: value });
+            this.setState({
+              isUnsavedChanges: true,
+              spec,
+            });
+          }}
+          onSaveDashboard={() => (
+            this.onSave()
+          )}
+*/
```
jonase: remove commented out code?

client/src/components/charts/DashMap.jsx
gabemart: I *think* that adding this key triggers a deeper re-render when the key changes - which is desirable with the maps visualisation because often, when the width changes, new tiles have to be loaded. Adding the key appeared to improve dynamic tile-loading on resizing, but I will test this further.

client/src/components/dashboard/DashboardEditor.jsx
gabemart: 👍

client/src/components/dashboard/DashboardEditor.jsx
gabemart: To be honest this is just how Sublime Text 2 auto-wraps long comments - but this is probably configurable somewhere, I'll take a look

jonase: ok. We can leave it like this then. it's just me needlessly bike-shedding 👍

jonase:
👍

PR #182, opened 2016-05-30
Issue/178 client transformations

jonase:


client/src/components/dataset/sidebars/ChangeDataType.jsx
```diff
@@ -0,0 +1,207 @@
...
+              options={dataTypeOptions}
+              onChange={type => {
+                if (type === 'date') {
+                  this.mergeArgs({ newType: 'date', dateFormat: 'YYYY-MM-DD' });
+                }
+                this.mergeArgs({ newType: type });
+              }}
+            />
+          </div>
+          {newType === 'date' && column.type === 'text' ?
```
gabemart: Apologies if this originally came from me, but what's the reason for checking the column type is text here?

client/src/components/dataset/DatasetTable.jsx
```diff
@@ -114,136 +54,118 @@ export default class DatasetTable extends Component {
   }
 
   getCellClassName(columnTitle) {
-    if (this.state.activeSidebar &&
-      this.state.activeSidebar.columnTitle === columnTitle) {
+    const { sidebarProps } = this.state;
+    if (sidebarProps != null && sidebarProps.columnTitle === columnTitle) {
```
gabemart: This works for "Filter" sidebar, but not for the change-data-type sidebar as that doesn't have the `columnTitle` property. For this to work correctly we need to also check if `sidebarProps.column.title === columnTitle`

client/src/components/dataset/sidebars/ChangeDataType.jsx
jonase: If the `column.type` is of type `number` this is a number->date transform. I will interpret the number as a timestamp so no date format needs to be specified. Perhaps we later add the ability to interpret the number as year, in which case you would here pick between timestamp or year. We haven't discussed this so that's why I don't do it.

TLDR; the available date format options (DD-MM-YYYY, MM-DD-YYYY, etc.) doesn't make sense when the column type in number

client/src/components/dataset/DatasetTable.jsx
jonase: 👍 I'll fix this. Also, we should compare columnName instead of title since we now have that property available

client/src/components/dataset/sidebars/ChangeDataType.jsx
gabemart: Ah I see - turns out I misunderstood what the date format option was for, I was viewing it as a display option rather than a parsing option. I can now see why this makes sense

jonase: Yeah, it's a hint on how to parse the text. We store dates as timestamps
internally

gabemart:
👍 LGTM

PR #186, opened 2016-05-30
Issue/98 transformations

iperdomo:


jonase:
@iperdomo It seems like the `/api/transformations` endpoint is not hooked up to the [system](https://github.com/akvo/akvo-lumen/blob/issue/98-transformations/backend/src/org/akvo/dash/system.clj#L17-L25) yet, so it can't be tested via curl?

iperdomo:
@jonase Should be fixed in rev bee8a2102

jonase:
@iperdomo Failed imports seems to be stuck on `PENDING`. i.e. `/api/imports/<job-exec-id>` doesn't seem to return `FAIL` in case of failure anymore. 

On a related note, I wonder if we should (a) fix this, and (b) move this to another endpoint (for example `/api/job-executions/<jid>`) and make it support polling for transformation status as well (which may already work, but I think failures are reported differently). This will be needed when I integrate the client.

backend/resources/org/akvo/dash/migrations_tenants/004-job-execution.up.sql
```diff
@@ -1,11 +1,12 @@
...
-    data_source_id text REFERENCES data_source,
-    started_at timestamptz NOT NULL DEFAULT now(),
-    finished_at timestamptz,
-    error_reason text,
-    created timestamptz DEFAULT now(),
-    modified timestamptz DEFAULT now()
+    data_source_id text REFERENCES data_source ON DELETE CASCADE,
+    dataset_id text REFERENCES dataset ON DELETE CASCADE,
+    type text NOT NULL,
+    log text,
```
jonase: Why is the `log` of type `text`? What kinds of things could it/should it hold?

jonase:
The `lumen_to_date` function returns a string, and not a timestamp, if the default value ends up being used:

```
dash_tenant_1=# select lumen_to_date('"24-12-2016"'::jsonb, '"01-01-2016"'::jsonb, 'DD-MM-YYYY', false);
 lumen_to_date
---------------
 1482530400

dash_tenant_1=# select lumen_to_date('"not-a-date"'::jsonb, '"01-01-2016"'::jsonb, 'DD-MM-YYYY', false);
 lumen_to_date
---------------
 "01-01-2016"
```

backend/resources/org/akvo/dash/migrations_tenants/007-database-functions.up.sql
```diff
@@ -39,50 +26,169 @@ BEGIN
...
   LANGUAGE plpgsql IMMUTABLE STRICT;
 
-CREATE OR REPLACE FUNCTION titlecase(val jsonb)
+CREATE OR REPLACE FUNCTION lumen_trim_double(val jsonb)
   RETURNS jsonb AS
-$BODY$
+$$
 BEGIN
-  RETURN initcap(val::text);
+  RETURN regexp_replace(val::text, '\s+', ' ', 'g');
```
jonase: Should `lumen_trim_double` collapse newlines and tabs into a single space as well?

backend/resources/org/akvo/dash/migrations_tenants/007-database-functions.up.sql
```diff
@@ -39,50 +26,169 @@ BEGIN
...
 END;
-$BODY$
+$$
   LANGUAGE plpgsql IMMUTABLE STRICT;
 
-CREATE OR REPLACE FUNCTION trim_space(val jsonb)
+CREATE OR REPLACE FUNCTION lumen_trim(val jsonb)
```
jonase: Should trim tabs/newlines as well?

```
dash_tenant_1=# select lumen_trim('"  foo   \n"'::jsonb);
     lumen_trim
---------------------
 "foo   \n"
```

backend/resources/org/akvo/dash/migrations_tenants/007-database-functions.up.sql
```diff
@@ -39,50 +26,169 @@ BEGIN
...
+
+  IF tbl_name IS NULL OR tbl_name = '' THEN
+    RAISE EXCEPTION 'tbl_name is required';
+  END IF;
+
+  IF col_name IS NULL OR char_length(col_name) = 0 THEN
+    RAISE EXCEPTION 'col_name is required';
+  END IF;
+
+  IF new_type IS NULL OR char_length(new_type) = 0 THEN
```
jonase: could we explicitly check for `text/date/number` here?

backend/src/org/akvo/dash/endpoint/transformation.clj
```diff
@@ -0,0 +1,11 @@
+(ns org.akvo.dash.endpoint.transformation
+  (:require [compojure.core :refer :all]
+            [org.akvo.dash.component.tenant-manager :refer [connection]]
+            [org.akvo.dash.transformation :as t]))
+
+(defn endpoint [{:keys [tenant-manager config]}]
+  (context "/api/transformations" {:keys [tenant] :as request}
+    (let-routes [tenant-conn (connection tenant-manager tenant)]
+      (context "/:id" [id]
```
jonase: If we only ever have one endpoint we could drop the context here and do `(POST "/:id" ...`?

backend/src/org/akvo/dash/import.sql
```diff
@@ -30,20 +30,20 @@ SELECT spec
 -- :name update-failed-job-execution :! :n
 -- :doc Update failed job execution
 UPDATE job_execution
-   SET finished_at=now(),
-       error_reason=:reason
- WHERE id=:id
+   SET log = :reason,
```
jonase: I'm not sold on the terminology `log` here. To me a log implies more than one thing, and not a `text`. If the log is used for storing errors, could we rename it to `error_log` or something instead?

backend/src/org/akvo/dash/endpoint/transformation.clj
```diff
@@ -0,0 +1,11 @@
+(ns org.akvo.dash.endpoint.transformation
+  (:require [compojure.core :refer :all]
+            [org.akvo.dash.component.tenant-manager :refer [connection]]
+            [org.akvo.dash.transformation :as t]))
+
+(defn endpoint [{:keys [tenant-manager config]}]
```
jonase: drop `config`. It's not used, and not even supplied to the endpoint

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+
+(defn- throw-invalid-op
+  [op-spec]
+  (throw (Exception. (str "Invalid operation " op-spec))))
+
+(defn required-keys
+  [{:strs [op args onError] :as op-spec}]
+  (or (boolean (and (ops-set op)
+                    (not-empty args)
+                    (re-find #"c\d+" (args "columnName"))
```
jonase: I didn't know that the column names must start with `c`? Do we depend on this somewhere?

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+                    (on-error-set onError)))
+      (throw-invalid-op op-spec)))
+
+(defmulti valid-op?
```
jonase: `valid-op?` doesn't seem to be a proper predicate even if its name ends with `?`. Instead, it either returns `true`, or throws an exception.

Perhaps rename to something like `validate-op`?

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+(def on-error-set #{"fail" "default-value" "delete-row"})
+(def type-set #{"number" "text" "date"})
+(def sort-direction-set #{"ASC" "DESC"})
+
+(hugsql/def-db-fns "org/akvo/dash/transformation.sql")
+
+
+(defn- throw-invalid-op
+  [op-spec]
+  (throw (Exception. (str "Invalid operation " op-spec))))
```
jonase: `clojure.core/ex-info` is nice for things like this: `(throw (ex-info "Invalid operation" op-spec))`.

Then, on catch, you can get the op-spec as data using `clojure.core/ex-data`.

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+                 (not-empty (args "defaultValue"))
+                 (if (= "date" (args "newType"))
+                   (not-empty (args "parseFormat"))
+                   true)))
+      (throw-invalid-op op-spec)))
+
+(defmethod valid-op? :core/change-column-title
+  [{:strs [args] :as op-spec}]
+  (or (boolean (and (required-keys op-spec)
+                    (not-empty (args "columnTitle"))))
```
jonase: `(not-empty (args "columnTitle"))` is roughly the same as `(contains? args "columnTitle")`.

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+  [op-spec]
+  (or (required-keys op-spec)
+      (throw-invalid-op op-spec)))
+
+(defmethod valid-op? :core/trim-doublespace
+  [op-spec]
+  (or (required-keys op-spec)
+      (throw-invalid-op op-spec)))
+
+(defn valid?
```
jonase: same argument as above. If the function is not a predicate, perhaps it should not end in `?`? Rename to `validate`?

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+                                               columns
+                                               op-spec)]
+              (if (:success? step)
+                (conj log step)
+                (throw (Exception. (str "Error applying operation: " op-spec))))))]
+    (try
+      (copy-table tenant-conn {:source-table source-table
+                               :dest-table imported-table})
+      (copy-table tenant-conn {:source-table imported-table
+                               :dest-table table-name})
```
jonase: Why do we clone twice?

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -1,225 +1,157 @@
...
+      (let [result (reduce f [] transformations)
+            log (str/join "\n" (filter :message result))
+            cols (:columns (last result))]
+        (new-dataset-version tenant-conn {:id (str (squuid))
+                                          :dataset-id dataset-id
+                                          :job-execution-id job-id
+                                          :table-name table-name
+                                          :imported-table-name imported-table
```
jonase: Should this not point to `source-table`?

backend/resources/org/akvo/dash/migrations_tenants/004-job-execution.up.sql
iperdomo: The idea behind `log` as text, is that we should it could hold just one line (an error reason) or a list of _statements_ from a transformation execution (lines separated by `"\n"`)

backend/src/org/akvo/dash/transformation/engine.clj
```diff
@@ -1,31 +1,180 @@
...
-  [engine value]
-  (js-eval engine (str "'" value "'.toUpperCase()")))
+  (:require [clojure.java.jdbc :as jdbc]
+            [hugsql.core :as hugsql])
+  (:import java.sql.SQLException))
+
+
+(hugsql/def-db-fns "org/akvo/dash/transformation/engine.sql")
+
+(def available-ops
```
jonase: Could be a set instead?

backend/resources/org/akvo/dash/migrations_tenants/007-database-functions.up.sql
iperdomo: Good question, not sure if we should trim new-lines, tabs seems more reasonable

backend/resources/org/akvo/dash/migrations_tenants/007-database-functions.up.sql
iperdomo: Yes. There has been several requests to implement _assertions_ in plpgpsql, but `Assert` is already a reserved word in PG, so this checks become verbose.

backend/src/org/akvo/dash/import.sql
iperdomo: As mentioned the `log` is just to hold statement(s) of the execution. If the status is `FAILED` you get the error from the log, if the status is `OK`, you could get the list of changes applied in the transformation. In the future will hold the execution log for any async process.

backend/src/org/akvo/dash/endpoint/transformation.clj
iperdomo: agree

backend/src/org/akvo/dash/transformation.clj
iperdomo: No, it's not written anywhere, i just wanted to add a "statement" that is not empty with a valid _schema_. I can just change that to `not-empty?`

backend/src/org/akvo/dash/transformation.clj
iperdomo: Agree

backend/src/org/akvo/dash/transformation.clj
iperdomo:
````clojure
(contains? {"columnTitle" ""} "columnTitle")
;; => true
````
I guess we want non-empty string?

File: backend/src/org/akvo/dash/transformation.clj (hunk 12 above)
Comment (iperdomo) [inline backend/src/org/akvo/dash/transformation.clj]: Agree. those started as predicate functions returning true/false and then the implementation went in a different direction :wink:

File: backend/src/org/akvo/dash/transformation.clj (hunk 13 above)
Comment (iperdomo) [inline backend/src/org/akvo/dash/transformation.clj]: We have a unique constraint in the `dataset_version` table that doesn't allow us to have the same `dataset_version` with the same `imported_table_name` (a waste of space!). We can remove this double cloning and more than one `dataset_version` can depend on the same `imported_table_name`

File: backend/src/org/akvo/dash/transformation.clj (hunk 14 above)
Comment (iperdomo) [inline backend/src/org/akvo/dash/transformation.clj]: See the explanation in my previous comment.

File: backend/src/org/akvo/dash/transformation.clj [hunk 16]
```diff
@@ -1,225 +1,157 @@
 (ns org.akvo.dash.transformation
```
jonase: Could we try to normalize the terminology a bit regarding transformations?

Right now we're using at least three different words for the same thing: `op-spec`, `operation` and `transformation`. Personally I like `transformation` the most so I would suggest 

* `transformation`: a map with keys `:op`, `:args`, `:onError``
* `transformation-log`: A vector of transformations

backend/src/org/akvo/dash/transformation.sql
```diff
@@ -0,0 +1,45 @@
+
+-- :name new-job-execution :! :n
```
jonase: should this be renamed to something like `new-transformation-job-execution` since we're setting the type to `TRANSFORMATION`? Or, take the type as argument.

backend/src/org/akvo/dash/transformation.sql
```diff
@@ -0,0 +1,45 @@
...
+-- :doc Inserts a new transformation job execution
+INSERT INTO job_execution (id, dataset_id, type)
+VALUES (:id, :dataset-id, 'TRANSFORMATION')
+
+-- :name update-transformations :! :n
+-- :doc Updates the transformations property for a given dataset
+UPDATE dataset
```
jonase: Did we talk about putting the transformations on the `dataset_version` instead?

backend/src/org/akvo/dash/transformation.sql
```diff
@@ -0,0 +1,45 @@
...
+-- :doc Inserts a new dataset version
+INSERT INTO dataset_version (id, dataset_id, job_execution_id, version,
+                             table_name, imported_table_name, columns)
+VALUES (:id, :dataset-id, :job-execution-id, :version,
+        :table-name, :imported-table-name, :columns::jsonb)
+
+
+-- :name copy-table :!
+-- :doc Copy a table defition (including data)
+CREATE TABLE :i:dest-table AS TABLE :i:source-table
```
jonase: We have the exact same function elsewhere I think

backend/src/org/akvo/dash/transformation/engine.clj
```diff
@@ -1,31 +1,180 @@
...
+  [columns op-spec]
+  (throw (Exception. (str "Column metadata changes not supported on operation [" (op-spec "op") "]"))))
+
+(defmethod column-metadata-operation :core/sort-column
+  [columns op-spec]
+  (let [col-name (get-column-name op-spec)
+        col-idx (get-column-idx columns col-name)
+        sort-idx (get-sort-idx columns)
+        sort-direction (get-in op-spec ["args" "sortDirection"])]
+    (update-in columns [col-idx] assoc "sort" sort-idx "direction" sort-direction)))
```
jonase: you can use `clojure.core/update` for one-level-deep updates

iperdomo: :+1: will do

backend/test/org/akvo/dash/endpoint/transformation_test.clj
```diff
@@ -0,0 +1,10 @@
+(ns org.akvo.dash.endpoint.transformation-test
+  (:require [clojure.test :refer :all]
+            [org.akvo.dash.endpoint.transformation :as transformation]))
+
+(def handler
+  (transformation/endpoint {}))
+
+(deftest a-test
```
jonase: remove unused?

backend/resources/org/akvo/dash/migrations_tenants/004-job-execution.up.sql
jonase: Should we not do `text[]` in that case? What if the log message contains `\n`?

backend/resources/org/akvo/dash/migrations_tenants/007-database-functions.up.sql
jonase: trim usually has the semantics of "trim whitespace" which includes both tabs and newlines

backend/src/org/akvo/dash/transformation.clj
jonase: 👍

backend/src/org/akvo/dash/transformation.clj
jonase: yes, we should remove that unique constraint in that case

backend/src/org/akvo/dash/transformation.sql
iperdomo: After the merge, we can move some `common` functions to a different `.sql` file.

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -115,26 +114,22 @@
                 (throw (Exception. (str "Error applying operation: " op-spec))))))]
     (try
       (copy-table tenant-conn {:source-table source-table
-                               :dest-table imported-table})
-      (copy-table tenant-conn {:source-table imported-table
                                :dest-table table-name})
       (let [result (reduce f [] transformations)
-            log (str/join "\n" (filter :message result))
+            log (mapcat :execution-log result)
```
jonase: nice!

backend/src/org/akvo/dash/transformation.clj
```diff
@@ -115,26 +114,22 @@
...
-                                           :status "OK"
-                                           :log log}))
+        (update-job-success-execution tenant-conn {:id job-id
+                                                   :exec-log log}))
       (catch Exception e
-        (update-job-execution tenant-conn {:id job-id
-                                           :status "FAILED"
-                                           :log (.getMessage e)})))))
+        (update-job-failed-execution tenant-conn {:id job-id
+                                                  :error-log (.getMessage e)})))))
```
jonase: this will probably fail? getMessage returns a string, and not an array

PR #191, opened 2016-05-31
Issue/178 client transformations

jonase:


client/src/reducers/transforms/changeDatatype.js
```diff
@@ -6,7 +6,7 @@ function makeParser(parser) {
   return (rows, idx, onError, args) => {
     const deleteMarker = {};
     const { defaultValue } = args;
-    const dv = defaultValue != null ? parser(defaultValue) : null;
+    const dv = defaultValue != null ? defaultValue : null;
```
gabemart: Is the purpose of this to catch cases where defaultValue is undefined and convert it to null?

client/src/components/dataset/sidebars/TransformationLog.jsx
```diff
@@ -10,9 +10,9 @@ function transformationDescription({ op, args }, columns) {
...
       return `${columnTitle(columnName, columns)} to lowercase`;
     case 'core/to-uppercase':
-      return `${columnTitle(columnName, columns)} to lowercase`;
+      return `${columnTitle(columnName, columns)} to uppercase`;
     case 'core/to-titlecase':
-      return `${columnTitle(columnName, columns)} to lowercase`;
+      return `${columnTitle(columnName, columns)} to titlecase`;
     case 'core/trim':
       return `${columnTitle(columnName, columns)} trimmed whitespace`;
     case 'core/trim-doublespace':
```
gabemart: Perhaps include a case for sorting. Something like

````    
case 'core/sort-column':
  return `${columnTitle(columnName, columns)} sorted ${args.sortDirection}`;
````

gabemart:
LGTM aside from the small notes mentioned

client/src/reducers/transforms/changeDatatype.js
jonase: I'll change this to just let the default value flow through.. I missed this in the refactor. Good catch!

client/src/components/dataset/sidebars/TransformationLog.jsx
jonase: Yeap I'll add this

gabemart:
Looks good 👍

Issue #187, opened 2016-05-31
Share dashboard backend


kardan:
Extend current "share" feature to include dashboards.

- [x] Backend for creating & removing a "share" of a dashboard.
- [x] Extend the public endpoint to work with shares of type dashboard.
- [ ] Nginx rule for docker (https://github.com/akvo/akvo-provisioning/issues/305).

PR #197, opened 2016-06-02
#187 share dashboard

kardan:


backend/test/org/akvo/dash/endpoint/dataset_test.clj
```diff
@@ -10,16 +9,3 @@
 (deftest handler-test
```
iperdomo: Remove is not used?

iperdomo:
The last push failed because an infrastructure issue (travis didn't manage to install PG 9.5). I've executed the tests in a local environment.

PR #199, opened 2016-06-02
[#190] dashboard saving

gabemart:
#190


client/src/actions/dashboard.js
```diff
@@ -0,0 +1,164 @@
...
+export function saveDashboardChanges(dashboard) {
+  const now = Date.now();
+  const dash = Object.assign({}, dashboard, {
+    modified: now,
+  });
+  const id = dash.id;
+
+  return (dispatch) => {
+    dispatch(editDashboardRequest);
+    fetch(`/api/dashboards/${dash.id}`, {
```
jonase: you have `id` bound to `dash.id`. Could use that one instead

client/src/containers/Dashboard.jsx
```diff
@@ -67,23 +67,28 @@ class Dashboard extends Component {
...
-        modified: dash.modified,
-      });
-
-      Object.keys(dash.entities).forEach(key => {
-        const entity = dash.entities[key];
-        if (entity.type === 'visualisation') {
-          this.onAddVisualisation(this.props.library.visualisations[key].datasetId);
-        }
-      });
+      const dashboardIsEmpty = Boolean(Object.keys(this.state.entities).length === 0 &&
```
jonase: For my own understanding: Why do you wrap the expression in `Boolean`. AFAICT the expression already evals to a bool.

jonase:
lgtm 👍

PR #202, opened 2016-06-03
#187 share dashboard

kardan:


backend/src/org/akvo/dash/endpoint/public.clj
```diff
@@ -28,24 +28,33 @@
     {"visualisation" (dissoc v :id :created :modified)
      "datasets"      {(:id d) d}}))
 
+(defn- visualisation-id-list
+  [dashboard]
+  (remove nil? (mapv (fn [m]
+                       (if (= (get m "type") "visualisation")
+                         (get m "id")))
+                     (vals (:entities dashboard)))))
```
iperdomo: what about this expression:
````clojure
(->> dashboard
       :entities
       vals
       (filter #(= "visualisation" (get % "type")))
       (mapv #(get % "id")))
````

File: backend/src/org/akvo/dash/endpoint/public.clj [hunk 2]
```diff
@@ -28,24 +28,33 @@
...
+
+(defn visualisation-list [tenant-conn visualisation-ids]
+  (reduce conj {} (map (fn [v-id]
+                         {v-id (visualisation tenant-conn v-id)})
+                       visualisation-ids)))
+
+(defn dataset-list
+  [tenant-conn visualisations]
+  (let [dataset-ids (vec (reduce conj #{} (map (fn [v]
+                                                 (:datasetId v))
```
iperdomo: there is no need to wrap the keyword in a function as keywords can be used in the function position, e.g.
 `(map :datasetId (vals visualisations))`

backend/src/org/akvo/dash/endpoint/public.clj
```diff
@@ -30,10 +30,11 @@
 
 (defn- visualisation-id-list
   [dashboard]
-  (remove nil? (mapv (fn [m]
-                       (if (= (get m "type") "visualisation")
-                         (get m "id")))
-                     (vals (:entities dashboard)))))
+  (remove nil? (->> dashboard
```
iperdomo: There will be no `nil` values since you're filtering by `"visualisation"` ? so no need to `(remove nil? coll)` ?

Issue #196, opened 2016-06-01
Add component for viewing shared/published dashboards


gabemart:

gabemart:
Closed via #205

PR #205, opened 2016-06-03
[#196] Dashboard sharing and viewer component

gabemart:


client/src/components/dashboard/DashboardViewer.jsx
```diff
@@ -0,0 +1,144 @@
...
+      canvasWidth: width,
+      viewportType: viewport,
+    });
+  }
+
+  render() {
+    const { dashboard, datasets } = this.props;
+    const layout = dashboard.layout;
+
+    const sortFunc = (a, b) => {
```
jonase: Should we move this function out of the `render()` method? It gets recreated on each render which is not necessary.

client/src/components/dashboard/DashboardViewer.jsx
```diff
@@ -0,0 +1,144 @@
...
+        return 1;
+      } else if (ax < bx) {
+        return -1;
+      } else if (ax > bx) {
+        return 1;
+      }
+      return 0;
+    };
+
+    const sortedDashboard = getArrayFromObject(dashboard.entities).sort(sortFunc);
```
jonase: Since [Array.prototype.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) is destructive, should we slice(0) first?

```
> foo = [2,1,3];
[2, 1, 3]
> foo.sort();
[1, 2, 3]
> foo
[1, 2, 3] // Note, foo changed
> foo = [2,1,3];
[2, 1, 3]
> foo.slice(0).sort();
[1, 2, 3]
> foo
[2, 1, 3]
```

jonase:
@gabemart Can some styling improvements be made to the share modal?

gabemart:
@jonase yes this is on the todo list but there is a separate issue for this

client/src/components/dashboard/DashboardViewer.jsx
gabemart: Does it make a difference here? getArrayFromObject() returns a new array that isn't stored, I'm not sure I see the advantage of cloning it before sorting

jonase: if getArrayFromObject returns a new array there's no need to slice 👍

jonase:
lgtm 👍

PR #210, opened 2016-06-03
Issue/98 178 transformations integration

iperdomo:


backend/src/org/akvo/dash/import/csv.clj
```diff
@@ -32,7 +32,7 @@
   (format "CREATE %s %s (%s, %s)"
           (if temp? "TEMP TABLE" "TABLE")
           t-name
-          (if temp? "rnum serial" "rnum integer")
+          (if temp? "rnum serial primary key" "rnum integer primary key")
```
jonase: for my understanding. Why is non-temp not `serial`?

iperdomo: Serial is just a short-hand for having an `integer` column and a default value a call to a _sequence_ function. In the case of the non-temp table, we're not going to add more rows, so we no need to generate more `rnum` values via call to the sequence function. It's a minor optimization to not have more dependent objects than needed.

https://www.postgresql.org/docs/8.1/static/datatype.html#DATATYPE-SERIAL

jonase: Thanks for the explanation, that makes total sense. The integers are
already generated in the temp table

Issue #229, opened 2016-06-08
Text to number transform with {onError: "fail"} fails


jonase:
We need to catch if parsing fails and don't touch the dataset. Open question on how to report this to the user.

PR #233, opened 2016-06-09
[#229] Catch and return original dataset if the error strategy is 'fail'

jonase:
* We need to figure out how to communicate to the user that a transformation
  was aborted (as requested by the user).


gabemart:
Checked 👍

Issue #220, opened 2016-06-08
Dashboard editor text input elements cross-browser issues


gabemart:
We should replace the contentEditable component with a basic solution using a `<textarea>` tag. react-contenteditable appears to have cross browser issues

Additionally, we should delete placeholder text on focus

gabemart:
Closed via #239

PR #239, opened 2016-06-09
[#220] dashboard editor text entities

gabemart:
#220


jonase:
This seems to work as expected 👍 

I noticed another possible problem in safari but perhaps that should be treated as a separate issue:

image

jonase:
and in firefox:

image

gabemart:
@jonase I agree, I think we should treat the title issue as a separate issue. I will open one and self assign

jonase:
👍 then I'll merge this one.

Issue #238, opened 2016-06-09
Sort by name in library view is breaks when dashboards are defined


jonase:
Probably because dashboards has `title` instead of `name`.

PR #240, opened 2016-06-09
[#238] Resolve library sorting by name

jonase:
* Datasets & visualisations have `name` property while
  dashboards have `title` property instead
* We'll move to using `title` everywhere later, both on
  the client and on the backend.


gabemart:
:+1:

PR #244, opened 2016-06-09
[#242] Make dataset permalinks work properly

jonase:

### DIFF
--- a/client/src/components/Library.jsx
+++ b/client/src/components/Library.jsx
@@ -30,11 +30,7 @@ class Library extends Component {
   }
 
   componentDidMount() {
-    const { datasets } = this.props;
-    // TODO better heuristics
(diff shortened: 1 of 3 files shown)


jonase:
@gabemart please hold the merge on this one.. there are some subtle issues with pending transformations that needs to be resolved

gabemart:
@jonase The library container now makes an API call every time it is mounted, e.g. it will refetch the library if you open a dataset, then return to the library. This is the intended behaviour, correct? Because the library call only fetches metadata about the entities, so we can afford to refetch on every mount?

jonase:
@gabemart that was my thinking, yes. We could be smarter as well and check if all the entity types are empty, and only fetch if that's the case.. but perhaps that's an optimisation for the future?

gabemart:
@jonase yep :+1:

PR #269, opened 2016-06-13
[#261] Make keycloak url a config

kardan:
- [ ] Copyright header
- [x] Code formatting
- [x] Documentation

### DIFF
--- a/backend/src/org/akvo/lumen/middleware.clj
+++ b/backend/src/org/akvo/lumen/middleware.clj
@@ -1,9 +1,9 @@
 (ns org.akvo.lumen.middleware
   (:require [akvo.commons.jwt :as jwt]
             [clj-http.client :as client]
-            [clojure.pprint :refer [pprint]]
             [clojure.string :as s]
-            [ring.util.response :as response]))
+            [ring.util.response :as response])
+  (:import [java.sql SQLException]))
 
 
 (defn wrap-auth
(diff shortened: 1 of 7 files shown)


backend/src/org/akvo/lumen/middleware.clj
```diff
@@ -1,25 +1,26 @@
 (ns org.akvo.lumen.middleware
-  ""
-  (:require
-   [clojure.string :as s]
-   [akvo.commons.jwt :as jwt]
-   [clj-http.client :as client]
-   [ring.util.response :as response]))
+  (:require [akvo.commons.jwt :as jwt]
+            [clj-http.client :as client]
+            [clojure.pprint :refer [pprint]]
```
iperdomo: pprint?

kardan: There is a (pprint e) form in wrap-jwt
https://github.com/akvo/akvo-lumen/pull/269/files#diff-eba5c89e38c7b1943887582cee409557R46

kardan:
@iperdomo ready for review again.

client/webpack.config.js
```diff
@@ -56,7 +56,10 @@ module.exports = {
     new webpack.HotModuleReplacementPlugin(),
     new SystemBellPlugin(),
     new webpack.DefinePlugin({
-      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV)
+      'process.env': {
+        'NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+        'LUMEN_KEYCLOAK_URL': JSON.stringify(process.env.LUMEN_KEYCLOAK_URL)
```
iperdomo: for my understanding: Why are we using `JSON.stringify` if the  `LUMEN_KEYCLOAK_URL` is just a string?

kardan: Without it we get syntax error cased by the ":" in the url. My JS skills are not good enough to explain why.

jonase: @iperdomo The stings are inserted verbatim from the environment **at build time**. Without JSON.stringify we would end up with something like

```
url: http://localhost:8080
```

instead of

```
url: "http://localhost:8080"
```

backend/src/org/akvo/lumen/middleware.clj
iperdomo: Still the exception doesn't need to be `pprint`ed ? ... perhaps you want (.printStackTrace e) ?

client/webpack.config.js
iperdomo: @jonase thanks for the explanation. This doc suggests that the strings should be strings? 

https://nodejs.org/api/process.html#process_process_env

PR #278, opened 2016-06-17
[#247] dashboard permalinks

gabemart:
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/components/dashboard/DashboardCanvasItem.jsx
+++ b/client/src/components/dashboard/DashboardCanvasItem.jsx
@@ -103,7 +103,7 @@ export default class DashboardCanvasItem extends Component {
           </div>
         }
         <button
-          className="clickable deleteButton"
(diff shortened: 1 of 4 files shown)


jonase:
Refreshing the browser on a dashboard like `http://t1.lumen.localhost:3030/dashboard/<uuid>`, the page is rendered correctly ~50% of the time. When it fails I get 

> ReactCompositeComponent.js:559Uncaught TypeError: Cannot read property '_currentElement' of null

in the console. Can you reproduce this @gabemart ?

gabemart:
@jonase reproduced and fixed, thank you. Error was caused by a race condition where both the dataset and dashboard API calls returned before the library api call

jonase:
👍 Tested, and this seems to work. Well done @gabemart !

Issue #282, opened 2016-06-19
Enhance dashboard editor name input


gabemart:

gabemart:
closed via #283

PR #283, opened 2016-06-19
[#282] dashboard name input

gabemart:
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/components/dashboard/DashboardEditor.jsx
+++ b/client/src/components/dashboard/DashboardEditor.jsx
@@ -74,12 +74,13 @@ export default class DashboardEditor extends Component {
     this.state = {
       gridWidth: 1024,
       propLayout: [],
+      saveError: false,
     };
     this.handleLayoutChange = this.handleLayoutChange.bind(this);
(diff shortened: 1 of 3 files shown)


gabemart:
@jonase I want to add a notification bar message if the user tries to save while the name input is blank, but I think the easiest way to do this is to wait for that PR to hit release and then create a new branch, rather than merging into that branch now

client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -181,6 +183,22 @@ export default class DashboardEditor extends Component {
     this.props.onUpdateEntities(newEntities);
   }
 
+  handleNameFocus() {
```
jonase: Would it be possible to use the placeholder `input` attribute (as described in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) instead of hooking into `onFocus`?

gabemart:
@jonasa agreed this is a cleaner approach

jonase:
👍

Issue #289, opened 2016-06-21
Rearrange inputs in visualisation editor


gabemart:
![image](https://cloud.githubusercontent.com/assets/2589977/16236292/c4c44e98-37cf-11e6-9569-e50c44b83a39.png)

gabemart:
closed via #291

PR #291, opened 2016-06-21
[#289] Change visualisation editor input order to match feedback

gabemart:
#289 

- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

@jonase these are the requested changes to the visualisation editor in advance of the MVP demo.

It's clear that some deeper changes are needed - if we want to keep this new order, then the names for the various visualisation properties should also be changed to make things less confusing. But given the time constraints, I think these changes are enough for now, and we can have a discussion of the best way forward after the demo.


jonase:
👍 lgtm

Issue #258, opened 2016-06-10
Move middleware to auth-middleware namespace


kardan:
Move org.akvo.lumen.middleware to org.akvo.lumen.auth-middleware

PR #304, opened 2016-06-28
[#258] [#303] Lumen now pass correct issuer to wrap-jwt-claims

kardan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

Also renamed middleware namespace to auth.

### DIFF
--- a/backend/src/org/akvo/lumen/auth.clj
+++ b/backend/src/org/akvo/lumen/auth.clj
@@ -42,4 +42,4 @@
       (jwt/wrap-jwt-claims handler (jwt/rsa-key certs 0) issuer))
     (catch Exception e
       (println "Could not get cert from Keycloak")
-      (.printStackTrace e))))
+      (throw e))))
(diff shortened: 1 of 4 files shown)


backend/src/org/akvo/lumen/auth.clj
```diff
@@ -33,13 +33,13 @@
...
-  [handler issuer]
+  [handler keycloak-url realm]
   (try
-    (let [certs (-> (str issuer "/realms/akvo/protocol/openid-connect/certs")
+    (let [issuer (str keycloak-url "/realms/" realm)
+          certs (-> (str issuer "/protocol/openid-connect/certs")
                     client/get
                     :body)]
       (jwt/wrap-jwt-claims handler (jwt/rsa-key certs 0) issuer))
     (catch Exception e
```
jonase: @kardan Why do we need to catch, print and swallow exceptions here? Is there any point in continuing starting the server if we can't get the cert, especially since we'll return `nil` and not a handler?

kardan: Agree, we should probably kill the server. I added a try / catch to be able to print "Could not get cert from Keycloak" to make it easier to understand if you are new to the system.

PR #305, opened 2016-06-29
Issue/58 update client deps

jonase:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


gabemart:
@jonase added a small commit to fix behaviour with the new version of the `Select` component. Previously we could give it a number as a value and it would coerce it to a string, now we need to explicitly give it a string (which is for the best anyway).

Code checked and tested, please feel free to merge once you have looked over my commit

jonase:
@gabemart LGTM. Nice to see the `placeholder` prop. Is this a new feature for this version of react-select?

Issue #263, opened 2016-06-13
Rethink how the http api works from the clients perspective


jonase:
- Current approach with redux-thunk works but is very low level and lots of boilerplate.
- Evaluate other approaches (promises, redux-saga, etc.)

jonase:
There are *a lot* of options here. This is not a complete list:

* [redux-thunk](https://github.com/gaearon/redux-thunk)
* [redux-promise](https://github.com/acdlite/redux-promise)
* [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware)
* [redux-saga](https://github.com/yelouafi/redux-saga)
* [redux-await](https://github.com/kolodny/redux-await)
* [redux-rx](https://github.com/acdlite/redux-rx)

jonase:
I think there are two issues here that should perhaps not be conflated.

1. **Boilerplate.** (with reducers, actions, action creators, action constants). [redux-actions](https://github.com/acdlite/redux-actions) could be an approach to remove some of that boilerplate. Using redux-actions would also "standardize" what actions look like within lumen.
2. **Async actions**. Reducers are synchronous but many actions are inherently async. One example is all client/server communications. Current solution is [redux-thunk](https://github.com/gaearon/redux-thunk), and if we use [redux-actions](https://github.com/acdlite/redux-actions) it will be relatively easy to move to either [redux-promise](https://github.com/acdlite/redux-promise), [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware), or [redux-rx](https://github.com/acdlite/redux-rx). In addition, odds are that "the next big thing" for async actions will have some kind of support for [redux-actions](https://github.com/acdlite/redux-actions).

The first step could be to reduce some of the boilerplate using redux-actions, and later evaluate if there is still a need to replace redux-thunk once some of the boilerplate is gone.

PR #306, opened 2016-07-05
[#263] Use redux-actions

jonase:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

* Less boilerplate with createAction and handleActions
* No need need the constants modules
* redux-promise (or redux-rx) relies on "standardized" actions provided by redux-actions


client/package.json
```diff
@@ -21,31 +21,33 @@
...
     "babel-preset-react": "^6.11.1",
     "css-loader": "^0.23.1",
+    "d3": "^3.5.17",
     "extract-text-webpack-plugin": "^1.0.1",
     "file-loader": "^0.9.0",
     "fixed-data-table": "^0.6.1",
     "history": "^3.0.0",
     "isomorphic-fetch": "^2.2.1",
-    "keycloak-js": "1.9.8",
+    "keycloak-js": "1.9.7",
```
gabemart: Is this intentional?

jonase: yes. We're using keycloak version 1.9.7 for local development. I suppose we'll use the same version for deployments later (ping @kardan, @iperdomo).

gabemart:
I'm finding that when I refresh on a saved dashboard in the dashboard editor (what I call "cold loading" a dashboard) some percentage of the time I get a blank "untitled dashboard"

Investigating...

gabemart:
Turns out the problem has nothing to do with redux-actions, it is just a pre-existing rare race condition bug #310

gabemart:
Code checked 👍 pushing two commits to this branch, please check and then merge

PR #312, opened 2016-07-06
Issue/262 immutable js and domain modules

jonase:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


client/src/containers/Dataset.jsx
```diff
@@ -49,31 +51,24 @@ class Dataset extends Component {
...
 }
 
 Dataset.propTypes = {
-  dataset: PropTypes.shape({
-    id: PropTypes.string.isRequired,
-    name: PropTypes.string.isRequired,
-    columns: PropTypes.array,
-    rows: PropTypes.array,
-  }),
+  dataset: PropTypes.object.isRequired,
```
gabemart: This gives an error 'Required prop `dataset` was not specified in `Dataset`.' when cold-loading a saved dataset in the dataset-editor

jonase: ah, right. It should probably not be `isRequired` as it previously was not.

client/src/reducers/transform.js
```diff
@@ -16,23 +16,21 @@ const availableTransforms = {
...
-  const transformations = (dataset.transformations || []).slice(0);
-  transformations.push(transformation);
-  return Object.assign({}, dataset, { transformations });
-}
-
 export default function applyTransformation(dataset, transformation) {
-  const transformedDataset = availableTransforms[transformation.op](dataset, transformation);
-  if (transformedDataset === dataset) {
+  const transformedDataset =
+    availableTransforms[transformation.get('op')](dataset, transformation);
```
gabemart: If you apply a "sort" to a column in the dataset editor, then try to cancel the sort by clicking the `x` button in the column header, it fails with the message "Transformation aborted: transformation.get is not a function". I think this line is the source of the error, but I'm not totally sure (I trust you can debug this more easily than I can)

client/src/components/dataset/sidebars/Filter.jsx
```diff
@@ -18,42 +18,42 @@ export default class Filter extends Component {
...
-    this.setState(Object.assign({}, this.state, { args: newArgs }));
   }
 
   render() {
     const { onClose, onApply, column } = this.props;
-    const { args } = this.state;
+    const { transformation } = this.state;
     const [expressionOperator, expressionValue] =
-      getExpressionOperatorAndValue(args.expression);
+      getExpressionOperatorAndValue(transformation.getIn(['args', 'expression']));
```
gabemart: This is perhaps a very small point, but `expressionValue` can now be `undefined` sometimes.

Because we feed this into the text input as its `value` on line 98, when this is `undefined` it triggers a react warning that we are switching the text input from controlled to uncontrolled

image

You can reproduce this by opening the Filter sidebar in the dataset editor and switching from "value is" to "value contains"

I don't know if it's intended that `expressionValue` can be undefined sometimes? If it is, you can suppress this warning simply by giving the text input a blank string when there is no `expressionValue` like `<input value={expressionValue || ''}`. If `expressionValue` shouldn't be undefined in the first place, I guess this is a bigger bug

client/src/components/dataset/sidebars/ChangeDataType.jsx
```diff
@@ -114,51 +115,39 @@ export default class ChangeDataType extends Component {
...
-        break;
-      }
-      default:
-        throw new Error(`Unkown error strategy ${errorStrategy}`);
-    }
+    const transformation = this.state.transformation;
+    const defaultValue = errorStrategy === 'empty-cell' ?
+      null : transformation.getIn(['args', 'defaultValue']);
+    this.setState({
+      errorStrategy,
```
gabemart: Transformations always default to "blank cell" on error, no matter what is chosen in the "if cell format is invalid" input

I think this is because we record the new error strategy in `this.state.errorStrategy`, but in `onApply` we only pass `this.state.transformation`, which always has an `onError` value of `default-value`

gabemart:
@jonase the DeleteConfirmationModal component is still trying use the old representation of datasets, so isn't working - it displays incorrectly, and cascading delete of dependent visualisations also isn't triggered

gabemart:
@jonase line 120 in DatasetTable - `if (newColumnType !== column.type) { ` - I suspect this should be `column.get('type')`

gabemart:
Similarly, line 25 in DataTypeContextMenu `selected={column.type}` -> `selected={column.get('type')}`

gabemart:
Additionally, there are non-immutable calls to dataset in DatasetSettings.jsx and in Settings.jsx - but I don't think these are really problems, as they are only fed immutable datasets when you click the "Settings" button in the dataset editor, which doesn't work anyway even in develop ^_^

client/src/reducers/transforms/sortColumn.js
```diff
@@ -32,46 +31,50 @@ function composeComparators(comparators) {
...
   const sortLevels = columns
-    .map(column => column.sort)
+    .map(column => column.get('sort'))
     .filter(n => n != null);
-  if (sortLevels.length === 0) {
+  if (sortLevels.size === 0) {
     return 1;
   }
-  return sortLevels[sortLevels.length - 1] + 1;
+  return sortLevels.get(-1) + 1;
```
gabemart: It might just be my unfamiliarity with immutable.js, but what does `sortLevels.get(-1)` do?

jonase: yeah, this might be a bit too clever :) Negative indexes are allowed and counting from the end of the list. So index -1 is the last element in the vector.

gabemart:
👍 looks good to me

Issue #311, opened 2016-07-06
Dashboard Editor requests datasets multiple times on cold load


gabemart:
When a dashboard is cold-loaded, `onAddVisualisation()` is called once for each visualisation in the dashboard. This sends a request for that visualisation's dataset if the dataset is not yet loaded, which is necessary to populate the dashboard editor properly. But if two or more visualisations are based on the same dataset, that dataset is requested once for each visualisation, which is clearly incorrect.

To fix, we need to store which datasets we have already requested.

PR #313, opened 2016-07-06
[#311] Prevent multiple API requests for the same dataset or dashboard

gabemart:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

#311

### DIFF
--- a/client/src/components/dashboard/DashboardEditor.jsx
+++ b/client/src/components/dashboard/DashboardEditor.jsx
@@ -84,6 +84,14 @@ export default class DashboardEditor extends Component {
     this.handleSave = this.handleSave.bind(this);
   }
 
+  componentWillMount() {
+    if (this.props.dashboard.layout) {
+      this.setState({
+        propLayout: this.props.dashboard.layout,
+      });
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


jonase:
This seems like a very general problem. Instead of adding logic to the different components (in the lifecycle methods, making them less declarative) I wonder if we could take a more general approach? Perhaps an `inFlight` key on the entities to keep track of in-progress api requests?

So for example [`fetchDataset(id)`](https://github.com/akvo/akvo-lumen/blob/develop/client/src/actions/dataset.js#L38-L49) would check if this request is already in progress, something like 

```js
export function fetchDataset(id) {
    if (inFlight(getState().library.datasets[id])) {
      return;
    }
    ...
}
```

and add reducers for `fetchDataset{Request|Failure|Success}`:

```js
function datasetFetchRequested(id) {
  // switch the inFlight key to true
}

function datasetFetchSuccess(dataset) {
  // update the dataset, and set inFlight key to false
}
```

jonase:
👍 LGTM. After discussions more general solutions to `inFlight` requests and if all data dependencies are loaded for a given entity will go in different issues.

Issue #300, opened 2016-06-24
Code Splitting PoC


gabemart:
We should used [code splitting](https://webpack.github.io/docs/code-splitting.html) to avoid delivering the entire app on first-load. This will improve first-load performance.

This issue will track a proof-of-concept implementation of code splitting for lumen.

gabemart:
I suggest the following approach:

# Application

I suggest we split the three main containers - `Dataset`, `Visualisation` and `Dashboard` - into separate bundles. They are well isolated from each other, so there isn't a lot of code duplication between bundles. It seems likely that at least some users will only need to access one or two of these containers in a given session, so loading on-demand seems like a performance win.

I also suggest that, because the dependencies for visualisations themselves are quite large and are needed in both `Visualisation` and `Dashboard`, we split these into a separate `ChartPreload` bundle. This bundle is pre-loaded before loading either the `Visualisation` or `Dashboard` containers. This means that in either the visualisation or dashboard editor, all visualisations are guaranteed to be immediately available, but we avoid loading the visualisation dependences when loading the Dataset editor, where no visualisations are needed.

# Public

For public visualisations and dashboards, I suggest we split by visualisation dependency, so that `rd3` is bundled with `DashBarChart`, `DashPieChart` etc. and `reactLeaflet` is bundled with `DashMap`.

(I still need to remove the outdated "Dash" prefix from these components, but that will be easiest to do once the initial code-splitting PR is merged)

Given that shared visualisations and dashboards will be viewed on a much wider range of devices and connection speeds than the main application, performance is critical here, so I think it makes sense to have separate bundles for separate visualisation libraries, whereas in the main application we bundle all visualisation dependencies together.

# Bundle sizes
````
current, dev
============
# app
app.bundle.js  4.34 MB

# public
pub.bundle.js  1.81 MB

current, production (minified)
=============
# app
app.bundle.js  1.69 MB

# public
pub.bundle.js   685 kB

Code-splitting, dev
==============
# app
app.bundle.js                   2.85 MB
1.DashChartPreload.bundle.js    909 kB
2.Visualisation.bundle.js       49.4 kB
3.Dataset.bundle.js             304 kB
4.Dashboard.bundle.js           244 kB
//                              -------
// TOTAL                        4.16 MB

# public
pub.bundle.js                   793 kB
6.reactLeaflet.bundle.js        671 kB       
7.rd3.bundle.js                 640 kB
//                              -------
// Total                        2.01MB
[...]

PR #319, opened 2016-07-15
[#300] Initial code-splitting for containers

gabemart:
#300 

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/utilities/chart.js
+++ b/client/src/utilities/chart.js
@@ -1,11 +1,6 @@
-import Immutable from 'immutable';
-
 export function getChartData(visualisation, datasets) {
   const { datasetId, spec } = visualisation;
-  // Test to see if the dataset is already an Immutable object
-  const dataset =
-    Immutable.Iterable.isIterable(datasets[datasetId]) ?
-      datasets[datasetId] : Immutable.fromJS(datasets[datasetId]);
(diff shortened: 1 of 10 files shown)


client/src/utilities/chart.js
```diff
@@ -1,6 +1,11 @@
+import Immutable from 'immutable';
+
 export function getChartData(visualisation, datasets) {
   const { datasetId, spec } = visualisation;
-  const dataset = datasets[datasetId];
```
jonase: I think the dataset should always be an Immutable object. Is this not always the case?

gabemart: It's currently not the case when viewing a shared visualisation or dashboard - as far as I can see these functions are both broken in develop as a result

jonase: ok, I think we should add the `fromJS` there then.

gabemart: "There" as in the public index? Yep, that makes sense, will change.

jonase:
@gabemart Looks good, I'll merge this.

Some follow up questions for my own understanding

* Did you follow https://webpack.github.io/docs/code-splitting.html or did you have some other reference as well?
* Do you know how and if code splitting will change for es6 modules once webpack 2 is released?

PR #320, opened 2016-08-08
[#58] Update client deps

jonase:
* Biggest change is that eslint now warns on string refs which we've been
  using quite a lot.
* https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute


gabemart:
Code checked and tested 👍

Issue #321, opened 2016-08-11
Prevent unnecessary re-renders in Dashboard editor


gabemart:
Related to the performance of #317 but it makes sense to split this out as a separate issue

Renders in the dashboard editor are expensive. We should prevent performing a new render when nothing of importance has changed

PR #322, opened 2016-08-12
[#321] Only rerender dashboard items when dimensions changed

gabemart:
#321

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/components/dashboard/DashboardCanvasItem.jsx
+++ b/client/src/components/dashboard/DashboardCanvasItem.jsx
@@ -33,13 +33,14 @@ export default class DashboardCanvasItem extends Component {
     const newLayout = this.getItemLayout(nextProps);
     const dimensionsUnchanged = oldLayout.w === newLayout.w && oldLayout.h === newLayout.h;
     const canvasWidthUnchanged = this.props.canvasWidth === nextProps.canvasWidth;
-    const datasetLoaded = this.getIsDatasetLoaded(this.props);
+    const needDataset = this.props.item.type === 'visualisation';
+    const datasetDependencyMet = needDataset ? this.getIsDatasetLoaded(this.props) : true;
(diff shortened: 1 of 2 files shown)


client/src/components/dashboard/DashboardCanvasItem.jsx
```diff
@@ -28,12 +28,33 @@ export default class DashboardCanvasItem extends Component {
...
+    const styleTransitionFinished = this.state.style.boxShadow === 'none';
+
+    const shouldNotUpdate = Boolean(
+        dimensionsUnchanged &&
+        canvasWidthUnchanged &&
+        datasetDependencyMet &&
+        styleTransitionFinished
+      );
+
+    return !shouldNotUpdate;
```
jonase: the expression `!shoultNotUpdate` reads a bit strange to me as it would be equivalent to `shouldUpdate`?. I wonder if it's possible to rewrite the boolean logic above to avoid the double negation?

gabemart: Yeah, this felt a bit funny to me too as I was writing it, but in the end I decided it was better to have one negation in the return statement than multiple in the test. But coming back to it, I suppose it might be clearer like:

````
+    const shouldUpdate = Boolean(
+        dimensionsChanged ||
+        canvasWidthChanged ||
+        !datasetDependencyMet ||
+        !styleTransitionFinished
+      );
+
+    return shouldUpdate;
````
I would be reluctant to change e.g. `datasetDependencyMet` to `datasetDependencyNotMet` to further avoid negation as that feels even more confusing to me

What do you think?

jonase: I think that approach is clearer 👍

jonase:
:+1: LGTM

Issue #337, opened 2016-08-25
Add migrations to backend init


kardan:
We should run migrations on app init to make sure we progress the db on
deployments.

PR #338, opened 2016-08-25
[#337] Migrate on init

kardan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

We should run migrations on app init to make sure we progress the db on
deployments. Bindings had to be moved to a separate namespace to
circumvent circular dependencies.


jonase:
👍 lgtm

Issue #326, opened 2016-08-22
Add dev helper functions to query API from the command line


kardan:
Since the API is protected, requests to the protected parts of the API needs a authorisation header containing a token from the Keycloak server in use. We should add helper functions we have locally to the repo for future developers. 

- [x] Bash (jonas)
- [x] Fish (daniel)

kardan:
Fixed in https://github.com/akvo/akvo-lumen/pull/347

PR #347, opened 2016-08-30
[#326] Add fish functions to call API from terminal

kardan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

This is the hipster version, using fish and httpie.


iperdomo:
Is not more _standard_ to use bash for this? just asking

kardan:
@jonase have some bash functions to add to. But since I use Fish I thought that there is no downside to add these to.

PR #351, opened 2016-09-07
Issue/260 domain logic

kardan:
- [ ] Copyright header
- [x] Code formatting
- [x] Documentation


backend/src/org/akvo/lumen/endpoint/dashboard.clj
```diff
@@ -1,183 +1,33 @@
 (ns org.akvo.lumen.endpoint.dashboard
   (:require [compojure.core :refer :all]
-            [clojure.java.jdbc :as jdbc]
-            [hugsql.core :as hugsql]
             [org.akvo.lumen.component.tenant-manager :refer [connection]]
-            [org.akvo.lumen.util :refer [squuid]]
-            [ring.util.response :as resp]))
+            [org.akvo.lumen.lib.dashboard :as dashboard]
```
jonase: @kardan You chose `lib` instead of `domain`, but we still tend to talk about "domain" in our chats/calls.. Is there a specific reason you prefer this terminology?

backend/src/org/akvo/lumen/endpoint/dashboard.clj
```diff
@@ -1,183 +1,33 @@
...
 
       (context "/:id" [id]
 
         (GET "/" _
-          (if-some [dashboard (dashboard-by-id tenant-conn {:id id})]
-            (resp/response (handle-dashboard-by-id tenant-conn  id))
-            (resp/not-found {:error "Not found"})))
+          (if-let [d (dashboard/fetch tenant-conn id)]
+            (response d)
+            (not-found {:error "not found"})))
```
jonase: @kardan Is `not-found` an error? In some cases we return 404 + the `id` that wasn't found (for example in job execution). Not sure which is preferable? Perhaps the redundant error string `"not found"` is telling?

backend/src/org/akvo/lumen/endpoint/dataset.clj
```diff
@@ -1,66 +1,27 @@
...
 
       (POST "/" {:keys [tenant body jwt-claims] :as request}
-        (import/handle-import-request tenant-conn config jwt-claims body))
+        (response (dataset/create tenant-conn config jwt-claims body)))
 
       (context "/:id" [id]
         (GET "/" _
-          (if-let [dataset (find-dataset tenant-conn id)]
-            (response dataset)
-            (not-found {:id id})))
```
jonase: @kardan like this change from `{:id id}` to `{:error "not found"}`. I would argue that "not found" is implicit in the response code

backend/src/org/akvo/lumen/endpoint/flow.clj
```diff
@@ -1,44 +1,18 @@
...
-                      (format flow-report-database-url org-id)
-                      {:folder-ids root-ids}
-                      {}
-                      {:identifiers identity})))
+      (let [items (flow/folders-and-surveys (:jwt-claims request)
+                                            flow-report-database-url
+                                            (-> request :params :org-id))]
+        (if items
+          (response items)
           (response ()))))
```
jonase: @kardan this is outside the change set, so probably ignore for now :)

But I think every endpoint/domain that returns a collection should never return 404/`nil` as this one apparently does. Much better IMO to just return an empty collection.

backend/src/org/akvo/lumen/lib/dashboard.clj
```diff
@@ -0,0 +1,29 @@
...
+  (:require [org.akvo.lumen.lib.dashboard-impl :as impl]))
+
+
+(defn all
+  "Returns all dashboards."
+  [tenant-conn]
+  (impl/all-dashboards tenant-conn))
+
+(defn create
+  "Creates a new dashboard. Body should include keys and spec conform
```
jonase: @kardan This docstring is still wip 👍 

I still wanted to mention that `"Body"` perhaps does not make sense in this context. Perhaps better (when we have clojure.spec/schema) to point to the var holding the spec definition. Something like `spec must conform with org.akvo.lumen.lib.dashboard/spec`

When I see `body` in a docstring I always think of macros, where body would be one or more expressions

backend/src/org/akvo/lumen/lib/dashboard_impl.clj
```diff
@@ -0,0 +1,129 @@
...
+(defn dashboard-keys-match?
+  "Make sure each key in entity have a matching key in layout."
+  [dashboard]
+  (= (keys (get dashboard "entities"))
+     (keys (get dashboard "layout"))))
+
+(defn filter-type
+  [dashboard-data kind]
+  (let [entities (filter #(= kind (get % "type"))
+                                       (vals (get dashboard-data "entities")))
```
jonase: @kardan indent seems off here?

backend/src/org/akvo/lumen/lib/dashboard_impl.clj
```diff
@@ -0,0 +1,129 @@
...
+  for each visualiation included in the dashboard insert an entry into
+  dashboard_visualisation with it's layout data."
+  [tenant-conn spec]
+  (if (dashboard-keys-match? spec)
+    (let [dashboard-id (str (squuid))
+          parted-spec  (part-by-entity-type spec)
+          visualisation-layouts (get-in parted-spec [:visualisations :layout])]
+      (jdbc/with-db-transaction [tx tenant-conn]
+        (insert-dashboard tx {:id    dashboard-id
+                                   :title (get spec "title")
```
jonase: @kardan indent fix?

backend/src/org/akvo/lumen/lib/dashboard_impl.clj
```diff
@@ -0,0 +1,129 @@
...
+  "We update a dashboard via upsert of dashboard and clean - insert of
+  dashboard_visualisations.
+  1. Unpack text & visualisation entities
+  2. Update dashboard tablefq
+  3. Remove old dashboard_visualisations
+  4. Add new dashboard_visualisations
+  "
+  [tenant-conn id spec]
+  (let [{texts          :texts
+         visualisations :visualisations} (part-by-entity-type spec)
```
jonase: @kardan If you don't rename the keys in the destructuring, perhaps prefer `{:keys [texts visualisations]}`?

backend/src/org/akvo/lumen/lib/dashboard_impl.clj
```diff
@@ -0,0 +1,129 @@
...
+        visualisations-layouts           (:layout visualisations)]
+    (jdbc/with-db-transaction [tx tenant-conn]
+      (update-dashboard tx {:id    id
+                                 :title (get spec "title")
```
jonase: @kardan indent off?

backend/src/org/akvo/lumen/lib/dataset_impl.clj
```diff
@@ -0,0 +1,56 @@
...
+            [org.akvo.lumen.endpoint.job-execution :as job-execution]))
+
+
+(hugsql/def-db-fns "org/akvo/lumen/lib/dataset.sql")
+(hugsql/def-db-fns "org/akvo/lumen/job-execution.sql")
+
+
+(defn create [tenant-conn config jwt-claims body]
+  (let [i (import/handle-import-request tenant-conn config jwt-claims body)]
+    ;; i is {:status 200, :headers {}, :body {"importId" "57cffd3f-35ab-4bd0-b09b-b4452d84451d"}}
```
jonase: @kardan this comment can probably be removed?

backend/src/org/akvo/lumen/lib/visualisation_impl.clj
```diff
@@ -0,0 +1,51 @@
...
+                                 :dataset-id (get body "datasetId")
+                                 :type (get body "visualisationType")
+                                 :name (get body "name")
+                                 :spec (get body "spec")
+                                 :author jwt-claims})]
+    {:id (-> v first :id)}))
+
+(defn delete [tenant-conn id]
+  (let [c (delete-visualisation-by-id tenant-conn {:id id})]
+    (when (not (zero? c))
```
jonase: `(when (not ...))` => `(when-not ...)`

backend/src/org/akvo/lumen/endpoint/dashboard.clj
kardan: I'm not sure I have a good answer to this. Domain looked odd to me. Suggesting that files outside of that was not the domain or that the domain dir contained different domains, like endpoints or components. 

I could argue that we have domain logic the files at src/org/akvo/lumen/* & in components to. But maybe auth.clj, import.clj, job-execution.clj & transformation.clj should move into the domain logic dir (currently lib) to. 

So in short I don't have a good answer, and I'm not married to "lib" I was actually considering "lumen" and "api" to. But i would be fine with something else.

jonase: @kardan I'm fine with `lib`, just wanted to hear the rationale

backend/src/org/akvo/lumen/lib/dashboard_impl.clj
kardan: 👍

backend/src/org/akvo/lumen/lib/dashboard_impl.clj
kardan: 👍

backend/src/org/akvo/lumen/lib/dashboard.clj
kardan: Agree, I'll remove that section. We should add the information once we have something.

backend/src/org/akvo/lumen/endpoint/flow.clj
kardan: Agree

jonase:
👍

PR #354, opened 2016-09-08
[#58] Update client libs

jonase:
I was unable to update react-hot-loader to `3.0.0-beta.3` because it broke hot reloading. But since 3.0.0 is still on beta it's just as well that we wait for a proper release


kardan:
👍

PR #375, opened 2016-09-28
Issue/361 transformation refactor

jonase:
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

### DIFF
--- a/backend/src/org/akvo/lumen/component/transformation_engine.clj
+++ b/backend/src/org/akvo/lumen/component/transformation_engine.clj
@@ -4,7 +4,8 @@
   (:import [java.util.concurrent LinkedBlockingQueue]))
 
 (defprotocol TransformationQueue
-  (enqueue [this command] "Enqueue a command (transformation or undo) to the transformation engine"))
+  (enqueue [this command]
+    "Enqueue a command (transformation or undo) to the transformation engine"))
 
 (defrecord TransformationEngine [queue running? sentinel]
   component/Lifecycle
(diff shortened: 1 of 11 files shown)


backend/src/org/akvo/lumen/component/transformation_engine.clj
```diff
@@ -11,25 +11,26 @@
...
+                     (try
+                       (when-not (identical? transformation sentinel)
+                         (let [{:keys [completion-promise tenant-conn job-id dataset-id command]} transformation]
+                           (condp = (:type command)
+                             :transformation
+                             (engine/execute-transformation completion-promise
+                                                            tenant-conn
+                                                            job-id
+                                                            dataset-id
+                                                            (:transformation command))
```
kardan: Is this line a misstake?

backend/src/org/akvo/lumen/transformation/engine.clj
```diff
@@ -198,95 +198,96 @@
...
-                                     :dataset-version-id new-dataset-version-id
-                                     :dataset-id dataset-id})))
+          source-table (:table-name dataset-version)]
+      (let [{:keys [success? message columns execution-log]} (apply-operation tenant-conn
+                                                                              source-table
+                                                                              columns
+                                                                              transformation)]
+        (if success?
+          (let [new-dataset-version-id (str (util/squuid))]
+            (clear-dataset-version-data-table tenant-conn {:id (:id dataset-version)})
```
kardan: Don't the db calls here need to be in a transaction?

backend/src/org/akvo/lumen/component/transformation_engine.clj
jonase: Not a mistake, but I can change it if we choose to.

I usually do either

```
(cond
  test-a expr-1
  test-b expr-2)
```

or

```
(cond 
  another-test-1-perhaps-long-and-complicated
  another-expression-also-possibly-a-big-expression

  test-2
  long-and-complicated-expression-for-test-2)
```

kardan: I don't mind either but the second arg on new line I find a bit difficult to parse without counting parens. Like:
```
(cond test-a
  test-b)
```

Issue #447, opened 2016-12-07
Add reminder to update release notes to PR template


jonase:
## Context

We want to keep the `RELEASE_NOTES.md` file up-to-date in the `develop` branch so we don't have to spend effort writing it when we do a release.

## Solution

Add a reminder about updating the release notes to the PR template.

PR #448, opened 2016-12-07
[#447] Add release notes to PR template

jonase:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-
+- [ ] Update release notes
 - [ ] Test plan | Unit test | Integration test
 - [ ] Copyright header
 - [ ] Code formatting


punchagan:
We (RSR team) have started doing this, and it's pretty useful when trying to get release notes before a release. 

One problem that we face, though, is that PRs tend to have conflicts a lot more than before because of the  RELEASE_NOTES.md file.  Not sure there's a good way to work around this.

jonase:
@punchagan Thanks, Good point. 

These merge conflicts should be easy to resolve since we're only adding lines, never changing what someone else previously added so the merge strategy should be trivial.

So if we merge develop into our feature branch before we open the PR we will often see a merge conflict in `RELEASE_NOTES.md` but my hope is that with some experience we'll be able to quickly resolve these.

(Maybe there's even a `git merge -s <strategy>` to only add new lines?)

punchagan:
If PRs are getting merged quickly after they are opened, it shouldn't be a big problem. But, often when a handful of PRs are open, some of them conflict the others, and they can't be merged from the GitHub UI without first fixing the PR.

jonase:
@punchagan yeah, I can see that. Let's see if it causes more problems than it solves. I can't think of a good solution right now, but I like the idea that the person who's in charge of the feature/bug-fix updates the release notes.

Are you thinking of doing something else in RSR?

punchagan:
Yes, it is indeed a very nice idea and I like it too.  But, it's sometimes annoying when you are reviewing a PR and you have to resolve this 'stupid' conflict.  Just on the look out for other ideas, to work around this.

PR #460, opened 2016-12-21
Feature/429 derived column backend

jonase:
- [x] Update release notes
- [ ] Copyright header
- [x] Code formatting
- [x] Documentation


kardan:
At the moment it's possible to deprive a column with the same title as a previous one. I guess this is problematic. When I tried it seems like the browser reloads itself at times.

jonase:
@kardan I can add a check for this

Issue #430, opened 2016-11-24
Visualisation UI implementation phase 1


mtwestra:

nadiagorchakova:
We've mapped out the bar chart workflow today and here's the resulting story map:

<img width="988" alt="screen shot 2016-12-16 at 12 02 56" src="https://cloud.githubusercontent.com/assets/21127166/21260993/17b7f3ee-c389-11e6-8a52-b52bb874b268.png">

It also references stacked chart functionality towards the end (related to https://github.com/akvo/akvo-lumen/issues/434)

We agreed that it's necessary to improve the bar chart workflow according to the map.

PR #469, opened 2016-12-29
[#430] visualisation editor rethink with sub-buckets for bar charts

gabemart:
- [x] Update release notes
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation


client/src/components/charts/MapVisualisation.jsx
```diff
@@ -25,8 +26,11 @@ export default function MapVisualisation({ visualisation, datasets, width, heigh
           attribution="&copy; <a href='http://osm.org/copyright'>OpenStreetMap</a> contributors"
         />
         {
-          chartData[0].values.map((entry, index) =>
-            <Marker position={entry.position} key={index}>
+          chartData.values.map((entry, index) =>
+            <Marker
+              position={[parseFloat(entry.x), parseFloat(entry.y)]}
```
jonase: At some point we should consider not parsing when rendering charts, as it's not really necessary if we only show numeric columns for properties that needs to be numeric.

jonase:
lgtm 👍

PR #499, opened 2017-01-16
[#352] Started fresh branch to fix #352. Updated component to allow i…

### LINKED ISSUE #352: Can't pick 0 as default value in column to number cast transformation
When converting a column to number and picking a default value, you can't pick "0", Is this expected?

peeb:
…nput of zero as default value.

- [x] Update release notes
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/components/dataset/sidebars/ChangeDataType.jsx
+++ b/client/src/components/dataset/sidebars/ChangeDataType.jsx
@@ -6,32 +6,23 @@ import SidebarControls from './SidebarControls';
 import DateFormatSelect from './DateFormatSelect';
 
 function DefaultValueInput({ defaultValue, onChange, newType }) {
+  const emptyValue = newType === 'number' ? 0 : '';
   return (
     <div className="inputGroup">
       <label htmlFor="defaultValueInput">
         Default value:
       </label>
       <input
-        type="text"
-        value={defaultValue || ''}
+        type={newType === 'number' ? 'number' : 'text'}
+        value={defaultValue !== null ? defaultValue : emptyValue}
         onChange={(event) => {
           const value = event.target.value;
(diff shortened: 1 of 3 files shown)


client/src/components/dataset/sidebars/ChangeDataType.jsx
```diff
@@ -6,32 +6,23 @@ import SidebarControls from './SidebarControls';
 import DateFormatSelect from './DateFormatSelect';
 
 function DefaultValueInput({ defaultValue, onChange, newType }) {
+  const emptyValue = newType === 'number' ? 0 : '';
```
gabemart: This gets fed back to `DefaultValueInput` as `defaultvalue` -  now that it can be a number, we should probably change the propType of `defaultValue` from `PropTypes.string` to `PropTypes.oneOfType([PropTypes.string, PropTypes.number])` to suppress the console warning when it's 0

peeb: ^^ didn't know about this. that's nice. Agreed we should definitely do that.

peeb: That would fix the "is not a NaN" console warning messages too, I'm guessing.

gabemart: @peeb I think if you wanted to fix the "is not a NaN" console warning, you would have to add an explicit NaN check on line 17 (i.e. `(defaultValue !== null && !isNan(defaultValue) ? ...`

peeb:
@gabemart This is clear to re-review when you have time

gabemart:
lgtm 👍

Issue #494, opened 2017-01-12
Fix Dev / Test environment


kardan:
# Context
Since we tried to move to the production cluster setup with an ingress the dev/test environment is not working any longer.

# Problem!

## Try to connect via https:
```
$ openssl s_client -connect dev.test.akvo-ops.org:443 -prexit                                                                                  dator.local:~/W/a/a/ingress ‹develop*›
CONNECTED(00000003)
5924:error:140790E5:SSL routines:SSL23_WRITE:ssl handshake failure:/BuildRoot/Library/Caches/com.apple.xbs/Sources/OpenSSL098/OpenSSL098-64.30.2/src/ssl/s23_lib.c:185:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 0 bytes and written 130 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
---
```
## Make sure TLS works in the cluster.
- [x] Verify cert. @orifito deployed new cert to cluster
- [x] Keycloak is running using the cluster cert (different Ingress). Hope this implies that the TLS is not the problem.

## Inspect Google load balancer
- [x] The load balancer reports the backend servers to be unhealthy.
- [x] Made sure appropriate ports (nodePorts service) are open.

## Inspect the service
- [x] Made sure the container is running on expected port.
- [x] Make sure service is running and accessible on expected port
 
# Solution or next step
- [x] Ponder
- [x] Try stuff
- [x] Ponder
- [x] Fix
- [x] Update DNS to point to correct IP
- [x] Make Lumen use https://kc.akvotest.org Keycloak

kardan:
In the end it worked if we: 
- added all nodePorts ports on all backend services to the correct firewall rule
- manually (outside of k8s) added a https route and added the cert.

My guess is that earlier when we had a service of type loadBalancer the correct ports was added automagically. When using an ingress object we need to add correct ports to the firewall by ourselves. It's still unclear why the https & cert is not picked up for the lumen ingress. As it is right now, all k8s ingresses reports that they have broken ssl certs. But ssl still works and side stepping k8s (kubectl) & looking via gcloud, the cert seem fine.... so a bit unclear what's going on there.

PR #509, opened 2017-01-18
[#494] Update deployment to use keycloak url value from config

kardan:
Also changed ingress to point to new host

- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


kardan:
Turns out we need to fix https://github.com/akvo/akvo-lumen/issues/510 to make this a runtime config

Issue #470, opened 2017-01-02
Next iteration of maps using Leaflet library (estimate: 64)


nadiagorchakova:
# Context
As part of delivery to UNC/Hilton, we're supposed to visualise data points on a map.
Leaflet library that we use right now seems quite capable and will be able to meet partners' expectations.

# Problem or idea
The map should enable marker, colouring and configurable popups (with an image). Image will be linked via a url.
The basic UI for this iteration should contain column selection for the popup window. A mockup would help. 

# Solution or next step

nadiagorchakova:
To sum up our mockup review today:

- custom colour coding can possibly be done using default browser colour coding option. Needs to be researched if it will work well for all browsers ( @gabemart ). In the next iteration we can think how to improve that. 
- auto-scaling of the map will be part of this iteration too. 
- images in popup windows are not present in the mockup but are meant to be part of this iteration

PR #519, opened 2017-01-25
Issue/470 maps interface

gabemart:
#470 

- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


client/src/components/charts/MapVisualisation.jsx
```diff
@@ -1,18 +1,202 @@
...
+DisplayLayer.propTypes = {
+  layerData: PropTypes.object.isRequired,
+};
 
 export default function MapVisualisation({ visualisation, datasets, width, height }) {
-  const chartData = chart.getChartData(visualisation, datasets)[0];
+  let tileLayerUrl;
+  let tileLayerAttribution;
+
+  const processedLayerArray = [];
```
jonase: Building this "layer data" could perhaps be split out into a separate function so you would in this render function only do `processedLayerArray = visualisation.spec.layers.map(getLayerData);` or something similar

client/src/components/charts/MapVisualisation.jsx
```diff
@@ -1,18 +1,202 @@
...
+          legend: layer.legend,
+          pointSize: layer.pointSize,
+        });
+      }
+    }
+  });
+
+  const bounds = calculateBounds(processedLayerArray);
+
+  switch (visualisation.spec.baseLayer) {
```
jonase: Same here. Split into a separate function perhaps?

`const { tileLayerUrl, tileLayerAttribution } = baseLayer(visualisation.spec.baseLayer)` or similar

client/src/components/charts/MapVisualisation.jsx
gabemart: I agree this would make sense as a separate function. But I think it might be better use something like `getProcessedLayerArray` rather than mapping the spec layers, because in some cases we don't want to include a layer from the spec in the display layers at all, e.g. when it has incomplete position data, or it's hidden. I'll extract a function on this basis and see how it goes

client/src/components/charts/MapVisualisation.jsx
gabemart: That's a really nice, clear, concise syntax for this 👍

PR #539, opened 2017-02-06
Issue/455 admin console

peeb:
- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


kardan:
@jonase Before this is merged we should manually massage existing users and add appropriate groups .

backend/src/akvo/lumen/auth.clj
```diff
@@ -2,35 +2,61 @@
...
+(defn tenant-admin?
+  [{:keys [tenant jwt-claims]}]
+  (contains? (claimed-roles jwt-claims)
+             (format "akvo:lumen:%s:admin" tenant)))
+
+(defn public-path? [{:keys [path-info request-method]}]
+  (and (= :get request-method)
+       (or (= "/api" path-info)
+           (= "/env" path-info)
+           (s/starts-with? path-info "/s/")
```
jonase: I think something happened with the merge from `develop` here as this should be `/share/`

README.md
```diff
@@ -14,5 +14,26 @@ An open-source, easy to use data mashup, analysis and publishing platform.
...
+```
+
+Available users are:
+
+- "lumen" service account
+- "jerome" t1 admin
+- "salim" t1 user
+- "ruth" t2 admin
+- "harry" t2 user
+- kaj keycloak user not on a tenant
```
jonase: quote "kaj" as the rest?

RELEASE_NOTES.md
```diff
@@ -71,6 +71,13 @@ Date 2017-01-25
 
 * Fixed regression where sharing and flow import where broken
 
+
+### New and noteworthy
```
jonase: This note needs to be moved up to v. 0.6.0

backend/resources/akvo/lumen/migrations/tenants/010-invites.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE invite (
+     id text PRIMARY KEY DEFAULT gen_random_uuid()::text,
+     email text NOT NULL,
+     expiration_time timestamptz,
+     consumed_at timestamptz,
```
jonase: I was thinking about the naming here with `expiration_time` vs. `consumed_at`? Is there an inconsistency? Perhaps `expiration_at`? If these differ in naming I could be fooled to believing that expiration_time would be an integer (in `ms`, `s` or something)? We also haven't added postfix (either `_at` or `_time`) to other `timestampz`'s.

I was also wondering what `consumed_at` means? I suppose it means time when the invitation was accepted? So we don't track if it was accepted/rejected/canceled? Only the time when one of these happened? (not that we currently can reject/cancel invitation)

backend/src/akvo/lumen/component/emailer.clj
```diff
@@ -0,0 +1,80 @@
...
+  (stop [this]
+    this)
+
+  ISendEmail
+  (send-email [this email]
+    (assert (email? email) "Not a proper email (:body :from :subject :to)")
+    (pprint email)))
+
+(defn dev-emailer
+  "Emailer will pprint email. Included link will be https."
```
jonase: > Included link will be https.

Is this still true?

backend/src/akvo/lumen/component/emailer.clj
```diff
@@ -0,0 +1,80 @@
...
+  (start [this]
+    this)
+  (stop [this]
+    this)
+
+  ISendEmail
+  (send-email [{config :config} email]
+    (postal/send-message config email)))
+
+(defn smtp-emailer
```
jonase: If we don't use the `smtp-emailer`, should we remove it (and the dependency on `postal`)?

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -0,0 +1,139 @@
...
+                          api-root group-id)
+                  request-options)
+      :body json/decode))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Helpers
+;;;
+
+;;; This is kinda fragile, this relies on that one user not belog to multiple
```
jonase: Can it be made less fragile? I.e. is it inherently fragile or WIP fragile?

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -0,0 +1,139 @@
...
+                              "id")
+          admins (map #(assoc % "admin" true)
+                      (group-members keycloak admin-group-id options))
+          members (filter #(and (get % "emailVerified") (get % "enabled"))
+                          (concat admins users))
+          response-filter ["admin" "email" "firstName" "id" "lastName"
+                           "username"]]
+      (response (map #(select-keys % response-filter) members)))
+    (catch Exception e
+      (prn e)
```
jonase: drop the `prn`'s?

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -0,0 +1,139 @@
...
+      (let [ed (ex-data e)]
+        (prn ed)
+        (response {:status (:status ed)
```
jonase: You catch `Exception` but assume it's an `ExceptionInfo` thrown from the http client. If it's another kind of exception `ex-data` will return null, and so will `(:status ed)`.

backend/src/akvo/lumen/endpoint/env.clj
```diff
@@ -3,5 +3,7 @@
             [ring.util.response :refer (response)]))
 
 (defn endpoint [{:keys [config]}]
-  (GET "/env" _
-    (response {"keycloakURL" (:keycloak-url config)})))
+  (GET "/env" request
+    (response {"keycloakClient" (:keycloak-public-client-id config)
```
jonase: Perhaps we should use `keycloakClient` on the client as well?

backend/src/akvo/lumen/endpoint/invite.clj
```diff
@@ -0,0 +1,30 @@
+(ns akvo.lumen.endpoint.invite
+  (:require [akvo.lumen.component.tenant-manager :refer [connection]]
+            [akvo.lumen.lib.invite :as invite]
+            [compojure.core :refer :all]
+
```
jonase: ws

backend/src/akvo/lumen/endpoint/invite.clj
```diff
@@ -0,0 +1,30 @@
...
+    (let-routes [tenant-conn (connection tenant-manager tenant)]
+      (GET "/" _
+        (invite/active-invites tenant-conn))
+
+      (POST "/" {:keys [body]}
+        (invite/create tenant-conn emailer keycloak
+                       (select-keys request
+                                    [:body :jwt-claims :server-name :server-port])))
+
+      #_(context "/:id" [id]
```
jonase: remove comment?

backend/src/akvo/lumen/endpoint/user.clj
```diff
@@ -0,0 +1,48 @@
+(ns akvo.lumen.endpoint.user
+  (:require [akvo.lumen.component.tenant-manager :refer [connection]]
+            [compojure.core :refer :all]
+            [akvo.lumen.component.keycloak :as keycloak]))
+
+
+;; Each tenant will already have one admin user
+;;
+;; The following endpoints are required for an MVP:
```
jonase: this looks like a temporary comment?

backend/src/akvo/lumen/endpoint/user.clj
```diff
@@ -0,0 +1,48 @@
...
+
+(defn endpoint [{:keys [keycloak tenant-manager]}]
+  (context "/api/admin/users" {:keys [jwt-claims params tenant] :as request}
+
+    (let-routes [roles (get-in jwt-claims ["realm_access" "roles"])]
+
+      (GET "/" _
+        (keycloak/users keycloak tenant)))))
+
+;; (defn user-invitation-endpoint [{:keys [tenant-manger]}]
```
jonase: rm this comment?

backend/src/akvo/lumen/lib/invite_impl.clj
```diff
@@ -0,0 +1,85 @@
...
+            [cheshire.core :as json]
+            [clj-time.coerce :as c]
+            [clj-time.core :as t]
+            [clojure.string :as str]
+            [hugsql.core :as hugsql]
+            [ring.util.response :refer [not-found response redirect]]))
+
+(hugsql/def-db-fns "akvo/lumen/lib/invite.sql")
+
+(defn dev-env? [server-name]
```
jonase: Why do we need `dev-env?`. Is it just to be able to pprint the full url? Perhaps it would be enough to just print the invite id?.

backend/src/akvo/lumen/lib/invite_impl.clj
```diff
@@ -0,0 +1,85 @@
...
+  (.contains '("t1.lumen.localhost" "t2.lumen.localhost") server-name))
```
jonase: `(#{"t1.lumen.localhost" "t2.lumen.localhost"} server-name)`?

backend/src/akvo/lumen/lib/invite_impl.clj
```diff
@@ -0,0 +1,85 @@
...
+          (first (insert-invite tenant-conn {:email email
+                                             :expiration_time expiration-time
+                                             :author claims}))
+          email {:body (email-body request id)
+                 ;; :from "noreply@akvolumen.org"
+                 :from "devops@akvo.org"
+                 :subject "Akvo Lumen invite"
+                 :to email-address}]
+      (emailer/send-email emailer email))
+    (prn (format "Tried to invite non existing user with email (%s)" email))))
```
jonase: Should we just `prn` here or return a proper status code+body?

backend/src/akvo/lumen/lib/invite_impl.clj
```diff
@@ -0,0 +1,85 @@
...
+  "Returns all active invites"
+  [tenant-conn]
+  (response (select-active-invites tenant-conn)))
+
+(defn create
+  "Creates a new invite"
+  [tenant-conn emailer keycloak request]
+  ;; If existing user and no other active invite on same email
+  (do-create-invite tenant-conn emailer keycloak request)
+  (response {:invite "created"}))
```
jonase: if `do-create-invite` fails we still return `(response {:invite "created"})`.

Would it be better to return 201 instead of 200?

backend/src/akvo/lumen/lib/invite_impl.clj
```diff
@@ -0,0 +1,85 @@
...
+      (if (dev-env? server-name)
+        (redirect (format "http://%s:3030" server-name))
+        (redirect "/"))
+      (-> (response
+           {:body (format "<html><body>%s</body></html>"
+                          "Problem completing your invite.")
+            :status 422})
+          (assoc :headers {"Content-Type" "text/html"})))
+    (do
+      (prn (format "Tried to verify invite with id: %s" id))
```
jonase: drop `prn`?

client/default.conf
```diff
@@ -35,18 +34,31 @@ server {
...
-    location /env {
+   location /verify {
+        if ( $request_method !~ ^(GET)$ ) {
+           return 405;
+        }       
         proxy_set_header        Host $host;
         proxy_set_header        X-Real-IP $remote_addr;
         proxy_set_header        X-Forwarded-Host $host;
-        proxy_pass http://localhost:3000/env;
+        proxy_pass http://localhost:3000/s;
```
jonase: proxy to `/verify` instead of `/s`?

backend/src/akvo/lumen/auth.clj
kardan: Yes, merge errors. Should be fixed now.

README.md
kardan: 👍

RELEASE_NOTES.md
kardan: 👍

backend/resources/akvo/lumen/migrations/tenants/010-invites.up.sql
kardan: Agree, fixing.

backend/src/akvo/lumen/component/emailer.clj
kardan: No, removing the https comment.

client/src/components/modals/InviteUser.jsx
```diff
@@ -0,0 +1,61 @@
...
+          <h2 className="modalTitle">Invite User</h2>
+          <div
+            className="close clickable"
+            onClick={onClose}
+          >
+          +
+          </div>
+          <div className="contents">
+            <p>Please enter the email address you would like to invite below.</p>
+            <form onSubmit={onInviteUser}>
```
jonase: @peeb This is problematic as it will result in a page refresh, which I don't think we want/need here.

There are a couple of ways resolve this.

1. Quick fix is to add `preventDefault` to `onInviteUser`
2. We don't tend to use `<form ..>` , instead just handle the form state manually and have an ordinary `<button ..>` to "submit". There are scss for input fields and buttons to keep the styling consistent which I can look up if you pick this approach.

client/.sass-lint.yml
```diff
@@ -0,0 +1,56 @@
+####################
```
jonase: @peeb Have you discussed sass linting with @gabemart?

* How do I run this? I don't see any deps added to package.json. Do I need a global (npm) install or is it part of sass itself?
* Where are these rules from or did you write them specifically for lumen?
* If we want this in the CI, should we still use `output-file`?
* Is `max-warnings` what we want? Or should we do the same as with eslint, so a lint error results in a broken build?
* If there are lots of errors/warnings for existing sass files, should this be part of a separate PR, with the errors/warnings fixed?

peeb: This shouldn't be here anymore, precisely because it requires discussion with you and @gabemart - it's a leftover from me trying out sass linting. I'll remove it.

backend/src/akvo/lumen/component/keycloak.clj
kardan: Added docs re this. "Fragile" might have been a bit too harsh description.

backend/src/akvo/lumen/component/keycloak.clj
kardan: 👍

backend/src/akvo/lumen/endpoint/env.clj
kardan: Client should pick up the keycloak client form /env now

backend/src/akvo/lumen/endpoint/invite.clj
kardan: 👍

backend/src/akvo/lumen/endpoint/invite.clj
kardan: 👍

backend/src/akvo/lumen/endpoint/user.clj
kardan: 👍

backend/src/akvo/lumen/endpoint/user.clj
kardan: 👍

backend/src/akvo/lumen/lib/invite_impl.clj
kardan: "do-create-invite" will always return nil. This is by design. As a Lumen user you should probably not be able to query the system to see if users (emails) exists. Returning the status would however make UI's with more feedback possible.

backend/src/akvo/lumen/lib/invite_impl.clj
kardan: 👍

client/default.conf
```diff
@@ -35,18 +34,31 @@ server {
...
+        proxy_pass http://localhost:3000/env;
+    }
+
     location /share {
         proxy_set_header        Host $host;
         proxy_set_header        X-Real-IP $remote_addr;
         proxy_set_header        X-Forwarded-Host $host;
         proxy_pass http://localhost:3000/share;
-    }
+    }
```
jonase: trailing ws. There are other places in this file also I believe that doesn't show up in this diff

tip: https://github.com/jonase/.emacs.d/blob/master/init.el#L67 ;)

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -0,0 +1,133 @@
...
+(defn group-members
+  [{:keys [api-root credentials openid-config]} group-id request-options]
+  (-> (client/get (format "%s/groups/%s/members"
+                          api-root group-id)
+                  request-options)
+      :body json/decode))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Helpers
```
jonase: double `Helpers` comment? There's one above as well

client/default.conf
kardan: 👍

jonase:
:+1: LgtM

Issue #570, opened 2017-02-08
Exact match doesn't work on number type in data filter


nadiagorchakova:
# Context
The issue happened in UNC instance.
Setting 'exactly matches' filter does not result in any changes in the scatter plot.
When you use 'higher than' instead - it works fine.

<img width="1258" alt="screen shot 2017-02-08 at 11 59 01" src="https://cloud.githubusercontent.com/assets/21127166/22735510/4ab0e94a-edfa-11e6-909f-4927a9e0fead.png">

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

jonase:
fixed

PR #571, opened 2017-02-08
[#570] Fix filter "exactly matches" for numeric columns

jonase:
* We use `===` when comparing so we need to parse the test value to a float
  for numbers and to an int for dates.

- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


gabemart:
lgtm

Issue #600, opened 2017-02-17
% values in pie and donut charts 


nadiagorchakova:
# Context

Currently, it's only possible to see count for pie chart slices:

<img width="1277" alt="screen shot 2017-02-17 at 17 39 22" src="https://cloud.githubusercontent.com/assets/21127166/23074122/29753ade-f538-11e6-8469-877a66d41817.png">

# Problem or idea

Karolina suggested that sometimes it's more useful to see % of each slice in relation to the whole data sample.

# Solution or next step

Perhaps we should have an option in pie chart editor to pick whether pie chart should display count or % for each slice.

mtwestra:
Repeated in #608, so closing that one.

One option is to show the percentage value in brackets as part of what is shown on hover. So in the above example, we would show '687 (67.3%)'

gabemart:
This has actually been the intended behaviour all along -- the fact that it is not showing currently is due to a bug in a recent refactor of the pie chart spec.

gabemart:
Closed via #612

PR #612, opened 2017-02-21
[#600] Show percentage values in pie and donut chart hover labels

gabemart:
#600 

This also fixes a small visual bug where the highlight color for segments that are being hovered over would not show correctly when the cursor was over the label text part of the segment

- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,7 @@
 ### Bugfixes
 
 * Fix bug where pie and donut chart labels would not show percentage values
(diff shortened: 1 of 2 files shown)


jonase:
lgtm

mtwestra:
@gabe, if possible, I would reverse the display, so 

24% (21)

instead of 

21 (24%)

mtwestra:
@gabe, otherwise, works fine.

Issue #578, opened 2017-02-08
Ability to choose colours in pie charts


mtwestra:
# Context
At the moment, colours in pie charts are determined in a way that leads to unwanted results. As shown below, "Yes" and 'No' answers are shown as two different colours in two consecutive pie charts, which is very confusing.

<img width="509" alt="screen shot 2017-02-08 at 15 33 48" src="https://cloud.githubusercontent.com/assets/478954/22741634/26b69322-ee14-11e6-879c-d229f1c30384.png">

In addition, it is necessary for the colours to be coordinated across visualisations, so that the colours used in a map can be made the same as the colours used in a related piechart

# Solution or next step
There are two ways this could be approached:
1) a fixed (default) colour scheme could be applied to the unique values in answers, but only after they have been alphabetised. 
2) similar to a map, a user could specify a styling, in which she chooses the colour for each value present.

gabemart:
I think for now, simply sorting the aggregated data by the label field before rendering it will be a big improvement. This ensures that segments with identical labels will be mapped to the same colors, even across different pie charts, *so long as those pie charts have the same number of total labels, and the labels are identical*.

Longer term, the solution might well be to let user pick colors per-segment, but I don't think that's realistic to target for this iteration.

gabemart:
As an example - sorting the aggregated data is enough to ensure that the colors are consistent between the first two charts below. But adding a third label, as in the final chart, will upset the color scheme, even with sorting.

Another option would be to use some kind of hash on the label text to determine the segment color, which would ensure consistent mapping of label text to color - but in that case, there would be no guarantee against two or labels in the same chart mapping to very similar colors. 

![image](https://cloud.githubusercontent.com/assets/2589977/23410087/ed29d65c-fdc5-11e6-88f7-1053bfb6a76a.png)

![image](https://cloud.githubusercontent.com/assets/2589977/23410150/2a82db84-fdc6-11e6-907f-743845bb7463.png)

![image](https://cloud.githubusercontent.com/assets/2589977/23410179/3a135934-fdc6-11e6-8c6f-3c0030c0920b.png)

mtwestra:
@gabe, I agree that sorting and then assigning is probably the only realistic way at this moment.

However, I think we will need to move towards colour pickers for pretty much all visualisations, as colours are so full of meaning usually, that they can't really be assigned automatically. In addition, if you include empty data this would always come first (at least I guess so), which would give rise to differences between pie charts that include or exclude 'no data'. 

[...]

PR #624, opened 2017-02-28
[#578] Sort pie chart segments to get consistent colors

gabemart:
#578 

By sorting the aggregated data before feeding it to vega, we increase the chances that pie charts will have more consistent segment colors across charts

- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/utilities/chart.js
+++ b/client/src/utilities/chart.js
@@ -100,7 +100,7 @@ export function getLineData(visualisation, datasets) {
   const haveAggregation = visualisation.spec.metricAggregation != null;
   const yIndex = getColumnIndex(dataset, spec.metricColumnY);
   const xIndex = getColumnIndex(dataset, spec.metricColumnX);
-  const xAxisType = xIndex === -1 ? 'number' : dataset.get('columns').get(xIndex).get('type');
+  const xAxisType = xIndex === -1 ? 'number' : dataset.getIn(['columns', xIndex, 'type']);
   const rowFilter = filterFn(spec.filters, dataset.get('columns'));
 
   const valueArray = dataset.get('rows')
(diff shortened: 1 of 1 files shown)


client/src/utilities/chart.js
```diff
@@ -195,6 +198,7 @@ export function getPieData(visualisation, datasets) {
   const { datasetId, spec } = visualisation;
   const dataset = datasets[datasetId];
   const bucketIndex = getColumnIndex(dataset, spec.bucketColumn);
+  const bucketColumnType = dataset.get('columns').get(bucketIndex).get('type');
```
jonase: Instead of chained `get`s you can also use `getIn`: dataset.getIn(['columns', bucketIndex, 'type']);

jonase:
👍 lgtm

mtwestra:
@gabe, tested by uploading the following dataset:
first_col, second_col, third_col
yes, no, yes
yes, no, no
no, no,	
no, yes, maybe
yes,  , yes
no, no, yes

So first and second column only containing 'yes/no', and third column containing 'maybe' as extra. Missing items in second and third column.

Expected behaviour: pie charts of first and second column should show same colours for yes and no. The third column will show different colours, as 'maybe' is the first alphabetically.

Actual results: 
screen shot 2017-03-07 at 11 36 54
screen shot 2017-03-07 at 11 37 06
screen shot 2017-03-07 at 11 37 20

I think the second image shows something unexpected: the No data is inserted as a colour in between yes and no. I think 'No data' should be put as the last colour always? The final image is consistent with this.

gabemart:
@mtwestra The cause of this problem is that we were replacing empty values with the "empty value" string (`No data` currently) before sorting / applying the color indexes. To fix, we just need to compare against the empty value string rather than doing a null check

gabemart:
Color results are as expected after changes in https://github.com/akvo/akvo-lumen/pull/641


image

image

image

mtwestra:
@gabe, excellent

Issue #621, opened 2017-02-28
Pivot tables - allow user-defined names for columns


mtwestra:
### Context
Currently, when we create a pivot table, we use the column names as the name shown in the table. However, in many cases this is not ideal as the names often contain things like underscores.

### Proposed solution
Proposed solution is to add a custom name field for both row and column, in the same way as is done for the x and y axis for other visualisation types.

PR #634, opened 2017-03-06
[#621] pivot column names

gabemart:
#621 

I added some update logic so we only hit the `pivot` api when we expect the aggregated data to have changed. If we don't do anything, we'll hit the api on every `onChange` event for the text inputs for the new `rowTitle` and `categoryTitle` fields on the spec.

- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


mtwestra:
@gabe, tested in visualisation editor, dashboard editor and live dashboard, and works fine.

There is a small issue: when the values in the columns are small, than the first column is also narrow. However, this leaves no room for the column title. See below:
screen shot 2017-03-07 at 12 16 00

Proposed solution: apply minimum column width also to first column.

gabemart:
@mtwestra created https://github.com/akvo/akvo-lumen/pull/640 to fix this

PR #658, opened 2017-03-16
Issue/645 delete confirmation

peeb:
- [x] Update release notes
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/components/Users.jsx
+++ b/client/src/components/Users.jsx
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 import EntityTypeHeader from './entity-editor/EntityTypeHeader';
+import ConfirmUserAction from './modals/ConfirmUserAction';
 import InviteUser from './modals/InviteUser';
 import * as api from '../api';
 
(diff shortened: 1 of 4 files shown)


peeb:
@gabemart I think the <select> dropdown needs some styling. Comments on that are especially welcome.

client/src/components/Users.jsx
```diff
@@ -1,64 +1,55 @@
...
+function UserActionSelector({ user, onChange }) {
+  const { active, admin } = user;
+  return (
+    <select
+      className="UserActionSelector"
+      onChange={event => onChange(user, event.target.value)}
+      value="?"
+    >
+      <option value="?">...</option>
+      <option disabled key="user-edit" value="edit">
```
gabemart: I don't think the 'key' props are strictly necessary here and below, but they don't do any harm

gabemart:
Looks good to me

client/src/components/Users.jsx
peeb: I think I saw a console warning about flattened children (!) - but it's probably harmless.

Issue #692, opened 2017-04-12
Show full cell values on hover in dataset view


gabemart:
In the dataset view, we show columns with cells.

Currently, columns cannot be resized, so cells are fixed-width. Long values will overflow the cell boundaries and not be fully visible.

Given that in the short term we cannot resize columns, we should show full cell values on hover

jonase:
merged

PR #693, opened 2017-04-12
[#692] Show full cell value on hover

gabemart:
- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


jonase:
@gabemart I think we should calculate `formatCellValue` only once. This function is called once per cell which is a lot.

gabemart:
@jonase you're totally right, changed now

PR #731, opened 2017-05-15
Issue/672 csv import

peeb:
- [x] Update release notes
- [ ] Copyright header
- [x] Code formatting
- [ ] Documentation

RELEASE NOTES

# 0.9.1

## New and Noteworthy

* Numeric values are now auto-detected when importing CSV data. Note that if there is any data of mixed types, it will be imported as text.

Quick overview:

CSV data is imported as is - i.e. numeric values are imported as numeric strings.
This data is then inspected and a column type is determined if all values in the column are of the same type (or nil).
If this is the case, numeric strings are cast to double precision numbers for insertion into PSQL. If not, the numeric strings are passed through unchanged.

The goal is to accommodate mixed/bad data where there are values in a column that may not be able to be coerced to a number.


iperdomo:
Just as _reminder_ to a quick conversation with @peeb 

Our current approach for detecting data types uses `cuerdas.core/numeric?` that is just based on numeric values using a US based locale (dots `.` as decimal separator)

A CSV can be actually a `;` separated values file (e.g. A file exported from Excel using Spanish locale) where the `,` is used as decimal separator. In this case we'll fail to detect a column as a numeric one.

Issue #771, opened 2017-05-29
Demarcate file drop area


Kiarii:
# Context
Why do we add this issue?

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?
![image](https://cloud.githubusercontent.com/assets/12692444/26547480/f2de1592-4477-11e7-93fd-cfc84c603c8b.png)

loicsans:
Priority: 5
effort: 1

PR #856, opened 2017-06-15
[#771] Dropping area is clearly marked

loicsans:


gabemart:
Made one small change as agreed. LGTM otherwise

@loicsans will you test and then then merge if all is well?

PR #891, opened 2017-06-27
Issue/840 transformation transactions

jonase:
- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation


kardan:
As usual things work and look fine. We should however rename the akvo.lumen.transformation/schedule it's easy to trick one self that some async with jobs is going on.

Issue #816, opened 2017-06-08
Rethink organisation of non-component utility code


gabemart:
# Context
We have a some utility code in `utilities` and some in`utils.js`. Some components have utility code written as pure functions within the component file.

# Problem or idea
It can be a confusing to know where to put new utility code and where to find existing utility code.

# Solution or next step
Let's reorganize and put all reusable utility code in one place.

PR #913, opened 2017-06-29
[#816] Fresh branch for client utils reorg

peeb:
- [x] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

Are release notes necessary here?

### New & Noteworthy

* Client utility code has been reorganised where it makes sense

### DIFF
--- a/client/src/components/dashboard/DashboardEditor.jsx
+++ b/client/src/components/dashboard/DashboardEditor.jsx
@@ -1,16 +1,16 @@
+import _ from 'lodash';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ReactGridLayout from 'react-grid-layout';
 import DashboardVisualisationList from './DashboardVisualisationList';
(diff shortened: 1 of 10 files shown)


client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -3,13 +3,12 @@ import PropTypes from 'prop-types';
 import ReactGridLayout from 'react-grid-layout';
 import DashboardVisualisationList from './DashboardVisualisationList';
 import DashboardCanvasItem from './DashboardCanvasItem';
+import getArrayFromObject from '../../utilities/general';
```
jonase: I believe `getArrayFromObject` is the same as `Object.values` (not yet standard, needs to be polyfilled) or Lodash [_.values](https://lodash.com/docs/4.17.4#values).

As we're already using lodash, perhaps use `_.values` instead of our own util?

peeb: 👍

Issue #801, opened 2017-06-02
Browser caching issues due to keycloak update


jonase:
# Context
We need to update keycloak for offline tokens to be handled correctly. This will be used for updateable flow datasets.

# Problem or idea
In some cases lumen will get into a refresh loop if the browser cache is not cleared.

We believe that the reason is that keycloak caches `login-status-iframe.html` for 30 days:

<img width="698" alt="screen shot 2017-06-02 at 9 45 23" src="https://cloud.githubusercontent.com/assets/101023/26714053/42c446e6-4778-11e7-8571-04f88747da05.png">

and with Keycloak 3.1.0 the iframe has changed:

```
$ diff old-iframe.html new-iframe.html
56c56,57
<                 if (sessionState === cookie) {
---
>                 var c = cookie.split('/');
>                 if (sessionState === c[2]) {
84c85
<             event.source.postMessage('error', origin);
---
>             return;
```

This means that if the browser has cached this file, it will get the 2.5.1 version even if keycloak is running 3.1.0 (for up to 30 days).

# Solution or next step

From the keycloak javascript adapter docs:

> By default, the JavaScript adapter creates a hidden iframe that is used to detect if a Single-Sign Out has occurred. [...] This feature can be disabled by setting checkLoginIframe: false in the options passed to the init method.

So what we can do is set `checkLoginIframe` to `false` at least for 30 days to let browsers expire their caches without the user needing to manually do it.

Since lumen is the only keycloak user right now we don't strictly need the single sign out feature yet.

We can add back the iframe later.

iperdomo:
@jonase @kardan thanks for the debugging/analysis.

It's true that this iframe gets cached for 30 days

    Cache-Control: max-age=<seconds>

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

![cache-control](https://cloud.githubusercontent.com/assets/178474/26717786/f5193e54-477e-11e7-9daa-feabc8280433.png)

IMHO this should not be cached at all, will check which layer is adding that `cache-control` header

kardan:
Ref:
https://issues.jboss.org/browse/KEYCLOAK-5022
https://issues.jboss.org/browse/KEYCLOAK-5065

jonase:
I think we can close this now. We'll reopen if we find other related issues

PR #954, opened 2017-07-13
[#801] Redirect to login page in case of token refresh failure

jonase:
- [ ] Update release notes
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation

### DIFF
--- a/client/src/auth.js
+++ b/client/src/auth.js
@@ -9,10 +9,13 @@ export function token() {
     throw new Error('Keycloak not initialized');
   }
 
-  return new Promise((resolve, reject) =>
+  return new Promise(resolve =>
     keycloak.updateToken()
       .success(() => resolve(keycloak.token))
-      .error(err => reject(err))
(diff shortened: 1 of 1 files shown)


client/src/auth.js
```diff
@@ -9,10 +9,13 @@ export function token() {
...
   }
 
-  return new Promise((resolve, reject) =>
+  return new Promise(resolve =>
     keycloak.updateToken()
       .success(() => resolve(keycloak.token))
-      .error(err => reject(err))
+      .error(() => {
+        // Refreshing will redirect to login page
+        window.location.reload(true);
```
iperdomo: Why not use `keycloak.login()` ?

jonase: Good point. Changed to `keycloak.login` instead.

iperdomo:
Will try it locally before merging

iperdomo:
👍

PR #995, opened 2017-08-09
Issue/992 adr 007

LarsSjogreen:
Does this look ok to you @dlebrero? Ok with you @bjelkeman and @kardan?

I have two questions:
* How are SQL queries represented in the monitoring system? Will there be a risk of information/data leak?
* Should we add a checkbox in the code review checklist to safeguard against unintended data leak (in monitoring and logging)?

### DIFF
--- a/doc/arch/adr-007-stackdriver.md
+++ b/doc/arch/adr-007-stackdriver.md
@@ -57,5 +57,6 @@ Preliminary decision by head of software development after discussion with lead 
 * Application monitoring for Windshaft can be set up
 * Operations and support handled by third party
 * An added SaaS cost
+* Added software development cost for identity hashing functionality. Estimated to be around 3 days of work.
 * Added process cost for reviewing monitoring data before any code changes to systems utilizing monitoring functionality
 * A step in the right direction on the long term strategy of unification of supporting software services in Akvo
--- a/doc/arch/adr-007-stackdriver.md
+++ b/doc/arch/adr-007-stackdriver.md
@@ -1,0 +1,61 @@
+# ADR 007: Using Google Stackdriver as the monitoring solution for Windshaft
+
+## Context
+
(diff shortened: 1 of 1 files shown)


doc/arch/adr-007-stackdriver.md
```diff
@@ -0,0 +1,62 @@
...
+
+It is also important that care is taken to ensure that data that is sent from Windshaft to Stackdriver does not identify the partner or leak any of the data on their subjects.
+
+To allow for efficient usage and to get value from the monitoring system, an application owner needs to be able to identify different cases of use to see which cases have a severe impact on application performance. A suggested strategy for achieving this is to use a strong hashing algorithm on the identifiers in the monitoring data (for example the tenant and user IDs) and to generate and store the hash keys, the salt(s) and the identity lookup table privately and securely for our convenience. Even when we obfuscate the identities in this way, we should still carefully consider what data and metrics will be sent to the monitoring systems so that these could not be used to glean important information on our users and their source data.
+
+Question: how will SQL queries be handled in the monitoring system? [Dan]
```
dlebrero: The SQL queries are created and executed deep in the mapnik C code, so we are not going to be able to explicitly monitor them, unless we decide to change the C code, which I suggest we should not :).

From the JS code, we can measure how long the C code takes to run, but the code includes the DB call + some other heavy work. 

For example, some local testing shows that JS reports 150ms but Postgres reports 20ms. 

Even if we cannot know from the JS code how long the SQL query took, it will still give us a general idea of where the time is spent.

LarsSjogreen: My concern was around what kind of information will be visible in the monitoring system with regards to the SQL queries. If these queries are not visible (with query parameters) then I'm ok with it.

I agree with you that the exact time elapsed is of less importance as long as we have some numbers that are consistently created and comparable to each other.

PR #1003, opened 2017-08-14
Issue/967 geo transformation

peeb:

RELEASE NOTES

Made our first steps down the path towards geospatial data support in Lumen. There is now support for transforming latitude and longitude data into PostGIS geometry point data.


backend/project.clj
```diff
@@ -16,6 +16,8 @@
                  [duct "0.8.2"]
                  [duct/hikaricp-component "0.1.0"]
                  [environ "1.1.0"]
+                 [garden "1.3.2"]
+                 [geo-clj "0.6.2"]
```
jonase: Are these two deps used? Could not find them in code.

backend/src/akvo/lumen/transformation/geo.clj
```diff
@@ -0,0 +1,43 @@
...
+
+(defmethod engine/valid? :core/generate-geopoints [op-spec]
+  (valid? op-spec))
+
+(defmethod engine/apply-operation :core/generate-geopoints
+  [tenant-conn table-name columns op-spec]
+  (try
+    (let [{:strs [columnNameLat columnNameLong]} (engine/args op-spec)
+          column-name-geo (engine/next-column-name columns)
+          column-title-geo (format "Geopoints (%s)" column-name-geo)
```
jonase: The new column title is not in the `op-spec` (from the browser)?

backend/src/akvo/lumen/transformation/geo.clj
```diff
@@ -0,0 +1,43 @@
...
+          opts {:table-name table-name :column-name-geo column-name-geo}]
+      (jdbc/with-db-transaction [conn tenant-conn]
+        (add-geometry-column conn opts)
+        (generate-geopoints conn (conj opts {:column-name-lat columnNameLat
+                                             :column-name-long columnNameLong})))
+      ;; TODO - remove original lat/long columns
```
jonase: Do we want to do that? The user could always remove them manually if they want to. Could also possibly be a boolean in the `op-spec`

backend/src/akvo/lumen/transformation/geo.clj
peeb: Yes.. should be. I'll fix this.

backend/project.clj
peeb: Ah yes - two more stragglers from the client work. This is because 967 started as 901.

backend/src/akvo/lumen/transformation/geo.clj
peeb: The boolean option might be a good compromise, I agree.

doc/spec/transformations.md
```diff
@@ -223,4 +223,22 @@ Examples:
...
+```
+
+### core/generate-geopoints
+
+* args
+  * columnNameLat (string): The column containing latitude data
+  * columnNameLong (string): The column containing longitude data
+  * columnTitleGeo (string): The title of the geopoint column to be generated
+  * columnTypeLat (string): The client data type of the latitude column
+  * columnTypeLong (string): The client data type of the longitude column
```
jonase: Why are these needed? When the backend knows the `columnNameLat/Long` it can figure out the types as well.

peeb: Good point.. I'll refactor.

PR #1013, opened 2017-08-18
Issue/1010 unlimited

kardan:

### DIFF
--- a/backend/src/akvo/lumen/lib/tier_impl.clj
+++ b/backend/src/akvo/lumen/lib/tier_impl.clj
@@ -3,7 +3,9 @@
    [ring.util.response :refer [response]]))
 
 (defn all [tenant-conn]
-  (response {"tiers" {"standard" {"numberOfExternalDatasets" 5
+  (response {"tiers" {"unlimited" {"numberOfExternalDatasets" "*"
+                                   "numberOfVisualisations" "*"}
+                      "standard" {"numberOfExternalDatasets" 5
                                   "numberOfVisualisations" 50}
                       "pro" {"numberOfExternalDatasets" 20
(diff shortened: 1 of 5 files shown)


backend/src/akvo/lumen/lib/tier_impl.clj
```diff
@@ -3,7 +3,9 @@
    [ring.util.response :refer [response]]))
 
 (defn all [tenant-conn]
-  (response {"tiers" {"standard" {"numberOfExternalDatasets" 5
+  (response {"tiers" {"unlimited" {"numberOfExternalDatasets" "*"
```
jonase: Should we use `nil` instead of `"*"` for "no limit"?

kardan: I was not sure what would be best but thought  * made more sense from the UI point of view. Rendering "nil" (not sure that's what you suggest) makes me think something is broken.

kardan: Maybe we should say "unlimited"

kardan: @nadiagorchakova thoughts?

jonase: What we render is different from what we the data structure contains. I was just thinking that it's a bit strange to have key/value pairs where the the value is sometimes a number and sometimes a string. `null` to me would mean "no value" and would therefor be unlimited.

We can still render *, unlimited, ∞ or whatever

kardan: Agree!

jonase:
`RELEASE NOTES`: Added "unlimited" tenant plan

Issue #1042, opened 2017-08-30
Filter menu save button looks disabled, but is still clickable


jonase:
# Context
When configuring a visualisation filter we "disable" the save button until the filter is fully configured.

# Problem or idea
Even if the button looks disabled, it's still possible to click on it and trigger the "apply filter" functionality. This will either result in a faulty filter or an error being thrown

# Solution or next step
Actually disable the button

PR #1043, opened 2017-08-30
[#1042] Properly disable filter menu save button if filter is not ful…

jonase:
…ly configured

`RELEASE NOTES (bug)` Properly disable filter save button if filter is not fully configured


gabemart:
Fixes the console error 👍

PR #1044, opened 2017-08-31
Issue/972 client geo transform

peeb:

RELEASE NOTES

### New and noteworthy

* It is now possible to generate PostGIS geometry data from user selectable latitude and longitude column data

## IMPORTANT: Do **not** merge this PR as is! The feature must be disabled before merging into develop, since there is currently no way to use the new geometry data for map visualisations yet. This could be very confusing for end users.


client/src/components/dataset/sidebars/GenerateGeopoints.jsx
```diff
@@ -0,0 +1,136 @@
...
+            value={args.get('columnNameLong')}
+          />
+          <div className="inputGroup">
+            <label
+              htmlFor="titleTextInput"
+            >
+              <FormattedMessage id="column_title_geo" />
+            </label>
+            <input
+              value={args.get('columnTitleGeo')}
```
gabemart: This generates a console error on first input - `Warning: GenerateGeopoints is changing an uncontrolled input of type text to be controlled.`

I guess that's because the value is initially null. Perhaps we should change to something like `value={args.get('columnTitleGeo') || ''}`

client/src/components/dataset/sidebars/GenerateGeopoints.jsx
```diff
@@ -0,0 +1,136 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import Immutable from 'immutable';
+import _ from 'lodash';
```
gabemart: I thought that using lodash here might bloat the bundle size, and I was going to suggest that perhaps it's overkill give that we're only using it to capitalize some strings - but in fact lodash is being loaded anyway in the `app` bundle, so using it here has zero effect on the size of the `Dataset` bundle. So this comment can be safely ignored :)

client/src/components/dataset/sidebars/GenerateGeopoints.jsx
peeb: :+1:

client/src/components/dataset/sidebars/GenerateGeopoints.jsx
peeb: Yeah, I figured it would have no impact here ^^

gabemart:
LGTM 👍

PR #1057, opened 2017-09-11
Stats2

kardan:


kardan:
Not sure what we think is "finished state" on this script. Hence opened a PR so it becomes more visible.

jonase:
The requests for stats are almost never the same (at least up til now). So the initial idea with the `run-query` and `run-manager-query` scripts were to be able to run any SQL query easily, get the result in json, and do whatever post-processing necessary separately.

The `report` namespace is now for one specific request and the next "report" will probably be very different.

kardan:
RELEASE NOTES: (New and noteworthy)
- Scripts for stats

PR #1074, opened 2017-09-19
Issues/1069 ssl postgres in dev

dlebrero:
Changed Postgres to make SSL mandatory
Importing the development Postgres certificate into the Backend
Change all url to be ssl=true in the backend
Windshaft version that supports SSL
Configuring SSL for Windshaft
Configuring SSL for CI
Tests to use also SSL connections


jonase:
@dlebrero I'm seeing the following in the log after `docker-compose up`

```
backend_1    | Sep 26, 2017 4:18:33 PM org.postgresql.Driver connect
backend_1    | SEVERE: Connection error:
backend_1    | org.postgresql.util.PSQLException: The server does not support SSL.
```

jonase:
Tested and works as expected.

`docker-compose down && docker-compose up --build` required

PR #1092, opened 2017-09-28
[#1091] Add healthz check

jonase:

`RELEASE_NOTES` Add /healthz check


client/server.js
```diff
@@ -13,6 +13,10 @@ new WebpackDevServer(webpack(config), {
   publicPath: config.output.publicPath,
   disableHostCheck: true,
   proxy: {
+    '/healthz': {
```
kardan: Guess we could make these alphabetic, but not sure it's relevant to this PR.

Issue #1033, opened 2017-08-28
Improvements to UI translations


nadiagorchakova:
# Context
When we released Spanish and French UI translations, people were super happy about it but also noticed some gaps/things to improve.
I'll document those suggestions in this issue.

# Spanish
Ivan:
* Library right now is translated to "Librería". In Spanish "Librería" is where you buy books. In IT terms the correct translation is "Biblioteca" [1], but in this case I would suggest  "Catálogo" [2]

* Dataset is translated to "Conjunto de datos" - is a long translation, and I was suggesting "Set de datos" that is a bit shorter. I don't have strong opinion on this one. Both terms have been used as valid translations to the term Dataset. World Bank's open data portal translates the term "Dataset" to "Bases de datos" which is also correct.

# French
How could we solve it?

# Missing translations

* Dataset filter in visualisaiton editor
<img width="374" alt="screen shot 2017-08-30 at 16 26 33" src="https://user-images.githubusercontent.com/21127166/29877515-1f88d994-8da0-11e7-98c1-5f7bcdedc781.png">

* When renaming a column, the 'Rename' button is not translated:
<img width="265" alt="screen shot 2017-08-30 at 16 21 55" src="https://user-images.githubusercontent.com/21127166/29877249-615db32c-8d9f-11e7-87a0-991f7931e25f.png">

* Visualisation type as shown in the library, e.g. map or pivot table:
<img width="297" alt="screen shot 2017-08-29 at 13 55 38" src="https://user-images.githubusercontent.com/21127166/29819670-ef2d538a-8cc1-11e7-886b-dcf895b9d590.png">

* Admin section
* Actions in Transformation log, e.g. 'deleted a column'
<img width="273" alt="screen shot 2017-08-30 at 16 22 54" src="https://user-images.githubusercontent.com/21127166/29877338-9b681616-8d9f-11e7-8588-fa7b32ca65ae.png">

* Map visualisation editor:

<img width="1181" alt="screen shot 2017-08-29 at 14 00 43" src="https://user-images.githubusercontent.com/21127166/29819802-8370d454-8cc2-11e7-8f91-12e807c989a0.png">

* Data type for columns in visualisations editor:

<img width="382" alt="screen shot 2017-08-29 at 14 01 36" src="https://user-images.githubusercontent.com/21127166/29819854-ac27d942-8cc2-11e7-91d2-49ce84fa2786.png">

* 'Untitled dashboard' and 'Untitled visualisation'

[...]

Issue #1077, opened 2017-09-20
Use Windshaft for rendering maps


kardan:
# Context
We have decided and done ground to switch to Windshaft for maps. At the moment we have Windshaft running in the Lumen pod next to lumen. This enables us to use it more or less as a library. 

# Problem or idea
We need to reach feature parity with the current map solution but using Windshaft as a tiler.

# Solution or next step
We have a static test for rendering maps using Windshaft, we should take this a step further and do a test with client, backend and Windshaft. 

ref:
- [#1004](https://github.com/akvo/akvo-lumen/issues/1004)
- [#1061](https://github.com/akvo/akvo-lumen/issues/1061)
- [#963](https://github.com/akvo/akvo-lumen/issues/963)
- [#901](https://github.com/akvo/akvo-lumen/issues/901)

plan:
We'll add the new map under the current one in the visualisation editor. This way we can build up a pair of maps with one visualisation editor and verify that they behave in a similar fashion. 
Things to do (WIP):
- [x] Communicate to windshaft over correct host (localhost in production)
- [x] Use dynamic url for POST to Windshaft
- [x] Use correct headers for POST to Windshaft
- [x] Build simple SQL case
- [x] Investigate how to render legends if not all data is available in the client. Will probably need to populate the pointColorMapping before returning from /api/visualisations/maps. 
- [x] Feature parity for legend positioning and overflow
- [x] Pop up support
- [x] Image popup support
- [x] Bugfix - points filtered out via dataset editor still have popup from utfgrid
- [x] ~~Zoom map? (map bounding box currently works, but potentially we should save the zoom and pan from the editor view /g)~~ not going to do this
- [x] Rate-limit client requests to backend (in progress)
- [x] Be smarter about when to rerender map (e.g. we don't need to refresh legend when poup changes)
- [x] Refactor / clean up backend code (especially around public)
- [x] Validate map specs on the backend. (D)
- [x] Bugfix - filtered-out values should not appear in the pointColorMapping
- [x] Make public maps work (new maps needs windshaft ref). This will require the public endpoint to depend on config.
- [x] Fix issue where public maps client tries to use the "rows" from a dataset. 
- [x] Remove the double maps and only show one
- [x] New version of akvo-maps does not use tenantDB in the url, we should sync up with the platform team and follow the release of akvo-maps
- [x] Only do backend requests if we have a "sensible" visualisation spec
- [x] Show lat / long dropdonwn for old maps and geom column dropdown for v2 and new map?
[...]

PR #1093, opened 2017-10-02
[#1033] Improving French translations

loicsans:
Seems like we are missing the map section translations.


client/src/translations/fr.json
```diff
@@ -76,8 +74,6 @@
   "filter_text": "Filtrer texte",
   "filter": "Filtre",
   "flow_application_url": "Flow Application URL",
-  "generate": "Générer",
-  "generate_geopoints": "Générer des données géométriques",
```
peeb: Why were these removed? Bad translations?

loicsans: Possibly a mistake

kardan:
@loicsans should we just add these two keys back and merge?

peeb:
The maps translations *are* in #1077 so I'm not sure what's happened.

I suggest we merge this PR since the geo translations will make it into develop from 1077. Thoughts? @loicsans @kardan

Issue #1098, opened 2017-10-08
Improve development stack based on Docker and docker-compose


iperdomo:
# Context

A great improvement has been done with an _easy_ command `docker-compose up`, however there are some differences when running this containerized development setup between macOS and Linux.

One of the main issues is the permissions assigned to new files created within the container on an exposed host folder using `--volumes`.

# Problem or idea

On Linux dev host:
````
# Create a new file inside a container inside an exposed folder from the host
# Try to remove/touch/edit the file as a host user

[ivan@utila tmp]$ docker run --rm --volume "$(pwd)":/code clojure:lein-2.7.1 touch /code/new-file.txt
[ivan@utila tmp]$ ls -l new-file.txt 
-rw-r--r-- 1 root root 0 Oct  8 14:17 new-file.txt
[ivan@utila tmp]$ rm -rf new-file.txt 
rm: cannot remove 'new-file.txt': Operation not permitted
````
The new file is created as is `root` was the owner and group leading to problems on the host machine (we don't need to use `sudo` to deal with files)

The same happens with usual `target` folder or any other file created inside the container, e.g. the `backend` folder:

````
[ivan@utila backend]$ ls -laht
total 72K
-rw-r--r--  1 root root    5 Oct  8 14:13 .nrepl-port
-rw-r--r--  1 root root   15 Oct  8 14:13 .lein-env
drwxr-xr-x  8 ivan ivan 4.0K Oct  8 13:54 .
drwxr-xr-x  3 root root 4.0K Oct  8 13:53 target
[...]
````

This is a known issue, the workaround is to setup a user inside the container with the same user id (`id -u`) and group id (`id -g`) as the host user.

*NOTE:* Docker has the ability to run as _any_ user via `docker run -u`, but this doesn't work because of leiningen and how maven cache works.

# Solution or next step
* Create a user in the container [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint) and execute the `lein <command>` as the newly created user
* Inspiration:
  * https://github.com/markmandel/wrapping-clojure-tooling-in-containers
  * https://github.com/iperdomo/tmate.clj/blob/master/startup.sh

kardan:
What's the status on this?

iperdomo:
@kardan done

PR #1099, opened 2017-10-09
[#1098] Run lein commands as with host's user and group id

iperdomo:
* Setup an .env file with the host's user and group id
* Use lein image based on Alpine linux: 96MB (lein-2.7.1-alpine) vs 313MB (lein-2.7.1)
* The `run-as-user` makes sure the `akvo` user and group exists
  with the hosts UID and GID
* docker-compose file adjustments to use .env settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ npm-debug.log
 client/dist
 client/yarn.lock
 
+# docker-compose
+.env
+
 # Backend
 backend/target
 backend/classes
--- a/README.dev.md
+++ b/README.dev.md
@@ -13,7 +13,7 @@ This Docker Compose environment will have:
 - A Redis DB
 
 ## Hosts
-Akvo Lumen is a multi tenant system and to do enable local routing to the tenants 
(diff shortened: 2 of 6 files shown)


dlebrero:
The client needs something similar, right? 

Have your tried https://docs.docker.com/engine/reference/run/#user? I see that https://denibertovic.com/posts/handling-permissions-with-docker-volumes/ explains that maybe some apps will not work for maybe our does.

iperdomo:
@dlebrero check the issue description

> NOTE: Docker has the ability to run as any user via docker run -u, but this doesn't work because of leiningen and how maven cache works.

Basically, by default `lein` expects that `$HOME/.lein` exists and `$HOME/.m2` ... you can override the location of lein with an environment variable [LEIN_HOME](https://github.com/technomancy/leiningen/blob/master/bin/lein#L102) and `$HOME/.m2/settings.xml`. See: https://github.com/technomancy/leiningen/blob/2.8.0/doc/FAQ.md

The main issue here is that `lein` expects `$HOME` and using  `docker run -u` may set a `UID` and `GUID` for a user that doesn't exist in the container.

PR #1110, opened 2017-10-23
[#1098] Apply changes to client (js) code

iperdomo:
- Use an alpine based image
- Install requirements for node/canvas
- Upgrade node-sass to a version that supports alpine linux
- Run `npm *` commands as `akvo` user


client/package.json
```diff
@@ -39,7 +39,7 @@
     "leaflet": "^1.0.3",
     "lodash": "^4.17.2",
     "moment": "^2.17.1",
-    "node-sass": "^3.13.0",
+    "node-sass": "^4.1.0",
```
gabemart: Does cause any problems in my testing 👍

PR #1153, opened 2017-11-08
Issue/1152 shape import

jonase:

`RELEASE NOTES`: Detect Well-Known Text (polygons and multipolygongs) as part of csv imports


backend/src/akvo/lumen/import.clj
```diff
@@ -6,7 +6,9 @@
              [akvo.lumen.util :as util]
              [cheshire.core :as json]
              [clojure.java.jdbc :as jdbc]
-             [hugsql.core :as hugsql]))
+             [hugsql.core :as hugsql])
+  (:import [org.postgresql PGConnection]
+           [org.postgresql.util PGobject]))
```
iperdomo: Shouldn't we import `org.postgis.Polygon` and `org.postgis.MultiPolygon` instead of `PGConnection` ?

backend/src/akvo/lumen/import/common.clj
```diff
@@ -45,7 +46,9 @@
 
        :text - java.lang.String
        :number - java.lang.Number
-       :date - java.time.Instant"))
+       :date - java.time.Instant
+       :geoshape - java.lang.String
+                   Well-known text (WKT) shape (POLYGON or MULTIPOLYGON)"))
```
iperdomo: Do we also support `POINT` ?

jonase: when we add support for `:geopoint`

backend/src/akvo/lumen/import.clj
jonase: 👍 fixed

PR #1158, opened 2017-11-13
Issue/1143 flow import

kardan:


peeb:
Looks good to me. Tested with a couple flow datasets from pilots.akvoflow.org.

PR #1159, opened 2017-11-13
Issue/1102 trim column titles

peeb:

# Bugfixes

- Column titles are now trimmed on import to avoid potential confusion later down the line


kardan:
Solution looks fine, but I wonder if Flow imports would benefit from the same trim treatment. Would it be possible to add it to Flow imports (if it's not already) or alternatively do the trim on a higher level so it's enforced for all and future import types?

peeb:
:+1: @kardan - good point

PR #1164, opened 2017-11-15
Issue/1096 puppeteer

baztan106483:
- Use Puppeteer library to simulate a user of lumen locally
- We do a test of that simulation when we push, using Travis CI


ci/build.sh
```diff
@@ -0,0 +1,31 @@
...
+    let TRIES=${TRIES}+1
+done
+
+docker-compose exec client su-exec akvo npm run lint
+docker-compose exec client su-exec akvo npm run test
+docker-compose exec backend su-exec akvo lein do clean, check, test :all
+
+docker run --interactive --tty --shm-size 1G --rm --network=akvolumen_default \
+       --volume "$PWD/client/ui-test/script-test.js":/app/index.js --link akvolumen_client_1:t1.lumen.local \
+       --link akvolumen_keycloak_1:auth.lumen.local --link akvolumen_client_1:t1.lumen.local \
```
dlebrero: links are duplicated?

client/src/components/Library.jsx
```diff
@@ -223,7 +223,10 @@ class Library extends Component {
 
 
     return (
-      <div className="Library">
+      <div
+        className="Library"
+        data-test-id="library"
```
dlebrero: why data-test-id and not just id?

client/ui-test/script-test.js
```diff
@@ -0,0 +1,129 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
```
dlebrero: Is this Copyright right?

client/ui-test/script-test.js
```diff
@@ -0,0 +1,129 @@
...
+		await page.click('li[class^="listItem"]');
+		await console.log("Typing dashboard name...");
+		await page.waitForSelector('div[data-test-id="dashboard-canvas-item"]', {timeout:10000});
+		await page.click('div[data-test-id="entity-title"]');
+		await page.type('input[data-test-id="entity-title"]', 'Dashboard of '+name);
+		await console.log("Saving dashboard...");
+		await page.click('button[data-test-id="save-changes"]');
+		await page.click('i[data-test-id="fa-arrow"]');
+		await console.log("Dashboard "+name+" was successfully created.\n");
+		await console.log("THE TEST WAS SUCCESSFUL");
```
dlebrero: I don't see any asserts in the test. I can see failing if the "waitForSelector" timeouts. Any other causes for it to fail?

docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
       - ~/.lein:/home/akvo/.lein
       - ./postgres/provision:/pg-certs
    links:
-      - keycloak:auth.lumen.localhost
```
dlebrero: Do this changes in the hostname require a change in the documentation? I can think of README.dev.md at least.

client/src/components/Library.jsx
baztan106483: Hi Dan, the reasons are:
1 - In HTML 5 there is a particular kind of attributes called data attributes:
          https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
2 - By agreement, we use these attribute names to call the attributes we use for tests.
3 - Those attributes are useful only when we do the test, so it would be interesting to delete or hide them when we build, and there is an option that allow us to do it.

ci/build.sh
baztan106483: Yes, that is a mistake, good observation

client/ui-test/script-test.js
baztan106483: Yes it is

client/e2e-test/script-test.js
```diff
@@ -78,25 +79,36 @@ const datasetName = Date.now().toString();
       return Promise.resolve(found.getAttribute('data-test-id'));
     });
     console.log(`ID extracted: ${id}\n`);
-
+    let pending;
+    do {
+      pending = await page.$(`[data-test-name="${datasetName}"] [data-test-id="pending"]`);
+    } while (pending);
```
dlebrero: infinite loop?

baztan106483: The data-test-id "pending" disappears when the dataset is completely created and available. Then, the variable pending will be undefined, and the JavaScript API says that this will return false: 
https://developer.mozilla.org/es/docs/Glossary/Truthy
So the loop will end in that moment.

dlebrero: So an infinite loop if the "pending" does not disappear because something is broken.

baztan106483: Now we added a timeout for the loop.

client/e2e-test/script-test.js
```diff
@@ -0,0 +1,144 @@
...
+    const id = await page.evaluate(() => {
+      const found = document.querySelector(`[data-test-name="${__datasetName}"]`);
+      return Promise.resolve(found.getAttribute('data-test-id'));
+    });
+    console.log(`ID extracted: ${id}\n`);
+    let pending;
+    const timeOut = setTimeout(() => { console.log('Error waiting for pending dataset'); process.exit(1); }, 60 * 1000);
+    do {
+      pending = await page.$(`[data-test-name="${datasetName}"] [data-test-id="pending"]`);
+    } while (pending);
```
dlebrero: Does this loop consume 100% cpu? Maybe add a small sleep?

README.dev.md
```diff
@@ -119,7 +119,7 @@ psql --host=akvolumen_postgres_1 --port=5432 --dbname=lumen_tenant_1 --username=
 This container has a development version of the Windshaft container, with plenty of hardcoded assumptions.
 
 The Windshaft server is not exposed directly to the external world, but it is proxied by the Webpack server
-on the url http://t1.lumen.localhost:3030/maps/**. That url forwards the requests to "windshaft:4000".
+on the url http://t1.lumen.local:3030/maps/**. That url forwards the requests to "windshaft:4000".
```
dlebrero: Doc how to run the tests locally

baztan106483: Done!

client/e2e-test/script-test.js
baztan106483: Now we added a sleep

PR #1174, opened 2017-11-28
Issue/1162 reverse geocoding

jonase:
`RELEASE NOTES`: Reverse geocoding transformation


nadiagorchakova:
@kardan I'm also concerned about the size of the world dataset. Perhaps we should only load it to trial.akvolumen.org and demo.akvolumen.org. We can load it to partner or hub instances if they specifically ask for it.

I think admin level 2 is good to start with. I'm not sure if that's enough for the majority of partners, we'll figure it out as partners start using it. In the Liberian case, for instance, people want to see data per county, which I think is level 2.

PR #1190, opened 2017-12-08
Issue/1162 reverse geocoding 2

kardan:


client/src/components/transformation/ReverseGeocodeTransformation.jsx
```diff
@@ -0,0 +1,125 @@
...
+          mergeColumn: null,
+        },
+      }),
+    };
+  }
+
+  componentWillMount() {
+    const { dispatch, datasetId } = this.props;
+    const { spec } = this.state;
+    dispatch(ensureDatasetFullyLoaded(datasetId))
```
gabemart: It's a shame that we are fetching the entire dataset (often several MB) just to get a list of column metadata for a dropdown... but we do exactly the same thing in many other places. 

I also think that if we were being really strict about the redux design pattern, we should move the `dispatch` call to the container and pass a plain callback to this component.

But I think we should address both these things separately - I really think it's time that we seriously looked into loading all column metadata by default, which would resolve both the above problems.

kardan: Agree on this.

gabemart:
LGTM!

Issue #1203, opened 2017-12-21
Use a release version of org.akvo/resumed


dlebrero:
Lumen is depending on a snapshot version of org.akvo/resumed, which means that we don't really know what we are deploying.

PR #1204, opened 2017-12-21
[#1203] using non-snapshot version of org.akvo/resumed

dlebrero:


kardan:
Closing after discussion with @dlebrero.

Issue #1226, opened 2018-01-16
Polyfills for IE10 and older browsers


gabemart:
# Context
We should support IE10 and older browsers where possible.

# Problem or idea
We take advantage of some newer browsers APIs like `fetch`, `Object.assign` and `Intl` which are not supported in older browsers

# Solution or next step
We should conditionally load polyfills for older browsers where we can

PR #1227, opened 2018-01-16
[#1226] Add polyfills to support older browsers and IE

gabemart:


gabemart:
Should work in IE10 now. I'm am using https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ to test IE10

PR #1242, opened 2018-01-29
Issues 1239/move to travis

dlebrero:
Moving build to Travis.

The build steps are now:

1. Run the Backend unit tests
2. Build the Backend production image
3. Run the Client unit tests + linting 
4. Build the Client production image
5. Build the Windshaft production image
6. Start a Docker Compose environment for additional testing
7. Run the Backend functional tests (using the previous env)
8. Run the end to end tests (using the previous env)
9. If master or develop branch, publish containers and deploy

The most significant change is that now the end to end tests run against the production images. This means:
1.  We test the production images before using them in production
2. We avoid the intermittent issue with the default Postgres DB that comes with Travis/CircleCI
3. The build takes 10 minutes instead of 6
4. The whole build can be tested and run locally
5. Some minor changes in the prod code, for example to not require Sentry DSN

We have also changed the way we build the production images. We now use the developer image to create the assets that will be use in the production images. This means:
1. Building the production images just requires docker, no need to install lein/jvm
2. Removed the duplication between the client dev and prod images
3. The Client image is significantly smaller: from 600mb to 50mb


backend/import-and-run.sh
```diff
@@ -2,4 +2,13 @@
...
 keytool -import -trustcacerts -keystore /usr/lib/jvm/java-1.8-openjdk/jre/lib/security/cacerts -storepass changeit -noprompt -alias postgrescert -file /pg-certs/server.crt
 
-./run-as-user.sh lein repl :headless
+if [ -z "$1" ]; then
+    ./run-as-user.sh lein repl :headless
+elif [ "$1" == "functional-and-seed" ]; then
+    # Two thing in one so that we avoid starting yet another JVM
+    ./run-as-user.sh lein do test :functional, run -m dev/migrate-and-seed
+elif [ "$1" == "kubernetes-test" ]; then
+    ./run-as-user.sh lein test :kubernetes-test
```
iperdomo: Do we have tests annotated with `:kubernetes-test`?

client/ci-build.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
iperdomo: if you use `#!/usr/bin/env bash` like the other script you can use the keyword `function`. I guess we should standardize to use bash, right?

backend/import-and-run.sh
dlebrero: No, I was planning to add one but now I realize that those will be with puppeteer.

client/ci-build.sh
dlebrero: Bash was/is not available in some of the prod containers

dlebrero: Ignore the previous comment

Issue #1239, opened 2018-01-25
Move to Travis CI


dlebrero:
We started the work to build Lumen in Travis, as Circle CI 1.0 do not support Docker. 

Instead of upgrading to CircleCI 2.0, we decided to use Travis, as Lumen is the only project using CircleCI.

Right now, Travis is just running the end to end tests, but do no publish artifacts or do deploys.

This issue is to finish that move to Travis and remove the CircleCI build

PR #1254, opened 2018-02-05
[#1239] Remove deprecated files

iperdomo:
- CircleCI is not longer needed
- Makefile setup is superseded by docker-compose setup


dlebrero:
There are also build scripts on the postgres and key cloak directories plus the ci/mini, ci/dev and ci/prod are no longer used.

I would also delete the ci/lumen-debug-pod.yaml as I don't think it is used or kept up to date.

PR #1259, opened 2018-02-06
Issues 1253/blue green

dlebrero:
Support for blue/green:

1. Lumen will have 2 deployments (blue/green)
1. Lumen will have 2 services (dark/live)
1. Added an nginx deployment in front of the services that routes the traffic to the correct service
1. Added a service called lumen-blue-green-gateway for the nginx
2. CI will push to dark deployment (which can be blue or green)
3. Blue/green state is queried from the service state
4. Added script to flip between blue/green

Dark Lumen will be available on the urls "dark-${tenant}."

All the above Kubernetes objects are new and after the merge to develop, they will just be deployed but not used. 

Once we are happy with this new deployment, the existing ingress must be changed to point to the lumen-blue-green-gateway and the old deployment and service must be manually removed.

Note that from the point of merging this branch, deployments will not happen in the old lumen deployment anymore, just in the new blue green


iperdomo:
LGTM, i made question on the possible need of adding `set -eu` to all bash code, but is up to @dlebrero  to make the change as he feels to :)

:+1:

Issue #1271, opened 2018-02-08
JS is being minimised in dev environment


gabemart:
The new `UglifyJsPlugin()` is being invoked in the dev environment where it has no advantages and significantly slows down bundle rebuilds.

We should only invoke it in the production environment.

PR #1272, opened 2018-02-08
[#1271] Only include uglify plugin in production env

gabemart:
#1271 


kardan:
If I understand it correctly (I know I also asked) this PR is deprecated by https://github.com/akvo/akvo-lumen/pull/1280

PR #1280, opened 2018-02-15
Optimizing webpack setup

finnfiddle:
#1744 - [ ] **Update release notes if necessary**


client/ci-build.sh
```diff
@@ -7,13 +7,12 @@ log () {
 }
 
 npm set progress=false
+
 log Installing npm
 npm install
+
 log Rebuild node-sass
 npm rebuild node-sass
-log Linting
```
finnfiddle: these are not needed because they get run by `npm run build`

client/tools/devServer.js
```diff
@@ -0,0 +1,106 @@
...
+  server: {
+    baseDir: 'src',
+
+    middleware: [
+
+      history({
+        rewrites: [
+          { from: '^/favicon.ico$', to: 'favicon.ico' },
+          { from: '^/s/.*$', to: '/index-pub.html' },
+          // { from: '^.*$', to: '/index.html' },
```
finnfiddle: this rewrite was problematic because it was making requests for js files getting returned the index.html file - we just need to check the same thing is happening on the index-pub.html page - Im unfamiliar with this @gabemart

gabemart: The index-pub page is served for shared visualisations and dashboards, and you're suspicion is correct, it's not working currently in the dev environment.

I think we can remove the config object entirely, as the favicon request is also not working. Removing the object (i.e. just `history()`) restores the function of `index-pub.html` for shared visualisations in the dev env

finnfiddle: is it ok now?

client/tools/devServer.js
```diff
@@ -0,0 +1,100 @@
...
+browserSync({
+  port: 3030,
+  ui: false,
+  open: false,
+  server: {
+    baseDir: 'src',
+
+    middleware: [
+
+      history(),
```
finnfiddle: is it ok now?

finnfiddle: @gabemart

gabemart:
@finnfiddle I'm afraid the new dev env is not working correctly with our multiple-entry-point webpack config.

We have two entry points, `app` (index.jsx) and `pub` (index-pub.jsx). The new dev env seems to be loading the code for *both* entry points at once, which is really weird.

This can be verified by inserting a log statement in `index-pub.jsx`. Anything in that file goes into the "pub" bundle which should only be loaded and executed on the `/s/${entityId}` route. This works on `develop`, but in this branch it gets executed on every page load on any route.
 
I'm going to dig a bit deeper to see if I can fix the config, otherwise let's catch up about this once you're back

gabemart:
I think to fix the public index, we need to edit `webpack.config.dev.js` to have two `HtmlWebpackPlugin()` entries, one for `[app]` and one for `[pub]`. That prevents the problem of both sets of bundles being loaded when only one should be loaded in the dev env.

Then, to get the `/s/${sharedId}` route to work, I think we can add
````
      history({
        rewrites: [
          { from: '^/s/.*$', to: '/../dist/index-pub.html' },
        ],
      }),
````

to `devServer`. Serving the pub index statically is obviously not perfect, but it gets touched so much less than the main bundle that I think it's fine for now,

finnfiddle:
@gabemart try the index-pub.html thing now - I think its resolved

gabemart:
Looks good to me!

Issue #998, opened 2017-08-11
Add a visualize button to dataset editor


Kiarii:
## Context
We need to shorten the path from a dataset editor to building a visualization, atm one has to take a detour via the Library/Collection

## Suggestion
+ Add a [visualize] action to dataset editor as illustrated below
+ If necessary a mouseover: "Create visualization"

![image](https://user-images.githubusercontent.com/12692444/29204490-11ba7d88-7e80-11e7-9d26-06c3020f6ab1.png)

PR #1282, opened 2018-02-19
Add a visualize button to dataset editor #998

finnfiddle:
#998

### DIFF
--- a/client/src/containers/Dataset.jsx
+++ b/client/src/containers/Dataset.jsx
@@ -98,11 +98,11 @@ class Dataset extends Component {
     }));
   }
 
-  handleNavigateToVisualise({ datasetId }) {
+  handleNavigateToVisualise() {
     this.props.dispatch(
       push({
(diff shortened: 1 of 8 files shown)


client/src/containers/Dataset.jsx
```diff
@@ -96,6 +98,15 @@ class Dataset extends Component {
     }));
   }
 
+  handleNavigateToVisualise({ datasetId }) {
```
gabemart: Perhaps this method could just grab the dataset ID from props? It might be a bit more direct, and probably less code than looking up the params in the child component and then passing the id back up. `getId(this.props.dataset)` would work I think

finnfiddle: youre right @gabemart Ill change it

finnfiddle: k done

gabemart:
LGTM

PR #1292, opened 2018-02-21
Allow backend/Dockerfile to build a production image

devurandom:
Steps to build a production image of the backend, before this change:
1. `lein uberjar`
2. `cp target/uberjar/akvo-lumen.jar .`
3. `docker build .`

After this change:
1. `lein uberjar`
2. `docker build .`

### DIFF
--- a/backend/.dockerignore
+++ b/backend/.dockerignore
@@ -1,2 +1,2 @@
 target/*
-
+!target/uberjar/akvo-lumen.jar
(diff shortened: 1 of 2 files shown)


kardan:
Hi, thanks for the work. Unfortunately, I think we have old artifacts in the repo that makes things confusing. We never build uberjars locally, but Travis our build tool does the job. Since the build system automatically deploys our new images I have to assume that things are working as expected and the suggested change is not needed. I'm sorry that our old docs was confusing. But closing this PR.

devurandom:
Indeed you execute step 2 of my "before" list in your `ci/build.sh` script: https://github.com/akvo/akvo-lumen/blob/develop/ci/build.sh#L19

PR #1300, opened 2018-02-23
#1744 added sourcemapped-stacktrace 3rd party lib

finnfiddle:
#1744
@gabemart needed to add it first in separate PR so that it could be served from develop branch


gabemart:
lgtm

PR #1302, opened 2018-02-23
Puppeteer sourcemaps from cdn to GitHub

finnfiddle:
#1744 

### DIFF
--- a/client/e2e-test/script-test.js
+++ b/client/e2e-test/script-test.js
@@ -60,7 +60,7 @@ async function test(page, shouldLogin) {
   }
   await page.waitForSelector('button[data-test-id="dataset"]', { timeout: TIMEOUT.waitFor });
   await page.addScriptTag({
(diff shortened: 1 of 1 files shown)


client/e2e-test/script-test.js
```diff
@@ -158,9 +171,19 @@ async function test(page, shouldLogin) {
...
+const getSourcemapsScript = () => new Promise((resolve) => {
+  https.get('https://raw.githubusercontent.com/akvo/akvo-lumen/develop/client/vendor/sourcemapped-stacktrace.js', (res) => {
+    res.setEncoding('utf8');
+    let body = '';
+    res.on('data', (data) => {
+      body += data;
+    });
+    res.on('end', () => {
+      sourcemapScript = body;
+      resolve(body);
```
gabemart: It's a small point, but perhaps we should change this to `resolve()` just to be a bit clearer that we're storing the value using the parent scope rather than the resolve value?

Issue #1304, opened 2018-02-26
Point border opacity in maps


gabemart:
# Context
We have a solid white border around map points to increase the contrast between points and their background and make them easier to see

# Problem or idea
When there is a high density of points and a small point size, the white border becomes visually overwhelming and it is extremely difficult to see any individual points

# Solution or next step
Removing the border or reducing the opacity too much will make individual points harder to see.

For now, let's reduce the point border by a moderate amount.

We can also reduce the border line-width to below 1.

PR #1306, opened 2018-02-26
[#1304] Reduce point border prominence in maps

gabemart:
* Reduce the visual weight of the borders of map points


finnfiddle:
LGTM @gabemart

Issue #1279, opened 2018-02-15
Dataset context menu unreachable


LarsSjogreen:
# Context
When the name of a dataset is too long, it's impossible to hit the context menu (and therefore rename or delete the dataset).

# Problem or idea
The name of the dataset is allowed to push outside the click area boundaries, taking with it the context menu and therefore making it impossible to hit because it is now under another dataset's hit box.

![image](https://user-images.githubusercontent.com/18190715/36259232-df2c3042-125d-11e8-9324-298cb2fb2129.png)

# Solution or next step
Maybe setting a maximum width for the dataset and cutting the name with ellipsis?

gabemart:
Specifically this looks like an issue when the dataset name is too long and without whitespace (long names with some spaces will still wrap).

We need to think about how to handle long names with spaces, but I think the answer to that issue will probably be limiting the length of names to some absolute number of chars

PR #1312, opened 2018-03-01
Dataset context menu unreachable #1279

finnfiddle:
Dataset context menu unreachable #1279


gabemart:
adding the ie10 failback as discussed

Issue #1250, opened 2018-01-31
Introduce a spinner when data is loading


nadiagorchakova:
# Context

Especially in countries with bad connectivity, users need visual feedback when something is loading.
When they see a white screen for a minute or two, they start doubting whether Lumen is still working and start to take actions, like re-loading the page, etc.

# Problem or idea
Perhaps we could introduce a spinner in parts of the UI where we know loading can take some time?

# Solution or next step

Something like that?

![image](https://user-images.githubusercontent.com/21127166/35620605-ea37939c-0681-11e8-9702-672ad0b12a69.png)

finnfiddle:
@nadiagorchakova this is a bit broad - are there specific places in the app that are especially bad?

nadiagorchakova:
I witnessed slow loading with large datasets (dataset editor), maps and even visualisations (where people saw a blank screen in the place of visualization preview

PR #1318, opened 2018-03-05
Introduce a spinner when data is loading #1250

finnfiddle:
#1250

### DIFF
--- a/client/src/reducers/datasets.js
+++ b/client/src/reducers/datasets.js
@@ -1,6 +1,7 @@
 import * as constants from '../constants/dataset';
 
 export const initialState = {};
+export const FETCHING = 'FETCHING';
 
 function createDataset(state, dataset) {
   const id = dataset.get('id');
@@ -44,9 +45,17 @@ function importDatasetSuccess(state, { importId }) {
 
 function saveDataset(state, dataset) {
   const id = dataset.get('id');
-  return Object.assign({}, state, {
+  return {
+    ...state,
     [id]: dataset.set('type', 'dataset'),
-  });
+  };
+}
+
+function fetchingDataset(state, action) {
+  return {
+    ...state,
+    [action.id]: FETCHING,
+  };
 }
 
 function saveDatasets(state, ds) {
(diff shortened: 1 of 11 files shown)


client/src/reducers/datasets.js
```diff
@@ -1,6 +1,7 @@
 import * as constants from '../constants/dataset';
 
 export const initialState = {};
+export const FETCHING = 'FETCHING';
```
finnfiddle: I know its a bit weird to use a string in place of a dataset object but I couldnt think of a better place to store this data (dataset is being fetched) in the state semantically-speaking. @gabemart Im open to suggestions

gabemart: I agree it feels a bit off. I'm also seeing some errors in the visualisation error when switching datasets, I think because there is some filtering code that is still expecting an object and chocking when it gets a string. I also think we should avoid filtering datasets out of the library view even when they are mid-fetch.

I can think of two alternate approaches. The first is to simply not store the `FETCHING` status at all, and infer it wherever we need it (i.e. something like `Boolean(dataset.get('rows'))` would tell us if the dataset is loaded or not. This has the advantage of not storing additional state, but it has the disadvantage that we can no longer infer if we *need* the dataset in question (whereas storing an explicit `FETCHING` state at the point of requesting a dataset also implicitly records that something requires the dataset to be loaded).

The second approach, which I think I prefer, is to move the `FETCHING` state to the existing `status` property of the dataset object. Something like

````
function fetchingDataset(state, action) {
  const { id } = action;

  return {
    ...state,
    [id]: state[id] ? state[id].set('status', FETCHING) : null,
  };
}
````

This would allow us to revert the changes to `LibraryListing.jsx`. 

It feels a bit impure, because `FETCHING` isn't really a property of the dataset but rather a property of the client - but I think on balance it has the fewest disadvantages. The changes to the containers to check the status should hopefully be fairly minor.

gabemart:
@finnfiddle one small problem is that the spinner shows when the user creates a *new* visualisation, before selecting any options, which might be a bit misleading (because nothing is loading at this point). Specifically, it's showing the spinner whenever `shouldRender` in `VisualisationPreview.jsx` returns `false`.

The logic in `shouldRender` is out of date anyway (e.g. it returns `false` for pivot tables when the dataset rows have not been loaded, but the dataset rows are no longer required for pivot tables because we aggregate on the backend)

To get this to work "properly", we would have to refactor `shouldRender` to separate the "dataset rows not loaded" case (and show the spinner) from the "invalid or incomplete spec" case (and show nothing, or some help text).

But I think we will have to refactor this soon anyway once all aggregations are on the backend, so I would suggest that for now we just revert to rendering `null` whenever `shouldRender` returns false.

Thoughts?

finnfiddle:
@gabemart ok I made the change you requested

Issue #1012, opened 2017-08-18
Help button in publicly shared visualisations and dashboards


nadiagorchakova:
# Context
Thomas brought to our attention that the Help button also appears on publicly shared visualisations and dashboards. Example: https://demo.akvolumen.org/s/gGMQtMkk2Fg

<img width="1277" alt="screen shot 2018-03-06 at 11 41 22" src="https://user-images.githubusercontent.com/21127166/37028097-9a0b835c-2133-11e8-83c7-9fe1b5d1e282.png">

# Problem or idea
The viewers of publicly shared visualisations and dashboards are not interested in our support channel, as they might not even be users in Lumen. They just look at publicly shared dashboards and visualisations. Therefore they don't need to see the 'support' button'.

# Solution or next step
Help button is enabled through a third party provider, Support Hero. 
How much can we customise it?

PR #1321, opened 2018-03-06
Help button in publicly shared visualisations and dashboards #1012

finnfiddle:
#1012


gabemart:
Checked and tested

Issue #1294, opened 2018-02-21
Dating of changed maps and datasets in Lumen library


nadiagorchakova:
Created via Reamaze:

Link: https://akvoo.reamaze.com/admin/conversations/akvolumen-support-request-number-94722-from-thomas-at-akvo-dot-org
Assignee: Unassigned

Message:

Request Info  
Device:  Personal computer  
Browser:  Safari 11.0.3  
OS:  OS X 10.11.6  
URL:  https://gt20.akvolumen.org/library  
 
 
 1- What do you need help with?  
I have a bug  
 
 2- What is the name of your instance? (xx.akvolumen.org)  
gt20.akvolumen.org  
 
 3- What is the name of your dataset, visualisation or dashboard?  
Dunkern  
 
 4- What did you expect to happen?  
When changing a layer on a map, the map is displayed under todays date in the Library. When doing a new derived column in a dataset it is still presented under the creation date. This is inconsistent.  
 
 5- What actually happened?  
When updating a dataset it should be consistent with how it is shown in the Library  
 
 6- What is your full name?  
Thomas Bjelkeman-Pettersson  
 
 7- Which Akvo Hub do you usually work with?  
Europe  
 
 Tell us more about your request  
A minor inconsistency in the UI. No need to give feedback to me, but put it in the list of things for the UI/UX overhaul.

PR #1322, opened 2018-03-06
Dating of changed maps and datasets in Lumen library #1294

finnfiddle:
#1294


iperdomo:
It turns out that I was completely wrong. There is no _easy_ way of running multiple DML statements as single hugsql function.

Issue #1303, opened 2018-02-26
Flow import error 'null value in column Identifier' 


nadiagorchakova:
# Context
When importing a Flow form, West Africa team gets the following error:

<img width="568" alt="screen shot 2018-02-26 at 11 06 01" src="https://user-images.githubusercontent.com/21127166/36674999-0dccf64a-1b08-11e8-9376-36255e143da3.png">

Both Gabe and myself checked raw data and couldn't find any issues with Identifier column or the specified row (and there are no null values in Identifier column)

# Problem or idea

The following Form form produces error at import:

https://mea-bf.akvoflow.org/admin/

Folder: 2. BOUBLE DU MOUHOUN-->Fiche de Collecte--> Fiche 6a: Inventaires des ouvrages ..
and the form Enregistrement AEP (Id 3450046)

I managed to reproduce the error in my own instance too.

# Solution or next step
@kardan Gabe and I had a look at raw data but we haven't noticed anything fishy...
Also, it's possible to load the same form as a csv file, so the problem seems to be with Flow import

kardan:
Me and Iván debugged this for a bit. For some reason data-points at https://github.com/akvo/akvo-lumen/blob/develop/backend/src/akvo/lumen/import/flow_v3.clj#L58
does not always include a point with the id from the form-instance
 https://github.com/akvo/akvo-lumen/blob/develop/backend/src/akvo/lumen/import/flow_v3.clj#L61. Looking in the App Engine data-store we do have a point with the data-point-id. I have the feeling that somewhere from the data store to above lines in Lumen the data point disappears.

kardan:
The issue is that we map over form-instances, using the dataPointId from the form instance to pull info about the data point from a list of data points. This list of data points has been crafted earlier based on the survey from the Flow API. The problem is that 2 form instances points to data points that _belongs_ to another survey. Hence the data point in question is not included in the list of data points. This results in Lumen trying to insert null data as the identifier. 

@muloem, @iperdomo and I concluded that to fix this properly we should fix the data on the Flow side.

PR #1324, opened 2018-03-07
Issue/1303 flow import

kardan:
New and noteworthy
[#1303]  Handle unsuccessful Flow imports better

### DIFF
--- a/backend/src/akvo/lumen/boundary/error_tracker.clj
+++ b/backend/src/akvo/lumen/boundary/error_tracker.clj
@@ -7,15 +7,18 @@
 (defprotocol IErrorTracker
   (track [this error]))
 
+(defn event-map [error]
+  (let [text (str (ex-data error))]
+    {:extra {:ex-data (subs text 0 (min (count text) 4096))}
+     :message (.getMessage error)}))
+
 (extend-protocol IErrorTracker
 
   SentryErrorTracker
   (track [{:keys [dsn]} error]
     (let [app-namespaces ["org.akvo" "akvo"]
-          event-info (raven-interface/stacktrace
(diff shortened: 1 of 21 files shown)


backend/src/akvo/lumen/boundary/error_tracker.clj
```diff
@@ -0,0 +1,17 @@
...
+  (:import [akvo.lumen.component.error_tracker SentryErrorTracker LocalErrorTracker]))
+
+(defprotocol IErrorTracker
+  (track [this error]))
+
+(extend-protocol IErrorTracker
+
+  SentryErrorTracker
+  (track [{:keys [dsn]} error]
+    (raven/capture dsn error))
```
iperdomo: There is something i don't fully _see_ ... It's supposed that `error` is a map that contains at least one key `:message`.
See the implementation of the ring middleware:
https://github.com/sethtrain/raven-clj/blob/1.4.3/src/raven_clj/ring.clj#L5-L9

In fact `raven/capture` states:
>  "Send a message to a Sentry server.
>  event-info is a map that should contain a :message key and optional
>  keys found at https://docs.getsentry.com/hosted/clientdev/#building-the-json-packet"


https://github.com/sethtrain/raven-clj/blob/1.4.3/src/raven_clj/core.clj#L46-L50

So just passing a Throwable doesn't follow that contract?

iperdomo: There is an a Sentry clojure wrapper on the official Java SDK .. https://github.com/getsentry/sentry-clj
This one _does_ support passing a `{:throwable e}` as parameter

iperdomo:
:+1:

Issue #1347, opened 2018-03-16
Cannot read property 'collectionId' of null - when activeModal = null


finnfiddle:
# Context
https://sentry.io/akvo-foundation/test/issues/495075099/?referrer=slack

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

PR #1348, opened 2018-03-16
Cannot read property 'collectionId' of null - when activeModal = null…

finnfiddle:
#1347


gabemart:
LGTM!

Issue #1346, opened 2018-03-15
Object.values undefined on old browsers


finnfiddle:
# Context
https://sentry.io/akvo-foundation/prod/issues/494999700/

# Problem or idea
What is the cause?

# Solution or next step
We need to add the `es7.object.values` core js polyfill

PR #1349, opened 2018-03-16
Object.values undefined on old browsers #1346

finnfiddle:
#1346

### DIFF
--- a/client/src/polyfill/polyfill.js
+++ b/client/src/polyfill/polyfill.js
@@ -23,6 +23,7 @@ const polyfill = (callback) => {
       require('core-js/es6/array');
       require('core-js/es6/object');
       require('core-js/es6/number');
+      require('core-js/modules/es7.object.values');
       require('intl');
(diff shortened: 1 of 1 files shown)


gabemart:
@finnfiddle Perhaps we should also add a `'values' in Object` test to the `browserSupported` const?

finnfiddle:
@gabemart whats wrong with the current approach?

gabemart:
@finnfiddle we conditionally load the combined polyfill bundle based on the `browserSupported` boolean

````
  const browserSupported = Boolean(
    'fetch' in window &&
    'assign' in Object &&
    'Intl' in window &&
    'filter' in Array.prototype
  );
````
 This pr updates the polyfill bundle to include the object values polyfill, but doesn't update the `browserSupported` const. If a browser supports `fetch`, `assign`, `Intl` and `filter`, but doesn't support `Object.values`, the polyfill bundle won't be loaded

finnfiddle:
@gabemart ok got it

Issue #1365, opened 2018-03-27
Implementation of a new vis library


nadiagorchakova:
# Context

based on https://github.com/akvo/akvo-lumen/issues/1354

# Problem or idea
we need to reach feature parity + colour picker + other issues (to be discussed) that will enrich user experience

# Solution or next step

finnfiddle:
## TODO

[ ] Pie
[ ] Print styles
[ ] Themes
[ ] conditional interactivity
[ ] Bar
[ ] Scatter
[ ] Donut
[ ] Line
[ ] Area

finnfiddle:
# Proposed data formats

## Pie

```
{
  data: [
    { key: 'abc', value: 100 },
    { key: 'xyz', value: 200 },
    ...
  ]
}
```

gabemart:
I'm going to leave my thoughts in multiple comments rather than one huge uber-comment to make them a bit more digestible, but the ordering of the comments has no significance.

---

Thought one:

I'm a big fan of the proposed approach of using the implicit ordering of elements in an array to separate out different axis values, without having to use labels like "x-axis" which are clunky and visualisation-specific.

But I'm not sure about the name `values` in the `metadata` object. There's an appealing purity to using the same key as we use in the `data` object, but it looks to me that the `values` array in the `metadata` object really functions as an ordered list of axis-specific metadata, using the order as the axis identifier (like we do with the values array for each datum). I think a name like `axes` would be clearer and better reflect the purpose of the array.

I also think that, almost inevitably, we will end up adding extra information to the `metadata` object over time, and if any of that extra information is not axis-specific, it might be confusing to have a `values` key on the same level as a `datasetVersion` or `serverTime` or whatever-we-end-up-needing key, because it's all just metadata.

I would propose:

````
  metadata: {
    axes: [
      { label: 'Temperature' },
    ]
  },
````

**Comment by finnfiddle** [comment]
just copying and pasting from slack for reference:
```
its like for a bar or pie you have one series containing single values;
for a scatter you have one series containing multiple values;
for a stack you have multiple series containing single values;
and for a multiline chart you have multiple series containing multiple values
```
so that I can propose a different format in this context.

@gabemart regarding `metadata.axes` I get what youre saying and I dont have much preference about this but I think we could think of a better word than `axes` - let's discuss.

finnfiddle:
# Proposed Formats V2

## Pie

```javascript
{
  series: [
    {
      key: 'hum',
[...]

PR #1380, opened 2018-04-18
Implementation of a new vis library #1365

finnfiddle:

#1365


gabemart:
Tiny point, but we could we rename `colorMappings` to `colorMapping`? We use the names `pointColorMapping` and `shapeColorMapping` for maps, and I think it will be confusing to have some singular and some plural

gabemart:
Perhaps we should move `ColorPicker.jsx`, `` and `Popover.jsx` and `ResponsiveWrapper.jsx` out of the root components directory. They could go in `common` maybe?

PR #1384, opened 2018-04-24
Fix CIDER find-var, can't jump to definition with a docker-nrepl connection

### LINKED ISSUE #1383: CIDER find-var, can't jump to definition with a docker-nrepl connection
# Context
Following the [development instructions](https://github.com/akvo/akvo-lumen/blob/66ba2cc88800ed101716d6222d5f74af30a6b0f9/README.dev.md ) we get a nrepl runnning  on `localhost:47480`, so from Emacs/CIDER we need to `cider-connect` with these args
# Problem or idea
The problem arises when we want  to use CIDER basic interactive progamming code navigation as `M - .` => `cider-find-var` that we get `No source location...` error 

## What is the cause?
CIDER tries to find the source in nrepl docker [path](https://github.com/akvo/akvo-lumen/blob/66ba2cc88800ed101716d6222d5f74af30a6b0f9/backend/Dockerfile-dev#L10) `/app/src` instead of `/src`

# Solution or next step
we could overwrite `cider-common.el` `cider--jump-to-loc-from-info` function  to adapt expected path `/app/src` to `/src` (and the same with `app/test` and others)
To do that we can use `.dir-locals.el` feature of emacs/elisp

tangrammer:
Fix #1383 

Relevant installation info in #1383  specially this [one](https://github.com/akvo/akvo-lumen/issues/1383#issuecomment-383894327)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ backend/*.jar
 backend/*.class
 backend/.lein-*
 backend/.nrepl-*
-backend/.dir-locals.el
 backend/profiles.clj
 backend/dev/resources/local.edn
 backend/dev/src/local.clj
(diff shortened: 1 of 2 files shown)


iperdomo:
:clap:

Issue #1046, opened 2017-09-01
Password-protected dashboard links


nadiagorchakova:
# Context

To ensure protection of sensitive data, a few partners are asking for a functionality to set passwords on publicly shared dashboards. It's also in line with GDPR legislation that we need to help partners protect their sensitive data.

# Problem or idea

When a dashboard is shared publicly via the 'share' option, anyone who obtains the link can view it. That is not safe enough as some survey data might contain personal data of respondents like religion, sexual orientation etc. If this data (or visualisations) that can identify certain people or communities gets into the wrong hands, the respondents might be in danger or persecution. There are other cases of possible dashboard data mishandling too. 

# Solution or next step
Currently, the dashboard sharing window looks like that:
<img width="986" alt="screen shot 2018-04-19 at 11 17 38" src="https://user-images.githubusercontent.com/21127166/38982885-51c8c480-43c3-11e8-8602-0deb57441e13.png">
It should be enhanced with password setting functionality.
@Kiarii, if you have time, would be helpful to hear your opinion on best UX practices for that.

nadiagorchakova:
yes!

Kiarii:
I think this shd be done similarly to the visualization's pwd protection; https://invis.io/46F1NKMEV#/262731142_2a-_Share_Dialog_Def for the sake of some consistency

nadiagorchakova:
thanks @Kiarii! this is helpful

kardan:
### Backend Todo
- [ ] Implement/api/shares/ :PUT endpoint that takes {"password": "the-password}
- [ ] Remove Tardis from the share db table
- [ ] Migrate share table to have a password column 
- [ ] Store password hashed on shares table 
- [ ] Add auth logic on /shares/<id> GET to if needed look for a {"X-Password" "the-password"} header.

PR #1392, opened 2018-04-26
Password protected dashboard links #1046

finnfiddle:
#1046


client/src/components/common/ErrorScreen.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+import './ErrorScreen.scss';
+
+const getTitle = (title, code) => title || ({
+  404: 'Not Found',
+  403: 'Access Restricted',
+})[code];
+
+const ErrorScreen = ({ title, children, code }) => (
+  <div className="error-screen">
```
gabemart: It's a small thing but we have a convention in Lumen that the parent element of each component has the name of the component as a className - i.e. `ErrorScreen` rather than `error-sceen` - and that the stylesheet for the component is scoped to that className

gabemart:
Should the password protection be working locally? Setting a password doesn't seem to change anything when viewing a shared dashboard locally

finnfiddle:
Did you reset the backend? About the css thing ill change it

Regards,
Finn

finnfiddle:
@gabemart I fixed the className. you need to `(reset)` the backend for it to work I think

gabemart:
There also seems to be a `PUT` when the "password protected" toggle is first switched to "on" (i.e. before user has entered password or hit "save")

the backend responds with `{"message":"Can't enable protection without valid password"}`

finnfiddle:
@gabemart about this one "

There also seems to be a PUT when the "password protected" toggle is first switched to "on" (i.e. before user has entered password or hit "save")the backend responds with {"message":"Can't enable protection without valid password"}
--

"
This is because we let the user toggle the privacy on and off and if they have previously set a password then it will just reactivate it. However, we have no way of knowing from the frontend whether a password has been set previously or not so we just send a request anyway and let it fail silently if no password has been set previously.

gabemart:
Looks good to me too 👍

Issue #878, opened 2017-06-21
No prompt for unsaved changes


nadiagorchakova:
Created via Reamaze by Joy:

Link: https://akvoo.reamaze.com/admin/conversations/no-prompt-for-unsaved-changes
Assignee: Nadia Gorchakova

Message:
CRITICAL
There is no prompt to save changes when you move away from a Lumen dashboard/visualisation if there are unsaved changes.
This leads to a lot of re-work and even after a few months use loosing a lot of work because of it.

So definitely a UI interaction issue.

The fine print in the top left does not help.

PR #1394, opened 2018-04-30
No prompt for unsaved changes #878

finnfiddle:
#878


gabemart:
@finnfiddle this works well but can you help me understand why `NavigationPrompt` takes a function as a child rather than the bare components? I.e. why we use `return this.props.children()` instead of `return this.props.children`?

finnfiddle:
@gabemart youre right it was unnecessary - Ive changed it

gabemart:
Ah ok just wanted to check 👍

gabemart:
LGTM 👍

Issue #1087, opened 2017-09-27
Renaming in maps editor


nadiagorchakova:
# Context
As shared by Thomas. 

Renaming in maps - two actions when one should be enough - When renaming a map layer it takes two actions to save. One to click the check mark and one to click save. One should be sufficient.

Also giving a name or renaming a layer in map has different UI than other title elements in Lumen.

<img width="1265" alt="screen shot 2017-09-27 at 16 37 24" src="https://user-images.githubusercontent.com/21127166/30919590-22895bb0-a3a2-11e7-81e5-0cfba6b1efd1.png">

finnfiddle:
@nadiagorchakova I dont think the layer names can be so easily separated from the rest of the map - ie saved independently. If the user has other changes then those will get saved as well. we could do it theoretically but its going to not be a very elegant solution I think. it seems that this is more an issue of there not being autosave - or not?

nadiagorchakova:
I think what Thomas meant here is that you need to use the ... button in order to give a name to an Untitled map layer or to rename it. The logic for naming or renaming the map itself or other visualisations types in Lumen is different - you just need to click on the 'Untitled' and start typing. I think I agree with him that it's not very intuitive.

Does it make more sense now...?

![naming of map layers](https://user-images.githubusercontent.com/21127166/37293978-1e287aec-2615-11e8-82bf-a09c9ab42dcd.gif)

finnfiddle:
@nadiagorchakova but then how do you get to editing the layer? Because currently if you click the layer name it takes you the the layer editor

nadiagorchakova:
hm hm, good point @finnfiddle 
I'll ask @Kiarii to have a look and share his opinion

Kiarii:
nb: the *map title* on the other hand is only an **editable element**, the *layer name* element is both **editable and clickable** - it cannot have these two properties simultaneously.. (so much for the double split experiment)  

one option how we could achieve a uniformity is by separating the clickability and editability properties of the *layer name* as follows: 
- user clicks add new layer
- new layer has a default name "untitled layer x"
- "untitled layer x" is only **clickable** 
- user clicks on "untitled layer x" 
- they navigate to the "untitled layer x" editing page where
- "untitled layer x" becomes an **editable** element     

Of course, this also comes with its own setbacks: to rename a layer, the user has to open it.. to rename multiple layers, the user has to go back and forth.. one might also wonder why a layer has an edit action **delete** and not another **edit** and so on..

[...]

PR #1400, opened 2018-05-02
Renaming in maps editor #1087

finnfiddle:
#1087

### DIFF
--- a/client/e2e-test/script-test.js
+++ b/client/e2e-test/script-test.js
@@ -148,7 +148,7 @@ async function test(page, shouldLogin) {
   console.log('Typing visualisation name...');
   await page.type('input[data-test-id="entity-title"]', `Visualisation of ${datasetName}`);
   console.log('Saving visualisation...');
-  await page.click('button[data-test-id="save-changes"]');
+  await page.click('body');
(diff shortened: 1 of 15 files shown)


finnfiddle:
@kardan @gabemart @nadiagorchakova @tangrammer  I made it so that if it fails to save for some reason (like being offline) then it will show a save button and retry over and over with longer intervals each time. if you want think it should work another way let me know

kardan:
I was thinking that I was assigned as reviewer to, hence tested it out. When I create a new map I directly get a modal asking me if I want to leave the page. This is in Chrome. That looks like a bug to me

gabemart:
There's an existing issue with dashboards that makes the autosave a bit weird. 

If you create a visualisation and don't give it a name, it autosaves as "untitled visualisation".

But if you create a dashboard and don't give it a name, the autosave doesn't trigger until you give it a name, which is a bit unintuitive for the user.

So even though the behaviour (dashboards won't save without an explicit title) isn't introduced by this pr, I think we should fix it before we merge this

finnfiddle:
I made the changes requested

gabemart:
👍 lgtm

PR #1411, opened 2018-05-09
Autosave fixes #1087

### LINKED ISSUE #1087: Renaming in maps editor

finnfiddle:
#1087 


gabemart:
LGTM

Issue #1416, opened 2018-05-11
Start colour in raster map preview different from chosen


nadiagorchakova:
# Context
In Theme tab in map layer it's possible to pick a custom start and end colour of the raster.

# Problem or idea
When I picked a start colour different from default, the map preview on the right showed a wrong colour.
Looks like it's showing the previous colour choice.

![raster colours](https://user-images.githubusercontent.com/21127166/39925585-89e6c02c-552c-11e8-8a8b-ba7dae41181d.gif)

# Solution or next step
How could we solve it?

gabemart:
The problem of the baselayer not updating is related to the same state propagation issue

PR #1419, opened 2018-05-14
Start colour in raster map preview different from chosen #1416

finnfiddle:
#1416 

### DIFF
--- a/client/src/components/visualisation/VisualisationEditor.jsx
+++ b/client/src/components/visualisation/VisualisationEditor.jsx
@@ -163,7 +163,7 @@ export default class VisualisationEditor extends Component {
 
         if (!this.state.visualisation || !this.state.visualisation.datasetId) {
           // Update immediately, without waiting for the api call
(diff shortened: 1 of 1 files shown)


gabemart:
@finnfiddle I'm still seeing the original problem when I run this branch locally

gabemart:
I'm pretty confident it's a race condition to do with saving happening simultaneously to a change to the spec. This is confirmed by the fact that wrapping the save callback in a `setTimeout()` at 1ms "fixes" the problem.

Somehow the close-to-simultaneous updates to the visualisation object are breaking the logic in the Maps component that controls when to update the map with the new `layerGroupId`.

This shouldn't happen. One way of fixing this would be to add some extra logic to check for layergroupid changes, but I'd prefer to first work out how exactly this is happening at the moment

finnfiddle:
@gabemart try now

gabemart:
@finnfiddle could you test my commit?

Issue #1422, opened 2018-05-14
Back button in vis doesn't bring back to Collection view


nadiagorchakova:
# Context
When I go to my collection and view a vis from there, I expect to go back to my collection view.

# Problem or idea
What happens is that when I click Back button in the vis editor, I'm brought back to the library and not to my collection, where I was working. 
![back to collection](https://user-images.githubusercontent.com/21127166/39999557-a4ecc5e2-5789-11e8-8d1f-b1d1e2463a60.gif)

# Solution or next step
Perhaps it's more intuitive to be brought back to the collection view, if that's where I opened my vis from.

PR #1429, opened 2018-05-15
[#1422] Use router.goBack for Header back button

gabemart:
#1422 

### DIFF
--- a/client/src/components/Library.jsx
+++ b/client/src/components/Library.jsx
@@ -272,7 +272,7 @@ class Library extends Component {
           }}
           onCreate={(type) => {
             const { params } = this.props;
-            const meta = { collectionId: params.collectionId };
+            const meta = { collectionId: params.collectionId, from: 'library' };
             if (type === 'dataset') {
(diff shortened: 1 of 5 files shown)


finnfiddle:
this works fine when you go forward then back but if you go directly to a page like for a visualization (ie copy and paste the url into a new tab) and then click back it takes you away from lumen altogether. maybe we need to check if there is a previous state in the history and if not then go to the library..?

finnfiddle:
looks fine to me but feels like a bit of a hack. Im surprised there isnt an easier way to do this! I guess its for security reasons. I did find [this](https://www.npmjs.com/package/react-router-last-location) approach though which seems less hacky - its only for react router v4 though so I guess we'd need to reimplement it ourselves. I think for the moment its fine as it is but if we start needing more back buttons around the place maybe its a good idea to use an HOC or something

gabemart:
Yeah I had the same thought, but as you say understandable from a security perspective.

agreed, if end up keeping track of the last location on more routes then a more generic solution would make sense

PR #1437, opened 2018-05-22
Fix #1225 tardis fails when alter public column tables

### LINKED ISSUE #1225: Review Tardis
# Context
On a few tables we use Tardis to keep track of history.

# Problem or idea
We can't add columns and then write data to these tables since Tardis relies on the table column order. We should also review how Tardis affect requirements on user history & GDPR.

# Solution or next step
Review it.

tangrammer:
Tested with https://gist.github.com/tangrammer/93dcb452d193cd4185097b7b06cf6239


tangrammer:
Instructions to test:
```

**Comment by tangrammer** [comment]
Although version seems to work I'm pretty sure that @iperdomo could add a few improvements to proposal changes :)

File: backend/test/akvo/lumen/db/tardis_test.clj [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+(use-fixtures :each tenant-conn-fixture)
+
+(deftest tardis-with-alter-table
+  (testing "with current data in public schema table, we alter table definition. Functionality keeps the same"
+    (is (= '() (get-data *tenant-conn* {:table-name "data_source"})))
+    (is (= 1 (insert-data-source *tenant-conn* {})))
+    (try
+      (insert-altered-data-source *tenant-conn* {:table-name "data_source" })
+      (assert false)
+      (catch Exception e (is (str/includes? (.getMessage e) "column \"test\" of relation \"data_source\" does not exist") )))
```
iperdomo: i think we want to use `thrown-with-msg?` instead of catching the `Exception`
> 
> "(is (thrown-with-msg? c re ...))" does the same thing and also
> tests that the message on the exception matches the regular
> expression re:
>
> (is (thrown-with-msg? ArithmeticException #"Divide by zero"
>                      (/ 1 0)))

https://clojure.github.io/clojure/clojure.test-api.html

iperdomo:
:+1:

Issue #675, opened 2017-03-29
Setting up analytics (Piwik)


nadiagorchakova:
# Context

There are two use cases for setting up Piwik analytics
- Akvo internally would like to track user behavior to provide better service and improve Lumen. For this case, it would be interesting to track the library page inside each partner's instance, e.g. https://demo.akvolumen.org/library and Lumen landing page.
- partners will eventually be interested in statistics on their publicly shared visualisations and dashboards. I imagine they would be interested to know stats on page views, geo location of viewers, etc. 
- partners might also be interested in stats on visualisations and dashboards embedded on their own websites

# Problem or idea

# Solution or next step
To configure Piwik, we need to install Piwik JavaScript tracking tag. More on that [here](https://piwik.org/docs/installation/#configuring-piwik).

Ideally, visualisations and dashboards should have their title reflected in the URL. Otherwise, it might be difficult to use Piwik analytics. For the first implementation, however, we could probably do without it.

kardan:
We should hook up Piwik. There are two different was to go about it. 

- Global: track usage in one "bucket" and then we only need to setup one thing one time in Piwik and add a js lib to our client app (+ an id). This is easiest to setup and makes global queries simple.
- Tenant: track in one bucket per tenant. This will require us to add the tenant to Piwik every time we create a new tenant. It will make stats per tenant easiest but more difficult to get the "full lumen" stats.

nadiagorchakova:
When I think about use cases, we'll need to track number of clicks on all publicly shared dashboards as part of PPP4 grant reporting. For this case, the 'global' option seems to be more suitable.

On the other hand, I foresee some partners would want to see their individual tenant stats. Will we be able to extract these stats from our 'global bucket'?

nadiagorchakova:
shall we at some point return to this issue @kardan?
It would be nice to track (global) usage of Lumen features to see which features are most popular (most used), and which are not. Could inform our decisions going forward.

I must say that my earlier assumption that partners would also be interested in their own instance stats seems to be wrong.

so we would need to define events that would be interesting for us to track https://matomo.org/docs/event-tracking/#tracking-events (as shared by @iperdomo )

iperdomo:
Just as a side-note: when _track the usage of Lumen features_ we need to make sure all data sent fo Piwik (now Matomo) is anonymous. Let's have GDPR always in mind :)

Lynngre:
@iperdomo & @nadiagorchakova: 

[...]

PR #1438, opened 2018-05-25
Setting up analytics (Piwik) #675

finnfiddle:
@iperdomo need to check Im doing things correctly on the backend for the env vars
@gabemart please check that you agree with how its been implemented

### DIFF
--- a/backend/src/akvo/lumen/endpoint/env.clj
+++ b/backend/src/akvo/lumen/endpoint/env.clj
@@ -8,6 +8,7 @@
       (cond-> {"keycloakClient" (:keycloak-public-client-id config)
                "keycloakURL" (:keycloak-url config)
                "flowApiUrl" (:flow-api-url config)
+               "piwikSiteId" (:piwik-site-id config)
                "tenant" (:tenant request)}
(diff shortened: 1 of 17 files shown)


backend/src/akvo/lumen/endpoint/env.clj
```diff
@@ -8,6 +8,7 @@
       (cond-> {"keycloakClient" (:keycloak-public-client-id config)
                "keycloakURL" (:keycloak-url config)
                "flowApiUrl" (:flow-api-url config)
+               "piwikSiteId" (:piwik-site-id config)
```
dlebrero: I dont see the value of piwik-site-id defined for production. Dont we want to do any tracking in production?

finnfiddle: I think we need a different value for each tenant. how can this be done?

dlebrero: If that is the case, the value should not come from the config but from the db.

Lets have a chat on Monday

gabemart:
Frontend looks good to me 👍

PR #1449, opened 2018-06-07
Issue/1408 delete pending raster

kardan:


backend/src/akvo/lumen/lib/raster.sql
```diff
@@ -66,3 +66,12 @@ SELECT stats.count, stats.sum, stats.mean, stats.stddev, stats.min, stats.max
 -- :name delete-raster-by-id :! :n
 -- :doc delete raster
 DELETE FROM raster_dataset WHERE id=:id;
+
+
+-- :name delete-raster-columns :! :n
```
tangrammer: ` grep -r "delete-raster-columns" ./ ` => `.//backend/src/akvo/lumen/lib/raster.sql:-- :name delete-raster-columns :! :n`
seems that this sql function is not currently referenced in any place? 
still would you like to have it in `develop`?

Issue #1381, opened 2018-04-19
Begin using new visualisation library


gabemart:
# Context
We have decided to [switch to a new viz lib](https://github.com/akvo/akvo-lumen/issues/1365) and to perform aggregations on the backend and only send aggregated data to the client. 

This issue will cover implementing those decisions, building on the work already done.

We have decided to tackle visualisations one-by-one.

This probably means a PR for each visualisation type that adds the appropriate backend aggregation and "switches on" that visualisation in the client using the new lib.

This approach is intended to break this large task down into more manageable chunks, and to make finding bugs and handling feedback for each new visualisation type more manageable. 

- [ ] line/area chart
  - [x] Date x-axis vs number x-axis
  - [x] Minor styling/positioning points
  - [x] keys warning on Collection
  - [x] Down-sampling
- [ ] scatter chart
- [ ] pie/donut chart
- [ ] bar chart

gabemart:
# Line Charts

Here are some potential issues with line charts:

## Date x-axis vs number x-axis
Right now we assume x-axis will always be a date-type. On text there is at least one line chart that has a number-type x-axis. 

With new charts, if you try to use a number column for the x-axis, the tick labels will display incorrectly because we are using functions that cast the value to a date

To make the new line chart work with number-type x-axis, the client would need to know explicitly know if the x-axis was a date or number type (perhaps backend could include this in `metadata` response) and we would need to handle things a bit differently for number types

The other option would be to decide that line and area charts are only for date-types on the x-axis, and enforce this in the editor (although that raises the question of what happens to any existing line or area charts with a numeric x-axis that already exist in production)

## Minor styling/positioning points
Y-axis tick labels aren't vertically centered on their tick marks
![image](https://user-images.githubusercontent.com/2589977/39529098-c9e75b98-4e1d-11e8-9ca1-d9b85875732e.png)

---

The x-axis label isn't horizontally centered on the x-axis
![image](https://user-images.githubusercontent.com/2589977/39529245-1c626ab6-4e1e-11e8-9bf2-aa61f8b4e625.png)

---

At large widths, there might be a bit too much negative space at the sides (although this is subjective)
![image](https://user-images.githubusercontent.com/2589977/39530121-30806898-4e20-11e8-8005-194118aff6a0.png)


---

gabemart:
[...]

PR #1453, opened 2018-06-11
[#1381] Backend aggregations and new visualisation library

gabemart:


backend/src/akvo/lumen/lib/aggregation/pie.clj
```diff
@@ -17,10 +17,24 @@
...
-    (lib/ok {"metadata" {"bucketColumnTitle" bucket-column-title
-                         "bucketColumnType" bucket-column-type}
-             "data" (mapv (fn [[bucket-value bucket-count]]
-                            {"bucketValue" bucket-value
-                             "bucketCount" bucket-count})
-                          counts)})))
+        counts (run-query tenant-conn table-name bucket-column-name filter-sql)
+        max-segments 50]
+    (if (> (count counts) max-segments)
+      (lib/ok
```
kardan: Does this mean the client expect a 200 response?

gabemart: Yeah, I wasn't sure about this. Currently the client expects a 200. Perhaps another type of response would be more semantically correct. It's not really an error or a bad request. I gues a `413 - Payload Too Large` might fit, but my impression was that's more for situations where the request itself is too large. A 4xx code feels weird because the client can't know ahead of time anything about the aggregation. Maybe a generic `500`? 

A `418` also seems a possibility

kardan: 418 sounds like a good option if we want to be semantically correct, otherwise, I'm fine with whatever. I mean it's a bit odd to have an error with a 200 :-). I would have no problem responding with a 400, but it's as I said not the end of the world. Noticed that some of the other aggregations don't have multiple response options at all. Since the client does a fair bit of checking I'm not sure we need to cover all cases. But maybe we can be a bit pragmatic and expand where needed, moving forward.

gabemart: Yes, for now we only respond with an error when the aggregation response will result in too many elements for a given chart type, and for now this only applies to pie/donut charts and bar charts.

I changed the response to a generic "bad-request" as already implemented and used in other places. After some thought this seems better than a 200, which I agree is weird

PR #1468, opened 2018-06-14
[#1467] Change linter to only autofix when running lint:fix

### LINKED ISSUE #1467: Don't auto-fix when linting client
Currently we have `npm run lint` set to autofix where possible

I think this is leading to slight weirdness - if you merge something that is auto-fixable without actually running lint locally, it will pass CI (because the CI will auto-fix when it runs the lint step) but the actual code as-merged is not linted

I suggest we go back to the old system of `npm run lint` simply running the linter and returning a pass/fail, and `npm run lint:fix` running the linter with the `--fix` flag

gabemart:
#1467 

@finnfiddle what do you think about this?

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -11,7 +11,8 @@
   "scripts": {
     "preinstall": "node tools/nodeVersionCheck.js",
     "start": "babel-node tools/devServer.js",
-    "lint": "esw webpack.config.* src tools --color --fix --ext jsx --ext js",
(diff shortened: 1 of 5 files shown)


gabemart:
Hrm, it seems like I'm getting different lint results locally (including with `docker exec`) than in CI. As such my explanation above of what's going is probably totally wrong.

I've tried nuking node_modules and the results are the same. I'll look into it some more tomorrow.

PR #1476, opened 2018-06-19
#675 tracking map layer type renders piwik

finnfiddle:


client/src/components/charts/MapVisualisation.jsx
```diff
@@ -419,6 +423,16 @@ export default class MapVisualisation extends Component {
 
     const newSpec = nextProps.visualisation.spec || {};
 
+    if (newSpec.layers.length && !this.state.hasTrackedLayerTypes) {
```
gabemart: Given that it *looks* like newSpec can sometimes be an empty object and `layers` will be undefined, I think we should use lodash `get` here for the length check

Issue #1475, opened 2018-06-19
X axis label appears cut off


nadiagorchakova:
# Context
In this vis, x axis label appears cut off.
https://dark-demo.akvolumen.org/visualisation/5b28abff-6b98-4ab1-8224-ea8a8003951a

<img width="917" alt="screen shot 2018-06-19 at 09 32 20" src="https://user-images.githubusercontent.com/21127166/41583072-b67d5d68-73a3-11e8-9886-b5f2db2ec45f.png">

Looks like for this vis Lars selected 'number of buckets to show: 200' and then sorted it in descending order.  

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

PR #1478, opened 2018-06-19
X axis label appears cut off #1475

finnfiddle:
not very elegant but should do the job


client/src/components/charts/LineChart.jsx
```diff
@@ -370,6 +367,10 @@ export default class LineChart extends Component {
                     stroke={'#1b1a1e'}
                     tickTextFill={'#1b1a1e'}
                     numTicks={xAxisTicks}
+                    tickLabelProps={val => ({
+                      transform: `rotate(45, ${xScale(val)}, 18)`,
```
gabemart: This is causing a new overlap for date types

image

client/src/constants/chart.js
```diff
@@ -13,4 +13,5 @@ export const connectionStyle = {
 
 export default {
   labelFont,
+  MAX_FONT_SIZE: 12,
```
gabemart: On my machine, `MAX_FONT_SIZE` is always `undefined` when imported into the charts

client/src/components/charts/LineChart.jsx
finnfiddle: try now. this is why I made the original change
screen shot 2018-06-19 at 2 17 20 pm

gabemart: For line charts with tick labels that are short numbers, the axis label is now being pushed out of the visible area. Might need to adjust padding or something

gabemart: image

gabemart:
@finnfiddle margin for line chart looks good at medium display size

image

At medium display size (right before the break point where we increase the proportional size of all the viz in the dashboard) there's still some overlap

image

This is on the smallest viz size (which is 1/3 the width of the dashboard) at a viewport width of around 1100px

at large display sizes, the offset between the label and the axis is quite large

image

I don't want to get too hung up on absolute perfection because I'm sure we could spend forever tweaking this, but @finnfiddle how much work would it be to decrease the offset a bit at large render sizes and increase it a big at small render sizes?

@nadiagorchakova @janagombitova thoughts?

gabemart:
@finnfiddle I made a change for bar charts

before:
image

after:
image

gabemart:
@finnfiddle I am OK to merge this now. Could you please test out my bar chart margin change, then merge if you approve?

PR #1480, opened 2018-06-19
Remove aggregation/query `donut` support

tangrammer:


tangrammer:
@gabemart as we chatted previously I've remove the BE legacy code for aggregation `donut`

gabemart:
@tangrammer there was one place where the "donut" agg was invoked so I pushed a small commit to change it to pie (which is what we do with line/area, so it's what I should have done in the first place)

tangrammer:
@gabemart ups, I missed this place, good catch!

gabemart:
👍 let's merge

Issue #1513, opened 2018-06-27
Remove black border around visualisations in dashboards 


janagombitova:
# Context
One of Lumen's strong values is the outputs users create from the data - the visualisations they then share in dashboards. Often our partners only see Lumen's dashboards as all the work behind them is done by Akvo staff as part of our service to the programme. So the dashboards are quite often the face of Lumen.  A dashboard is then the final product, a 'decision-making' tool for partners or a way to share their project outputs and outcomes.

One of our current goals is to improve the visualisations and dashboards, to make them more feature-rich and modern-looking. (https://docs.google.com/document/d/136zC-NpsUNruGhli9aQaRum4NcxBWYuZOb1rcqPuVAc/edit#)   

# Problem or idea
Currently the dashboard consist of text elements and visualisations (meaning also maps) and all of them have a strongly visible black border. 

<img width="922" alt="b8a2e9d4-f841-11e6-838c-22fccff3372b" src="https://user-images.githubusercontent.com/12456965/41960667-a7213918-79f0-11e8-95ee-b8bee03a3026.png">

This border does not help the visual attractiveness of our dashboards. With the UNICEF 9 country presentation, the final posters were created with Lumen visuals but we intentionally removed the black borders to make the posters fresher. 

# Solution or next step
The ides is to remove the black border. We however need to ensure that it is still clear to users on how to move the visuals around and how to make them smaller or larger.

finnfiddle:
@janagombitova I took the liberty of taking this task a bit further and styling the dashboards a bit more than just removing the black borders. Here are some screenshots. What do you think?
![akvo lumen](https://user-images.githubusercontent.com/4628721/42219004-11c68f0c-7ed3-11e8-8c71-b6389ffa4e4c.png)
![akvo lumen 1](https://user-images.githubusercontent.com/4628721/42219005-12141038-7ed3-11e8-8432-bee7684db150.png)

finnfiddle:
reopening because bottom elements are being cutoff: https://dark-lumen.akvotest.org/s/P_X64VRNzBo

Also "The dashboard is also cut off on the right hand side a bit so that the nice new borderless boxes are not fully visible."
![image](https://user-images.githubusercontent.com/4628721/42523813-7f5a3204-8477-11e8-8d29-14b35e7cc374.png)

finnfiddle:
@LarsSjogreen I only see the scrollbars when Im scrolling - is this what youre referring to? 
 
[...]

PR #1522, opened 2018-07-04
Remove black border around visualisations in dashboards #1513

finnfiddle:


gabemart:
@finnfiddle when a visualisation has a very long title, it overflows its container on the dashboard (note the position of the drag handle)

image

gabemart:
I'm also seeing double scroll bar in shared dashboards when height of dashboard is greater than the viewport - I assume because we have a `height: 100%` on the `DashboardViewer`

image

finnfiddle:
@gabemart I made some fixes you can try again

Issue #952, opened 2017-07-12
Download option for visualisations and dashboards


nadiagorchakova:
# Context

In some cases, sharing a visualisation/dashboard via a public link is not enough.
A few partners and partner team members inquired about the 'download' option. 

Use cases:
- sharing a graph/dashboard in paper format with people who don't have internet access
- using a a graph/dashboard in a printable report

As a workaround, it is possible to do a screen grab or to print a web page as a pdf, however, not everyone is happy with it. Apparently it's not easy to do on some devices.

A menu ribbon in visualisations and dashboards has a place holder for download option:
<img width="302" alt="screen shot 2017-07-12 at 16 27 43" src="https://user-images.githubusercontent.com/21127166/28122979-129ffd44-6720-11e7-82d5-a8bf92e68bcd.png">

janagombitova:
Today we discussed with @finnfiddle and @Kiarii this issue. For more notes check the mural board here: 
https://app.mural.co/t/akvo8194/m/akvo8194/1530179115317/37ab92510e4b61e456fbbc02ff9cfe5f63684dc3

# Why
As mentioned in the description above, the reason to allow users to download their visualisation or dashboard is to allow them to use the visualisation or dashboard offline, in their report to a donor, in their project report or presentation, to add to a poster and showcase. 

Currently users use screen grabbing to get an image of their visualisation or dashboard. This approach is not ideal as not every user is familiar with screen grabbing and you cannot always see the values (resolved for pie charts in a separate issue and will create another issue for stacked bar charts). 

# Idea 
We will enable Jerome to **download** his visualisation or dashboard from the editor. He hits download where we give him the option to download a svg or png file. Each download will have a default file size and name (the title of the visualisation). Once he selects the file type the download will start automatically in the browser. Each downloaded file will hold 'Built with Akvo Lumen' in the bottom right corner. 

For downloading dashboards the workflow is the same. When downloading Jerome downloads the entire thing. The difference is in the file types: png or pdf. 

<img width="798" alt="screen shot 2018-07-09 at 16 08 35" src="https://user-images.githubusercontent.com/12456965/42455370-5a7a2330-8392-11e8-9a0b-2051b241cf53.png">

@Kiarii will share details on where the download option should live and how should it behave. 

## Issue of file size
We discussed how to handle the file size. We considered three alternatives: 
- provide a default size that take the width of an A4
- at download create a zip file with three files of different sizes 
- allow Jerome to set the size of the file. 

[...]

PR #1586, opened 2018-08-01
Download option for visualisations and dashboards (connect #952)

finnfiddle:


exporter/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM zenato/puppeteer
```
dlebrero: Can we have a more specific version instead of latest?

dlebrero: I see there are plenty of puppeteer images. See https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=puppeteer&starCount=0. 
The most popular one has tags.

finnfiddle: I first went with the most popular one `alekzonder/puppeteer` but it uses an outdated version of puppeteer. Hence, my choice. But the 2nd one could work because it uses the latest puppeteer - https://hub.docker.com/r/buildkite/puppeteer/. Will try now

exporter/index.js
```diff
@@ -0,0 +1,142 @@
...
+      await page.close();
+    } catch (err) {
+      captureException(err);
+      res.sendStatus(500);
+    }
+  });
+});
+
+function exitHandler(options, err) {
+  if (browser) browser.close();
```
iperdomo: While this code is _concise_ I would prefer always the more verbose `if (expr) { statements }`.

kardan:
Am I wrong in thinking that there is no real check on who and what pages we export? As in there is no check in the exporter that the user is authenticated even if the exporter will be exposed?

finnfiddle:
@kardan yes - there is no auth. I consulted @dlebrero about it a while ago and I cant remember the details of the conversation - but conclusion was that it was fine for now. @dlebrero can you recall?

exporter/index.js
```diff
@@ -0,0 +1,148 @@
...
+
+(async() => {
+  browser = await puppeteer.launch({
+    args: [
+      '--no-sandbox',
+      '--disable-setuid-sandbox',
+    ],
+  });
+  // Create a new incognito browser context.
+  context = await browser.createIncognitoBrowserContext();
```
kardan: Since we send auth token to the exporter don' think we should reuse the browser context, I think we should create a new context per user request. Reading the docs on browser.createIncognitoBrowserContext() "Creates a new incognito browser context. This won't share cookies/cache with other browser contexts." https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#browsercreateincognitobrowsercontext

kardan:
Since this feature uses a new exporter service we need to make sure to deploy that in the k8s environment.

kardan:
Todo:
- [ ] Allow only authenticated requests
- [ ] Update build and deploy scripts to handle the new service

kardan:
Adding a viz to a Dashboard generates the following browser console warning:
````
warning.js?6327:33 Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in DashboardCanvasItem (created by DashboardEditor)
    in div (created by DashboardEditor)
    in Resizable (created by GridItem)
    in DraggableCore (created by GridItem)
    in GridItem (created by ReactGridLayout)
    in div (created by ReactGridLayout)
    in ReactGridLayout (created by DashboardEditor)
    in div (created by DashboardEditor)
    in div (created by DashboardEditor)
    in div (created by DashboardEditor)
    in DashboardEditor (created by Dashboard)
    in div (created by Dashboard)
    in NavigationPrompt (created by withRouter(NavigationPrompt))
    in withRouter(NavigationPrompt) (created by Dashboard)
````

File: backend/src/akvo/lumen/config.clj [hunk 4]
```diff
@@ -10,6 +10,7 @@
   (assert (:lumen-keycloak-client-secret env)
           (error-msg "LUMEN_KEYCLOAK_CLIENT_SECRET"))
   (assert (:lumen-keycloak-url env) (error-msg "LUMEN_KEYCLOAK_URL"))
+  #_(assert (:exporter-api-url env) (error-msg "EXPORTER_API_URL"))
```
tangrammer: maybe we can remove definitely this line

Issue #1571, opened 2018-07-19
Better support Caddisfly data 


janagombitova:
# Context
*Why do we add this issue?*

With this issue we enabled showing Caddisfly data in the Flow data set as a json string: https://github.com/akvo/akvo-lumen/issues/946 This implementation was intentionally basic to give us a starting point to support Caddisfly data. 

With the aim to have a better integration between Caddisfly - Flow - Lumen we need to also simplify this implementation to remove the need for Jerome to have to transform the column to get the values out of the sting in a way he can work with the data.

# Problem or idea
*What is the cause?*

Current implementation show Caddisfly answers as a json string and Jerome can use the derived column to extract the value from the test result and then use it further in a visualisation. This is not the most user friendly approach. Furthermore as the data is coming from our own product the we should be able to show it is a user readable way. 

# Solution or next step
*How could we solve it?*

Flow data exports create separate columns for the Cadidsfly tests. We can follow how Flow exports the Caddisfly data to ensure we handle Caddisfly in a unified way across both products. 

Here is an example of a Flow data export (the data in the sheet is all mock data): 
[Example of Caddisfly columns.xlsx](https://github.com/akvo/akvo-lumen/files/2210059/Example.of.Caddisfly.columns.xlsx)

tangrammer:
Uploaded file to GDrive
https://drive.google.com/file/d/1Pq0ECN9vsrLW026sFbFb054aWRIrGW6u/view?usp=sharing

tangrammer:
(WIP) More on Caddisfly data types here 
https://docs.google.com/spreadsheets/d/1y9XFjrvmmsHZ1z8nSqonAXy5VLao8xgylOE7H9J6GW8/edit#gid=0

tangrammer:
Syncing with @nhternup regarding Caddisfly https://docs.google.com/spreadsheets/d/1y9XFjrvmmsHZ1z8nSqonAXy5VLao8xgylOE7H9J6GW8/edit#gid=0

tangrammer:
Syncing with Joy Ghosh
https://docs.google.com/spreadsheets/d/1uauiqtoD8vkqXWMkCp8bAsS3_vciR-Pu8b70T5Zd_Lc/edit#gid=1307281381

tangrammer:
Syncing with Emmanuel Mulo
[...]

PR #1588, opened 2018-08-03
(DEPRECATED) (Changing approach ...) Transform caddisfly results to dataset columns

tangrammer:
# (DEPRECATED) WIP
**change approach!!, instead of transforming caddisfly results in _"import"_  phase I'm trying to create a new tx type for that, follows up  => https://github.com/akvo/akvo-lumen/tree/issue-1571-txs **
This PR adds work on: [Better support Caddisfly data #1571]
- [ ] do we need to load caddisfly schema tests from aws?? right now being loaded from packaged `resources`
- [ ] do we need to parse caddisfly test values to something different than `strings`?

- [ ] WIP Introduces a new id pattern for _child_ caddisfly new columns `(format "c%s%s" parent-id child-id)`

- [ ] We should set up a new updater version to keep compatibility with older imported datasets
- [ ] Is ok to **only** add this functionality to **flow_v3**? (does flow_v2 need to be extended too?)

# Steps for UI Testing 
- change `backend/dev/resources/dev.edn` :keycloak-url to point `test` environment (ask @tangrammer if you don't know how :) )
- click "+ Dataset" > select "Akvo Flow" > insert "uat1" > select survey "Caddisfly Test" > select sub-survey "Caddisfly Tests" > select "New form"
- that's all! (you can also verify the `update Dataset` functionality)


backend/src/akvo/lumen/import.clj
```diff
@@ -37,14 +37,16 @@
...
-                                                     (contains? columns :key) (assoc :key (boolean key))))
+                                  :columns (mapv (fn [{:keys [title id type key caddisflyResourceUuid]}]
+                                                   (cond->
+                                                       {:type (name type)
+                                                        :title (string/trim title)
+                                                        :columnName (name id)
+                                                        :sort nil
+                                                        :direction nil
+                                                        :hidden false}
+                                                     caddisflyResourceUuid (assoc :caddisflyResourceUuid caddisflyResourceUuid)
```
tangrammer: this change doesn't affect to older datasets updates so doesn't imply adaptations

Issue #1590, opened 2018-08-09
Changing gradient color in map layer theme produces 500 http error


tangrammer:
# Context
Joy reported that problem trying to change color gradient of this viz
https://emw.akvolumen.org/visualisation/5b6c0994-744a-4ec1-8129-e34e8c235ea8

<img width="865" alt="screen shot 2018-08-09 at 12 27 35" src="https://user-images.githubusercontent.com/731829/43893664-0afaa2c0-9bd0-11e8-84ae-0378ec379d50.png">

URL https://emw.akvolumen.org/api/visualisations/maps

payload that generates error:
<img width="910" alt="screen shot 2018-08-09 at 12 18 40" src="https://user-images.githubusercontent.com/731829/43893684-17558d28-9bd0-11e8-9284-a9b6faa6c488.png">
 
this payload (previous to changing gradient color) doesn't generate any error 
<img width="871" alt="screen shot 2018-08-09 at 12 32 52" src="https://user-images.githubusercontent.com/731829/43893799-797b887c-9bd0-11e8-9d07-f2ca37d8feb3.png">

# Problem or idea
still to explore it

# Solution or next step
no idea yet

tangrammer:
error happens in https://github.com/akvo/akvo-lumen/blob/develop/backend/src/akvo/lumen/lib/visualisation/map_config.clj#L160
the color that we are sending is "#7b1fa2"

so calling `(color-to-hue (clojure.string/upper-case  "#7b1fa2"))` 
generates 
```
1. Unhandled java.lang.IllegalArgumentException
   No matching clause: #7B1FA2
            map_config.clj:  116  akvo.lumen.lib.visualisation.map-config/color-to-hue
            map_config.clj:  115  akvo.lumen.lib.visualisation.map-config/color-to-hue
```

tangrammer:
tested against https://dark-lumen.akvotest.org/visualisation/5b74056a-5805-42f9-9e2b-70135fbbd968
and **works fine!**

tangrammer:
and tested against https://lumen.akvotest.org/visualisation/5b7407dc-c587-412a-a7dc-e2a20e0da790
and **works fine!**

PR #1591, opened 2018-08-09
[#1590] Add proper color conversion impl

tangrammer:
based on https://stackoverflow.com/a/7471183/1074389
and https://docs.oracle.com/javase/7/docs/api/java/awt/Color.html#RGBtoHSB(int,%20int,%20int,%20float[])

### DIFF
--- a/backend/src/akvo/lumen/lib/visualisation/map_config.clj
+++ b/backend/src/akvo/lumen/lib/visualisation/map_config.clj
@@ -1,7 +1,9 @@
 (ns akvo.lumen.lib.visualisation.map-config
   (:require [akvo.lumen.postgres.filter :as filter]
             [clojure.string :as str]
-            [hugsql.core :as hugsql]))
+            [hugsql.core :as hugsql])
+  (:import [java.awt Color]))
+
 
 (hugsql/def-db-fns "akvo/lumen/dataset.sql")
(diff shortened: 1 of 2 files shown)


tangrammer:
Tested locally using `docker-compose up --build -d && docker-compose logs -f --tail=10` 
and https://github.com/akvo/akvo-lumen/blob/fix-1590/backend/test/resources/liberia_adm2.csv as data and been able to change gradient theme color 
screen shot 2018-08-13 at 19 43 03

backend/src/akvo/lumen/lib/visualisation/map_config.clj
```diff
@@ -111,12 +113,16 @@
...
-    "#FF0000" "0"
-    "#00FF00" "120"
-    "#0000FF" "240"))
+(defn color-to-hue
+  "Should convert hex string to hue for hsl color."
+  [hex-string]
+  (try
+    (let [c (Color/decode hex-string)
+          hue (->> (Color/RGBtoHSB (.getRed c) (.getGreen c) (.getBlue c) (float-array 3))
+                   (into [])
```
iperdomo: `(first (float-array 3)) => 0.0` why the need to convert to vector with `(into [])` ?

tangrammer: you're right!

Issue #940, opened 2017-07-06
Canvas: select & bring recently added item into view


Kiarii:
# Context
Items are by default added to the bottom of the canvas; if the user is viewing upper parts of the canvas, it means there is no visual cue that the action of adding a textbox or visualization has gone via; they need to scroll to find the item. 
It forces the user into 2 actions, 
- scrolling to find the new box and 
- clicking to edit, resize

# Proposal
- Select recently added item so that it is editable (resize, enter text, moving)
- bring it into view
This improves the workflow a bit; the user can edit the item without having to scroll-to-find and select item

PR #1596, opened 2018-08-15
Canvas: select & bring recently added item into view #940

finnfiddle:

### DIFF
--- a/client/src/components/dashboard/DashboardEditor.jsx
+++ b/client/src/components/dashboard/DashboardEditor.jsx
@@ -2,6 +2,8 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 import ReactGridLayout from 'react-grid-layout';
+import { Element, scroller } from 'react-scroll';
+
 import DashboardVisualisationList from './DashboardVisualisationList';
 import DashboardCanvasItem from './DashboardCanvasItem';
 
@@ -69,6 +71,15 @@ const getFirstBlankRowGroup = (layout, height) => {
   return lastRow + 1;
 };
 
+const handleScrollToItem = (id) => {
+  scroller.scrollTo(`DashboardCanvasItem__${id}`, {
+    duration: 500,
+    smooth: true,
+    containerId: 'DashboardEditorCanvasContainer',
+    offset: -60,
+  });
+};
+
 export default class DashboardEditor extends Component {
 
   constructor() {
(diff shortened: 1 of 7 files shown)


client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -69,6 +71,15 @@ const getFirstBlankRowGroup = (layout, height) => {
   return lastRow + 1;
 };
 
+const handleScrollToItem = (id) => {
+  scroller.scrollTo(`DashboardCanvasItem__${id}`, {
+    duration: 500,
+    smooth: true,
+    containerId: 'DashboardEditorCanvasContainer',
```
tangrammer: could we extract the containerId string value to a var so it could be reused in other parts of this file?

client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -78,6 +89,7 @@ export default class DashboardEditor extends Component {
       propLayout: [],
       saveError: false,
     };
+    this.canvasEls = {};
```
tangrammer: what does `Els` stand for?

client/src/components/dashboard/DashboardEditor.jsx
```diff
@@ -132,10 +144,13 @@ export default class DashboardEditor extends Component {
   handleEntityToggle(item, itemType) {
     const newEntities = Object.assign({}, this.props.dashboard.entities);
     const newLayout = this.props.dashboard.layout.slice(0);
+    const itemExists = this.props.dashboard.entities[item.id];
```
tangrammer: what do you think if we use `entity` instead `itemExists` :) ?

client/src/components/dashboard/DashboardEditor.jsx
finnfiddle: short for elements

client/src/components/dashboard/DashboardEditor.jsx
finnfiddle: done

client/src/components/dashboard/DashboardEditor.jsx
finnfiddle: changed it to elements

client/src/components/dashboard/DashboardEditor.jsx
finnfiddle: done

PR #1614, opened 2018-08-24
Add extract multiple column transformation option

tangrammer:
### Add extract multiple column transformation option
Relates to #1571 

##  Data Testing 
- change `backend/dev/resources/dev.edn` :keycloak-url to point `test` environment (ask @tangrammer if you don't know how :) )
- click "+ Dataset" > select "Akvo Flow" > insert "uat1" > select survey "Caddisfly Test" > select sub-survey "Caddisfly Tests" > select "New form"
- finish and go into dataset 

##  click on "extract multiple column" transformation sidebar
screen shot 2018-08-24 at 12 07 32

- [ ] **Update `fr` translations**


client/src/components/dataset/sidebars/ExtractMultiple.jsx
```diff
@@ -0,0 +1,362 @@
...
+            <input
+              value={ui.name}
+              placeholder={api.name} // TODO: how to i18n this value?
+              type="text"
+              className="titleTextInput"
+              onChange={this.onColumnName}
+              data-test-id={`column-title-${api.id}`}
+            />
+          ) : (
+             undefined
```
finnfiddle: should return null instead I think

client/src/components/dataset/sidebars/ExtractMultiple.jsx
```diff
@@ -0,0 +1,362 @@
...
+  }
+  render() {
+    const { api, ui } = this.props;
+    return (
+      <div className="inputs">
+        <div className="inputGroup">
+          <div>
+            <FormattedMessage id="extract_column_question" />
+            {ui.extract ? (
+               undefined
```
finnfiddle: should return null instead I think

client/src/components/dataset/sidebars/ExtractMultiple.jsx
```diff
@@ -0,0 +1,362 @@
+import * as _ from 'lodash';
```
finnfiddle: you shouldnt import the whole of lodash because of its size. instead import just the functions you need eg:
`import { merge } from 'lodash';`

client/src/components/dataset/sidebars/ExtractMultiple.jsx
```diff
@@ -0,0 +1,362 @@
...
+            onExtractColumn={this.onExtractColumn}
+          />
+        </div>
+
+        <SidebarControls
+          positiveButtonText={<FormattedMessage id="extract" />}
+          onApply={
+            this.isValidTransformation()
+              ? () => onApply(this.state.transformation)
+              : null
```
finnfiddle: shouldnt this return a noop function?

client/src/components/dataset/sidebars/ExtractMultiple.jsx
```diff
@@ -0,0 +1,362 @@
...
+    );
+  }
+}
+
+ExtractMultiple.propTypes = {
+  onClose: PropTypes.func.isRequired,
+  onApply: PropTypes.func.isRequired,
+  columns: PropTypes.object.isRequired,
+};
+//    const { transformation } = this.state;
```
finnfiddle: you can remove commented out code

client/src/components/dataset/sidebars/ExtractMultiple.jsx
finnfiddle: eg: `() => {}`

finnfiddle:
seems to be working fine for me. needs thorough testing though I think to cover all edge cases if there are any. however in terms the ui/ux it is a little difficult to see the values for the json data in the cells. I see in the tooltip the json appears if you hover but its not formatted well. I would suggest having some sort of custom popover or something that shows you nicely formatted code when you hover/click. I can do it but not today I think. It could also be done at a later stage in the interest of time

tangrammer:
@finnfiddle  your comment https://github.com/akvo/akvo-lumen/pull/1614#issuecomment-416177882 about improving json representation although seems to me a good improvement IMO is not directly related with the PR or issue that we’re currently working on 😬

Issue #1281, opened 2018-02-16
Dataset update logic very fragile


nadiagorchakova:
# Context

Update of a dataset in Lumen will fail after any small change made to a Flow survey. Even changing the wording of a survey question will fail the update.
That is because derived column logic is built around typing the actual column name (survey question in Flow) for the transformation to happen. At least that's what I remember Jonas explained to me.

# Problem or idea
Even though it happens rare that Flow users change the entire survey structure (according to Jana), small corrections to survey questions happen more often. 
However, making a correction to a survey question in Flow would break dataset update in Lumen without users knowing what went wrong.

# Solution or next step
How can we make the update logic less fragile?
@kardan

kardan:
Have been investigating this a bit and I think we can do the updates way better. But - it requires us to change how derived columns work. Today we reference the column title, (what's seen in the UI) we need to change that to have the actual transformation to reference the columnName (Flow question-id) for us. Otherwise, we'll break a lot of things. We have talk about this before and we should make it a priority to rework how derived columns operate.

nadiagorchakova:
Jana enlightened Kiarii and me that there are actually two question IDs in Flow - internal numerical one and user-visible 'variable name'. You probably mean internal numerical one?

Issue #1585, opened 2018-07-25
Support full JS in derived columns 


janagombitova:
# Context
*Why do we add this issue?*

With changes to this issue: https://github.com/akvo/akvo-lumen/issues/1296 we can now support full JS in transformations. This change was introduced to enhance Lumen's security. The implementation has no change to the users, Jerome continues to put in expressions as before, we automatically add `return` to it to execute. In the UI the validation allows for expressions and not full JS code. 

# Problem or idea
*What is the cause?*

The current implementation that Jerome sees does not give him the option to fully utilise the possibilities of JS. We can create a way for Jerome to decide if we wants to use an expression in his transformation or write the full code. 

# Solution or next step
*How could we solve it?*

Ideally we want to simplify how transformations are handled in Lumen, moving away from writing expressions or code, to having a interface that can guide anyone in the transformations you want to do. Supporting more advance JS transformations is heading in the opposite direction of this goal. On the other hand, if there is a current gap that Lumen users cannot do, we can fill it. 

The next step is: 
- to research with current Lumen users, what expressions do they use most in Derived columns 
- to research with current Lumen users, what is it that they want to do, but cannot currently and why do they want to do it (what value)
- evaluate if this change would be beneficial to implement or not

nadiagorchakova:
Hi there, I did a quick survey to poll a few Akvonuats about the use of derived column and what is it that they cannot do right now.

Who took part in the survey: Joy, Geert, Veerle, Molo, Guneet.

You can find full responses to the survey here: https://docs.google.com/forms/d/1fkcXmG5nI-8c7UiuligyuFoSz_zWiHdOHd0eT4v9Vyc/edit#responses

The overview:

**What expressions do you use most in derived columns?**

- mostly if statements , array splits and concatenations
- Scoring, merging, splitting, date manipulations, (date) calculations. 
- creating groups, e.g. row ["column name"] < "number" ? “text" : (row["column name"] < number ? “text" : etc.
- parseFloat(), split(), indexOf()
- nested switch statements, split, addition

**What is it that you want to do but cannot currently?**

- Because lumen currently allows access only to row data and not column data, it limits the use of the derived columns. Ex: I want to mark all points in red where the time taken to conduct the survey is 20% above the average survey  time. Lumen does not allow computation of a column dynamically. Hence all visualisations where you are trying to compute dataset wide values dynamically cannot be executed
- a slightly complex mathematical calculation (like roots, logs), including logical statements

[...]

PR #1633, opened 2018-09-03
Support column titles changes in dataset update functionality [#1281]

tangrammer:
These changes will extend the update dataset functionality to those that did titles changes over current lumen dataset imported version

The approach chosen is localized on the`akvo.lumen.update` and `akvo.lumen.transformation.derive` namepaces 

Basically we analyse the js code looking for column-name references, then having the column-name we keep a reference to column-id. 
When we update the dataset we replace older column-names with updated column-names using common column-id 
`older-imported-dataset.column-name`=>`column-id`=>`new-updated-imported-dataset.column-name`

# tangential bounds with transformation `undo` op
the problem appears thus columns metadata (as a result of import operation) is stored in each dataset_version row. This PR updates previous dataset_version `transformations` and `columns` data rows with each update,  setting the initial version=X, [being X the version that has `[]` as a `transformations` value](https://github.com/akvo/akvo-lumen/pull/1633/files#diff-a3544a5d3ede3260172b252c0bc06273R30) . (Theoretically/Logically we shouldn't need more than that)

# testing locally the new changes
you need to do intensive testing to these new changes as far as they affect to several "trasnformation" parts of lumen... so is not realistic to test it with just flow manual changes ....
so, there is a different branch https://github.com/akvo/akvo-lumen/tree/dev-endpoint-to-download-local-files that open a “door” in development to download files from backend “resources” ... example=> http://t1.lumen.local:3000/local-server/dos.csv will response the contents of `akvo-lumen/backend/test/resources/dos.csv` ... very useful  to provide online dataset (LINK option) from local computer  so you could easily change the contents and try the `update`  dataset option again and again

Relates to #1281 
Relates to #1652

### DIFF
--- a/backend/src/akvo/lumen/dataset.clj
+++ b/backend/src/akvo/lumen/dataset.clj
@@ -3,7 +3,7 @@
   (:require [akvo.lumen.endpoint.job-execution :as job-execution]
             [akvo.lumen.import :as import]
             [akvo.lumen.lib :as lib]
-            [akvo.lumen.transformation.derive :as derive]
+            [akvo.lumen.transformation.engine :as engine]
             [akvo.lumen.update :as update]
             [clojure.java.jdbc :as jdbc]
             [clojure.set :refer (rename-keys)]
@@ -71,15 +71,7 @@
                                  [(select-data-sql (:table-name dataset) columns)]
                                  {:as-arrays? true}))
 
-          transformations (reduce (fn [transformations operation]
-                                    (if (= "core/derive" (get operation "op"))
-                                      (conj transformations
-                                            (assoc-in operation
-                                                      ["computed" "code"]
-                                                      (derive/construct-code (:columns dataset) operation)))
-                                      (conj transformations operation)))
-                                  '()
-                                  (:transformations dataset))]
+          transformations (engine/transformations* (:columns dataset) (:transformations dataset))]
       (lib/ok
        (-> dataset
            (select-keys [:created :id :modified :status :title :updated :author :source])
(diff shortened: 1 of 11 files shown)


tangrammer:
@kardan perhaps we'll have to rethink these changes if we are going to add support to full js functionality
... in relation to dynamic js column names 
Ex: 
```
var o ={row1:"uno", row2: "dos"}
for(var i=0; i<3 ; i++ ){console.log(o["row"+i]);}

//=> uno
//=> dos

```

tangrammer:
in relation to [Support full JS in derived columns](https://github.com/akvo/akvo-lumen/issues/1585)
Maybe, we can approach the "full js derive functionality" using some sort of checkbox option that shows a "dataset js view " with columnNames instead of Column Titles... (and internally clojure code will not do any columnTitle-to-columnName conversion ...)
@kardan @janagombitova

kardan:
Yes it's a good question how well this will work with full JS. I find it difficult to think about all the work we have put into the JS feature to have it working. (updates, run time...) while users are complaining that it's too difficult to use. Sometimes I almost wonder if we should not try to make transformations for the most common problems or create some mini DSL that we interpret and execute in Clojure land... Sorry just coming back, looking at it with fresh eyes...

tangrammer:
@kardan @janagombitova maybe we should schedule some time to talk about next steps based in targeting user value 🤔

tangrammer:
Found a problem with `undo` functionality

tangrammer:
just realising that not only `derive` transformation should be affected by this fragile-update support feature ... all transformations that refer the `column-title` should be changed too  (eg. renaming column)... the UI transformation log  should be adapted consequently

tangrammer:
so far `update` logic will update each row of dataset_version, starting in the greatest version with `transformations = '[]'` . .... thus we ensure that `redo` operations will be synced to last dataset `update` op



current commit seems that works fine with redo op :) 

only rest to ensure all trasnformation ops based in column `title` also update the transformation log that end users see in right sidebar UI

tangrammer:
tested locally using flow uat1 instance and worked fine 🎉 !

Issue #1638, opened 2018-09-05
Fix db inconsistency in dataset table source column 


tangrammer:
# Context
https://github.com/akvo/akvo-lumen/issues/778#issuecomment-418625073

# Problem or idea
Older dataset prior this [merge](https://github.com/akvo/akvo-lumen/commit/39b2b6406e8d3c077766854f231abe57ee3f3397) weren't updated 
# Solution or next step
How could we solve it?
TODO:

PR #1644, opened 2018-09-06
Drop dataset source column [#1638]

tangrammer:
Remove tenant db duplication. 
`dataset.source` was/is already being persisted in `data_source.spec`


kardan:
Should we not first do a release where we make the app use the data_source.spec value, deploy that then drop the column once both dark and live is running the new code?

tangrammer:
@kardan splitting this PR in 2, so migration was the second one to be merged

tangrammer:
(I wanted to check travis results with all changes in place) :)

kardan:
Perfect!

tangrammer:
@kardan This PR should be merged in develop/master after => https://github.com/akvo/akvo-lumen/pull/1644

Issue #1562, opened 2018-07-13
Disable update for datasets coming from Data files 


janagombitova:
# Context
Jerome can update a dataset if it comes from Flow or a url. This functionality does not work when the dataset is based on a data file. 

# Problem or idea
In the UI, Jerome can currently see that is data set is based on a data file. In the side menu, he also sees the option to update, but this is not possible to do. 
<img width="1032" alt="screen shot 2018-07-13 at 15 11 12" src="https://user-images.githubusercontent.com/12456965/42697659-0a6102a6-86bc-11e8-8131-5f415381bfa1.png">

# Solution or next step
We should either fully hide the option to Update for data sets based on data files, or disable this option in the menu.

tangrammer:
tested against https://dark-lumen.akvotest.org/library/collections/9fd07166-e83a-4ac9-a094-9964cd17b351 
<img width="635" alt="screen shot 2018-08-30 at 19 02 08" src="https://user-images.githubusercontent.com/731829/44867112-43089b00-ac87-11e8-999b-3159ca6b5e2e.png">

janagombitova:
@tangrammer it seems we did not count with one case around this implementation. In not all cases we know where the data set came from, as we did not store that information since the beginning of time. So we have data sets on Lumen, that do not show their 'origin' but for these the `update` button does not show either. But it should as they maybe came from Flow or a link, we do not know

PR #1648, opened 2018-09-07
Disable *again* update for datasets coming from Data files [#1562]

tangrammer:
Relates also to #1638


tangrammer:
this code was already reviewed so merging now ...

PR #1671, opened 2018-09-20
Improve derive/parse-row-object-references impl and test coverage

tangrammer:

### DIFF
--- a/backend/project.clj
+++ b/backend/project.clj
@@ -72,8 +72,9 @@
    :test          [:project/test :profiles/test]
    :uberjar       {:aot :all}
    :profiles/dev  {}
-   :profiles/test {}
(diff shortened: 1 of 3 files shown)


tangrammer:
@iperdomo could you take a look also to the reg expression https://github.com/akvo/akvo-lumen/pull/1671/files#diff-447779b93863a0b103d1697460471585R35 ?

iperdomo:
@tangrammer i guess this code gets executed *after* we have parsed the js expression and we know that is valid JS code? I mean, we assume that the code is valid, e.g. `row['foo]` or ` row['foo'']` returns empty sequence `()`.

kardan:
Meh, just saw that @iperdomo made a comment :-(

Issue #1224, opened 2018-01-15
Update to new version of the Duct framework


kardan:
# Context
We are using an old version of Duct.

# Problem or idea
We should keep up with development.

# Solution or next step
Migrate to use new Duct, that uses Integrant instead of Component.

kardan:
@tangrammer think this is something we should explore? Think we should start with a research issue. Moving to use Integrant instead of Component is fairly involved.

tangrammer:
I like this issue !

PR #1674, opened 2018-09-25
Update duct version (connect #1224)

tangrammer:


backend/dev/src/dev.clj
```diff
@@ -2,32 +2,53 @@
...
-  (load-system (keep io/resource
-                     ["akvo/lumen/system.edn" "dev.edn" "local.edn"])))
+(defn read-config []
+  (duct/read-config (io/resource "dev.edn")))
+
+(derive :akvo.lumen.component.emailer/dev-emailer :akvo.lumen.component.emailer/emailer)
+(derive :akvo.lumen.component.caddisfly/local :akvo.lumen.component.caddisfly/caddisfly)
+(derive :akvo.lumen.component.error-tracker/local :akvo.lumen.component.error-tracker/error-tracker)
+
+(defn dissoc-prod-components [c]
```
iperdomo: Not sure why are directly referencing the production one and then removing it at dev time, instead we should reference the generic *parent* one?

https://github.com/weavejester/integrant#derived-keywords

tangrammer: After talking with @iperdomo we agree on ignoring this comment 
(Expanded bits are that integrant needs the implementation (child) keyword in dev.edn or config.edn thus integrant will search for the init-key defmethod for the generic (parent) keyword component to be initiated)

tangrammer:
@kardan aligned with `develop`

tangrammer:
@kardan rebased with develop including export service now :)

kardan:
I could not get local configs (from local.edn) to work properly. I looked at the default Duct template and it looked like an empty local.clj file was created so did that locally but that didn't pick up local configs. Need to sign out for the day, will continue to look at it tomorrow but @tangrammer did you try this?

tangrammer:
@kardan I didn't commit local.edn nor added to settings thus i didn't know how not to commit the file to the git repo so i didn't want to track its changes. The problem is more related to how put this empty file in the repo but doesn't track it in git 🤔 ... maybe you know better how to do it or if it's worthing 


anyway there is no problem to add it and get it working

Issue #1689, opened 2018-10-04
Don't allow deletion of datasets that is used in a merge transformation


kardan:
# Context
For datasets used in viz there is a warning when trying to delete. For dataset used in merge transformations there is not.

# Problem or idea
The problem is that deleting a dataset gives no warnings but can cause a lot of work useless.

# Solution or next step
We should either warn of not allow dependent datasets to be deleted.

---------------------

kardan:
Possible options:
A: Don't allow deletion of datasets that are used in transformations
B: Warn that datasets have dependents but allow deletion of only the single dataset
C: Warn that datasets have dependents and delete everything that is dependent (other datasets & viz)
D: Do what we do today. Allow deletion and fails. But make sure the user can remove failing transforms to not get stuck
E: ?

tangrammer:
@kardan after playing a bit with the transformation-merge-datasets, I think we also need to take care of  the source columns dataset consistency, thus if  we e.g remove a column that participates in merge operation then "updating" dataset should fail 🤔

tangrammer:
this code is WIP
`select DISTINCT ON (dataset_id) dataset_id, id, version, transformations from dataset_version order by dataset_id, version desc`

kardan:
I'm not sure what I think. In one way it's odd if there is two users involved. Can the merger block the original author at the point? @janagombitova

janagombitova:
After our talk last week about this issue we agreed to also **not allow to delete a dataset that is used in a merge**? Should we create a separate issue for this change?

tangrammer:
@janagombitova not necessary another issue, I just started ensuring update operations to protect the database from inconsistencies .... working now in "not allow to delete a dataset that is used in a merge"

janagombitova:
And then the next step would be to also not allow to delete the column from the data set used in merge, right? @tangrammer

tangrammer:
yep!

PR #1693, opened 2018-10-09
Avoid database inconsistencies on updating problematic datasets (Connect #1689 )

tangrammer:
Problematic datasets are those that have merge transformations that use datasets columns that don't exist anymore, maybe the column was removed, maybe the dataset was removed

- [x] check new queries work properly with more test data
- [X] check dependent datasets
- [x] check dependent dataset columns

### DIFF
--- a/backend/src/akvo/lumen/dataset.clj
+++ b/backend/src/akvo/lumen/dataset.clj
@@ -8,7 +8,7 @@
             [clojure.tools.logging :as log]
             [clojure.java.jdbc :as jdbc]
             [clojure.string :as str]
-            [clojure.set :refer (rename-keys) :as set]
+            [clojure.set :refer (rename-keys)]
             [hugsql.core :as hugsql]))
 
 (hugsql/def-db-fns "akvo/lumen/dataset.sql")
--- a/backend/src/akvo/lumen/dataset.clj
+++ b/backend/src/akvo/lumen/dataset.clj
@@ -3,7 +3,7 @@
   (:require [akvo.lumen.endpoint.job-execution :as job-execution]
             [akvo.lumen.import :as import]
             [akvo.lumen.lib :as lib]
-            [akvo.lumen.transformation.merge-datasets :as md]
+            [akvo.lumen.transformation.merge-datasets :as transformation.merge-datasets]
             [akvo.lumen.update :as update]
             [clojure.tools.logging :as log]
             [clojure.java.jdbc :as jdbc]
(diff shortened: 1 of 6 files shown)


tangrammer:
@iperdomo perhaps you want to take a look to these SQL additions, I've tested locally the "update-dataset" UI functionality using https://github.com/akvo/akvo-lumen/tree/dev-endpoint-to-download-local-files branch 
also I've executed the new sqls against https://customer.elephantsql.com/instance dev database comparing manually the results  and they seem consistent :) 

Also I've tried the `EXPLAIN` clause with them but don't know very well how to intepret the results :!

example 

```
explain select DISTINCT ON (dataset_id) dataset_id, id, version, transformations, columns
FROM dataset_version
WHERE dataset_id IN ('1') 
order by dataset_id, version desc;
```

```
  | QUERY PLAN                                                                                        |
  |---------------------------------------------------------------------------------------------------|
  | Unique  (cost=9.51..9.52 rows=2 width=130)                                                        |
  | ->  Sort  (cost=9.51..9.52 rows=2 width=130)                                                      |
  | Sort Key: version DESC                                                                            |
  | ->  Bitmap Heap Scan on dataset_version  (cost=4.16..9.50 rows=2 width=130)                       |
  | Recheck Cond: (dataset_id = '1'::text)                                                            |
  | ->  Bitmap Index Scan on dataset_version_dataset_id_version_key  (cost=0.00..4.16 rows=2 width=0) |
  | Index Cond: (dataset_id = '1'::text)                                                              |

```

Maybe you can find more in these results :)

tangrammer:
@kardan here you are another PR to review! :)

backend/src/akvo/lumen/dataset.clj
```diff
@@ -3,10 +3,12 @@
   (:require [akvo.lumen.endpoint.job-execution :as job-execution]
             [akvo.lumen.import :as import]
             [akvo.lumen.lib :as lib]
+            [akvo.lumen.transformation.merge-datasets :as transformation.merge-datasets]
             [akvo.lumen.update :as update]
+            [clojure.tools.logging :as log]
             [clojure.java.jdbc :as jdbc]
             [clojure.string :as str]
-            [clojure.set :refer (rename-keys)]
+            [clojure.set :refer (rename-keys) :as set]
```
kardan: ":as set" is not needed, we're not referencing set anywhere

kardan: ":as set" is not needed, we're not referencing set anywhere

kardan: "set" is never used.

tangrammer: Updated!

Issue #1055, opened 2017-09-07
The use of count aggregation in bar chart


nadiagorchakova:
# Context
Yesterday we had a chat with @gabemart about it and the conclusion was that count aggregation in bar charts is a bit problematic.

# Problem or idea
The count aggregation counts values of the column from the X axis, which makes the Y axis scale redundant.

Bar chart example: https://lumen.akvotest.org/visualisation/59b00ddf-3abd-4057-9926-dc5a7cbdc1be

As you can see, with aggregation set to count, nothing changes as you switch between different columns in the Y axis

jonase:
I've also noticed the same issue (see https://github.com/akvo/akvo-lumen/issues/1016#issuecomment-325651526)

I think `count` aggregation is the most common one and should probably be the default

finnfiddle:
@nadiagorchakova this seems like its working as expected. what should I do?

nadiagorchakova:
Hi @finnfiddle, I believe the issue hasn't been resolved yet.
The issue is that when count aggregation is used, the y axis becomes redundant as Lumen counts on X axis 'buckets'. I tried to show it in this gif

![ezgif com-gif-maker 9](https://user-images.githubusercontent.com/21127166/42329425-f4911764-8070-11e8-8012-b1e57d154aa6.gif)

I discussed it with @janagombitova and she will take it forward.

The

finnfiddle:
@nadiagorchakova @janagombitova I realize the value isnt changing when you change the y axis metric column because its just showing a count of the buckets. However, I wouldnt say the y-axis is redundant because it is still showing information to the reader. Im confused what youd like me to do. If I remove it then they wont be able to read the count unless they hover

nadiagorchakova:
our initial idea was to automatically remove y axis label when count is selected. In the gif above, I removed it already manually. Jana had also other ideas how to improve the workflow.

janagombitova:
@finnfiddle I spoke with two of our strong users to ask them about this issue and how they create bar charts. 

# User research 

"I never knew that only seeing the counts for the categories was possible. I always thought you need to select both x and y axis, so for cases where we just wanted to show number of submissions by submitter, for example, we used pie - what is not good if you have more than 2 categories. But it is good to know this is possible. In this case, it makes sense to start with the x axis for bar charts Because you get to see the initial count of the values in the variable before starting various aggregations and adding y-axis metric column."

[...]

PR #1726, opened 2018-10-25
Use "count" metricAggregation  (Connect #1055)

tangrammer:
Use "count" metricAggregation when no metricColumnY but there is bucketColumn [#1055]

Extracted from slack ...
```
in `/backend/src/akvo/lumen/lib/aggregation/bar.clj` if we do not send `metricColumnY` in the query parameters but we do send bucketColumn then we need to use count aggregation for the y value. make sense?
```

### DIFF
--- a/backend/src/akvo/lumen/lib/aggregation/bar.clj
+++ b/backend/src/akvo/lumen/lib/aggregation/bar.clj
@@ -28,7 +28,7 @@
         column-y-name (get column-y "columnName")
         column-y-title (get column-y "title")
 
-        aggregation-method (if (and (not column-y) column-subbucket) "count" (get query "metricAggregation"))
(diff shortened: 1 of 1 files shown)


tangrammer:
@finnfiddle could you try this functionality against https://github.com/akvo/akvo-lumen/tree/The-use-of-count-aggregation-in-bar-chart-%231055 ?

finnfiddle:
@tangrammer I merged my branch into this and it doesnt work. See response in image 
screen shot 2018-10-25 at 4 59 13 pm

tangrammer:
@finnfiddle 
You specify that "Use count metricAggregation when no metricColumnY but there is bucketColumn "
But in your image there isn't bucket column selected ...

PR #1733, opened 2018-10-26
Add some monitoring

dlebrero:
Using Prometheus:

* Added monitoring stats from HikariCP (https://github.com/brettwooldridge/HikariCP/wiki/Dropwizard-Metrics)
* Added very basic HTTP metrics.
* JVM (memory/threads/gc)

Also fixed a connection pool leak.

About the HTTP metrics, there is no way to get the full compojure matched path, so instead of
having thousands of urls (due to the REST urls), prefer to have just one
stat, so at least we can see the traffic and error rates.

Compojure has a wrap-routes function to do this, but it doesnt work as we
need when using compojure/context. When using context, we can just capture
the latest bit of the matched api. For example:

(context "/visualization/:id"
  (GET "/meta"...

We are just able to get the "/meta" part, when what we actually want is
   "/visualization/:id/meta".

### DIFF
--- a/backend/dev/resources/dev.edn
+++ b/backend/dev/resources/dev.edn
@@ -9,7 +9,7 @@
   {:middleware
    {:functions {:stacktrace #ig/ref :akvo.lumen.middleware.ring.stacktrace/wrap-stacktrace}
     :applied   ^:replace [:not-found :json-response :ring-defaults :json-body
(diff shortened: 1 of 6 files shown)


kardan:
@dlebrero let's catch up on where to take this. I see that the PR is labeled as "In progress" but unless I'm missing something I guess we could merge and deploy? I'm very interested into seeing what insights this brings us.. 🥇

PR #1741, opened 2018-10-31
Issue/1710 speed up builds

iperdomo:
This is the first part of more work. We want to _cache_ all the dev docker images. Those images is now hosted in https://github.com/akvo/akvo-dockerfiles

A second iteration will check the time spent in integration tests.

## How to test?

```bash
docker system prune -a -f && cd /tmp && git clone --depth=50 --branch=issue/1710-speed-up-builds https://github.com/akvo/akvo-lumen.git && cd akvo-lumen && docker-compose up --build
```


ci/e2e-test.sh
```diff
@@ -1,20 +1,31 @@
...
 
-if [ "${DOCKER_COMPOSE_PROJECT}" == "akvolumen" ]; then
-    DOCKER_COMPOSE_ARGS=""
+DOCKER_COMPOSE_PROJECT="${1:-akvolumen}"
+LUMEN_URL="${2:-http://t1.lumen.local:3030/}"
+LUMEN_USER="${3:-jerome}"
+LUMEN_PASSWORD="${4:-password}"
+CYPRESS_RECORD_KEY="${CYPRESS_RECORD_KEY:-}"
+
+if [[ "${DOCKER_COMPOSE_PROJECT}" == "akvolumen" ]]; then
```
dlebrero: I don't get why the duplication is better.

iperdomo: It's not better, i was trying to fix the e2e tests (lots of changes, related and unrelated) and this seems to be executing something with _wrong_ arguments. I can revert it back to the original version

dlebrero: Not too fuss about it. Just wanted to know if there was something wrong with the original approach, so that I did not made the same mistake again.

I understand that the original approach was ok and I am also ok with the current approach

iperdomo:
@kardan who needs to hit the _Merge pull request_ button?

kardan:
@iperdomo trying to figure out why emails are not sent from invites. I'm thinking that we should fix that and then merge this. Hope that makes sense.

Issue #1710, opened 2018-10-15
Speed up builds


iperdomo:
# Context
The build time in CI has increased dramatically. We're adding more components and we're building everything from scratch on every build

# Problem or idea
We have Docker images that do not change too often and we can build them separately and just use them as part of the docker-compose system.

# Solution or next step

- Move development Docker images to https://github.com/akvo/akvo-dockerfiles
- Use the built images in the docker-compose system

kardan:
Should we close https://github.com/akvo/akvo-lumen/issues/1270 ?

iperdomo:
@kardan yes

PR #1758, opened 2018-11-07
[#1710] Fix ubound variable

iperdomo:
We use WAIT_FOR_DB as signal for develop and testing in CI. When
running a bash script with `set -o nounset` we need to provide defaults
to potentially uninitialized variables. We use the "${VAR:=default}"
as way to default a value.

> ${parameter:=word}
>
>    If parameter is unset or null, the expansion of word is assigned
>    to parameter. The value of parameter is then
>    substituted. Positional parameters and special parameters may not
>    be assigned to in this way.

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

### DIFF
--- a/backend/maybe-import-and-java-jar.sh
+++ b/backend/maybe-import-and-java-jar.sh
@@ -3,7 +3,7 @@
 set -o errexit
 set -o nounset
 
-if [[ "${WAIT_FOR_DB}" = "true" ]]; then
+if [[ "${WAIT_FOR_DB:=false}" = "true" ]]; then
   /app/wait-for-dependencies.sh
 fi
 


tangrammer:
@iperdomo 
I'm trying to get into the details of this `sh` logic and this is my attempt ... 

> when executing `backend/maybe-import-and-java-jar.sh` it `/app/wait-for-dependencies.sh` if `WAIT_FOR_DB`. after that `java -jar /app/akvo-lumen.jar`

here my questions are... 

why this script `backend/maybe-import-and-java-jar.sh` says `import` when we don't apparently import anything?

should we rename `WAIT_FOR_DB` to `WAIT_FOR_DEPS` ??

and as aside question? in shell scripting is it advised/necessary to use uppercase?

anyway, regarding the PR change seems very fine :)

Issue #1737, opened 2018-10-30
Double scroll bars when selecting visualisation type 


janagombitova:
# Problem or idea
*What is the cause?* 

When selecting a visualisation type you get double scroll bars, vertical and horizontal (not the intended blue scrollbar, which you also get). 

# Solution or next step
How could we solve it?

![screen-shot-20181030-at-09](https://user-images.githubusercontent.com/12456965/47708164-c74ab080-dc2d-11e8-9178-1f44727a6fd2.png)

finnfiddle:
I can remove the vertical scrollbar. However, reimplementing the native scrollbars is not a trivial task to get them working nicely across all browsers etc. So instead of replacing the native ones I added the blue indicator. This is so that on macOS - when the scrollbars hide themselves automatically - then you know the scroll position of the container by looking at the blue indicator. If you would like it differently let me know

loicsans:
@finnfiddle I actually did reimplement the native scrollbar and removed the vertical scrollbar last week, there's a pull request assigned to you

finnfiddle:
So when implementing this feature 2 requirements arose:

1. the user (Jerome) should be able to scroll sideways without a trackpad or mousewheel
2. Jerome should be able to see at all times the scroll position

I dont have any answers but the implementation was the best I could come up with

loicsans:
@finnfiddle I don't see the problem, with my implementation not meeting the requirements

loicsans:
![nov-27-2018 12-48-08](https://user-images.githubusercontent.com/1717219/49079347-fcf0b280-f248-11e8-8279-c75355199aea.gif)

finnfiddle:
@loicsans try on firefox. Firstly, there is no indication of the scroll position. Also the native scrollbars dont appear if i hover so I have no way of scrolling without a trackpad or mousewheel.

loicsans:
The vertical scrollbar is addressed, although, we decided that the issue is the choice of layout itself and won't accommodate the addition of more visualisations going forward. 
Therefore fixing the scrollbar isn't worth the effort for now, I'll reopen a design issue concerning the visualisation selector.

PR #1789, opened 2018-11-20
Double scrollbar (Connect #1737)

loicsans:


loicsans:
This should look like this:
screenshot 2018-11-21 at 10 20 17

client/src/components/common/ConfigMenu/ConfigMenuSection.scss
```diff
@@ -5,7 +5,7 @@
   border: 1px solid $lightGrey;
   @include border-radius(6px);
   margin-bottom: 15px;
-  padding: 15px 15px 0 15px;
```
finnfiddle: @loicsans can I ask whats the purpose of this because its breaking the other sections? I can try implement it just for the vis type section if thats the goal

finnfiddle:
closing because of new branch/PR made here https://github.com/akvo/akvo-lumen/pull/1819

client/src/components/common/ScrollBar.scss
```diff
@@ -8,26 +8,52 @@
...
   &-content {
     width: 100%;
-    padding-bottom: 4px;
+    padding-bottom: 15px;
     margin-bottom: 4px;
     
     &-inner {
       overflow: auto;
+      padding-bottom: 15px;
+      &::-webkit-scrollbar {
```
finnfiddle: these webkit-specific selectors will only work on chrome, safari - not firefox and others. In the docs it says: ```Non-standard
This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.```

client/src/components/common/ScrollBar.scss
```diff
@@ -8,26 +8,52 @@
...
     left: 0;
-    bottom: 0;
+    bottom: 6px;         
   }
 
   &-handle {
     position: absolute;
     height: 2px;
-    background: $anchorLinkLight;
+    background: #7C7C7C;
```
finnfiddle: we should store colors in variables in the `styles/modules/colors.scss` file

PR #1794, opened 2018-11-21
Integration tests (Connect #1710)

iperdomo:
I've tested 2 approaches for speeding up the tests without modifying the tests themselves:

* Only delete the data
* Create a new DB based on a _template_ instead of issuing a full rollback of the migrations

Here have some numbers (not scientific measures)

this measurements have the `collection-test` disabled as it requires a Datasource
```
== delete all data ==
real	0m30.719s
user	0m25.790s
sys	0m1.239s

== one db per test ==
real	0m52.985s
user	0m33.904s
sys	0m1.878s

== develop (migrate/rollback per test) ==
real	0m53.156s
user	0m0.411s
sys	0m0.075s
```

I think we could speed up the one of one db per test, there is an overhead of connecting to a new db, etc. But I didn't want to spend more time on this.

The proper fix is to migrate/seed and then the tests need to assert only the things they have changed and not expect an _empty_ database.

### DIFF
--- a/backend/test/akvo/lumen/fixtures.clj
+++ b/backend/test/akvo/lumen/fixtures.clj
@@ -4,14 +4,18 @@
             [akvo.lumen.test-utils
              :refer
              [import-file test-tenant test-tenant-conn]]
-            [ragtime.jdbc :as jdbc]
-            [ragtime.repl :as repl]))
+            [ragtime.jdbc :as rjdbc]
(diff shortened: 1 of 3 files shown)


kardan:
Good investigation but since the time savings are so minor compared to effort to rewrite all tests I'll close this PR for now.

Issue #1796, opened 2018-11-22
Refactor backend: Normalize namespaces pursuing modularity


tangrammer:
# Context
This is an improvement attempt to tidy up current backend namespaces. 
We'll try to improve the modular approach, that's to say to isolate main functionalities so we can understand, improve, test or extend without more deps 

# Problem or idea
if we tidy-up/improve the namespaces we'll understand better the current design. 

# Solution or next step
Refactoring

tangrammer:
Relates to #1441

tangrammer:
Extract protocols to akvo.lumen directory
```
[...]

PR #1797, opened 2018-11-22
Tidy up namespaces related to a.l.endpoint ns (Connect #1796)

tangrammer:
Relates to #1796


tangrammer:
@kardan although these changes could be approached in a different way (instead of using `commons` using `base` name as an example), at least we can group common things and remove dependencies as a first step of improving modularity of main functionalities of lumen

PR #1798, opened 2018-11-22
[#1796] Rename akvo.lumen.lumen-monitoring to akvo.lumen.monitoring

tangrammer:
Relates to #1796


tangrammer:
@dlebrero assigned to you as far as you were the main/only contributor to this monitoring feature (I'd like to ensure that the class name is not used outside lumen backend )

Issue #1808, opened 2018-11-26
Extracting multiple gives a 404


janagombitova:
This dataset: https://sierraleone.akvolumen.org/dataset/5bf97378-76c0-418c-86c1-392f12e78ad5 holds multiple `multiple` type columns. When trying to extract the data, Lumen gives a 404. 

After trying to update the dataset (what usually helped to mitigate this problem) I get a 400 error. 

![screen shot 2018-11-26 at 08 33 19](https://user-images.githubusercontent.com/12456965/48999579-5973a500-f157-11e8-9a2e-43bd650678fa.png)

tangrammer:
I've set up a PR for that
I couldn't reproduce the issue with flow survey locally nor in the production environment. So my suspicions are that flow question type wasn't properly set to `caddisfly` type in flow.

The PR affects only to update operation, so clicking update will fix this odd scenario

PR #1810, opened 2018-11-26
[#1808] Ensure multiple-type and multiple-id

tangrammer:

relates #1808 

add `multiple` details in `update` operation


tangrammer:
I've tested locally the fix, so firstly I imported in my local environment a flow dataset without caddisfly details 
then I applied the fix and clicked `update` 
the results were successful

tangrammer:
@kardan as far as this is a bug I've pushed it to develop to have it  ASAP

Issue #711, opened 2017-05-03
X and Y axis should meet at zero


nadiagorchakova:
# Context

When column values on the X axis range from negative to positive, the zero value appears in the middle of the axis:

<img width="677" alt="screen shot 2017-05-01 at 16 33 18" src="https://cloud.githubusercontent.com/assets/21127166/25651633/77914f00-2fe4-11e7-9676-4833a91a9438.png">
 
In this case, axis meet at (-30, 0) and not at (0,0)

# Problem or idea
It's more conventional to have X and Y axis meet at zero (0,0). Also, it's easier to perceive information this way. 

Example:

<img width="331" alt="screen shot 2017-05-03 at 09 44 58" src="https://cloud.githubusercontent.com/assets/21127166/25651825/55d79f8a-2fe5-11e7-9656-debf455e4719.png">

# Solution or next step
How could we solve it?

PR #1812, opened 2018-11-27
X and Y axis should meet at zero #711

finnfiddle:

### DIFF
--- a/client/src/components/charts/LineChart.jsx
+++ b/client/src/components/charts/LineChart.jsx
@@ -223,7 +223,6 @@ export default class LineChart extends Component {
                   dimensions.width - margins.right,
                 ]);
 
-
             const yExtent = extent(series.data, ({ value }) => value);
             if (yExtent[0] > 0) yExtent[0] = 0;
             const yScale = scaleLinear()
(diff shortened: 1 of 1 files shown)


finnfiddle:
@tangrammer maybe you want to review this seeing as @kardan is out of action

tangrammer:
please @finnfiddle, could you explain why this change will fix the issue?
sorry but I don't have enough context to ensure that the change is useful

client/src/components/charts/LineChart.jsx
```diff
@@ -365,7 +364,7 @@ export default class LineChart extends Component {
 
                   <AxisLeft
                     scale={yScale}
-                    left={xScale(series.data[0].timestamp)}
```
finnfiddle: this is the first item in the data array - so it would mean that the axis is inline with the first node.

client/src/components/charts/LineChart.jsx
```diff
@@ -365,7 +364,7 @@ export default class LineChart extends Component {
...
+                    left={xScale(0)}
```
finnfiddle: this means the axis will be inline with the value 0 instead of the first node

tangrammer: and that value is always `0` ?

finnfiddle: yes the ticket says it should always be `0`

Issue #1820, opened 2018-11-28
New visualisation: Polar Area


finnfiddle:
# Context
Why do we add this issue?

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

finnfiddle:
@janagombitova can I close this. Are you happy with the polar area charts?

janagombitova:
I am very happy. If you are happy and all works well, close ahead :)

PR #1821, opened 2018-11-28
New visualisation: Polar Area (Connect #1820)

finnfiddle:


finnfiddle:
just need an icon and french translations from @loicsans and then the feature will be complete

PR #1832, opened 2018-12-04
WIP Bubble vis backend

tangrammer:


backend/src/akvo/lumen/lib/aggregation/bubble.clj
```diff
@@ -29,36 +31,26 @@
...
     (lib/ok
-     {"series" [{"key" column-size-title
-                 "label" column-size-title
+     {"series" [{"key" (:title column-size)
+                 "label" (:title column-size)
                  "data" (mapv (fn [[size-value label]] {"value" size-value}) sql-response)
-                 "metadata"  {"type" column-size-type}}]
-      "common" {"metadata" {"type" column-label-type "sampled" (= (count sql-response) max-points)}
+                 "metadata"  {"type" (:type column-size)}}]
+      "common" {"metadata" {"type" nil
```
tangrammer: @finnfiddle which type should we return here? as far as we are not using  column-label anymore in this visualisation ...

tangrammer: @finnfiddle ping!

finnfiddle:
tested and works for me

Issue #1839, opened 2018-12-06
improve legend for bubble charts


finnfiddle:
# Context
The bubble chart legend suffers from very similar problems to the ones discussed about the scatter chart #1790 .

# Problems 
- legend title doesnt auto set
- no explanation of what bubble size represents
- confusing tooltip

# Solution or next step
How could we solve it?

finnfiddle:
Here is a screenshot of how it will look with these changes 
<img width="1440" alt="screen shot 2018-12-06 at 4 14 48 pm" src="https://user-images.githubusercontent.com/4628721/49589452-1f35af00-f972-11e8-835e-637b1ec3d960.png">

If @janagombitova approves of the changes then we'll need translations from @loicsans and @tangrammer  - you can see in the code what I added

PR #1840, opened 2018-12-06
improve legend for bubble charts (Connect #1839)

finnfiddle:

### DIFF
--- a/client/src/components/charts/BubbleChart.jsx
+++ b/client/src/components/charts/BubbleChart.jsx
@@ -7,6 +7,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Portal } from 'react-portal';
 import { randomColor } from '@potion/color';
+import { injectIntl, intlShape } from 'react-intl';
 
 import { labelFont } from '../../constants/chart';
 import { calculateMargins, replaceLabelIfValueEmpty, round } from '../../utilities/chart';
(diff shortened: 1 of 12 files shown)


finnfiddle:
Here is a screenshot of how it will look with these changes 
screen shot 2018-12-06 at 4 14 48 pm

If @janagombitova approves of the changes then we'll need translations from @loicsans and @tangrammer  - you can see in the code what I added

janagombitova:
@finnfiddle can I test this somewhere? dark-test? if not from the screen shot here are some comments: 
- it seems you selected Adm1_name as the bucket column, thus the variable that makes the bubbles render and gives them distinct colours. To me this column title should then be shown above the different colour specs in the legend, not above the size - like we have with bar charts 
- I then assume the size is based on First_fl_1-mean. But because the Adm1.. is above it is seems like it is the title of the size, but it is not. 
- Is it possible to distinguish the two pieces of information - what defines the bubble colours and what defines the bubble sizes as two pieces of info in the legend? is there a visual way to indicate the size in the legend? 
- I assume Salim can also edit the Legend titles, right, like we have with the bar charts?

Overall, I was hoping for something like this. Perhaps ping @Kiarii and @loicsans to see if you three can come up with a nice solution. 

screen shot 2018-12-06 at 16 19 26

finnfiddle:
good comments @janagombitova

finnfiddle:
ok how about this @janagombitova?
screen shot 2018-12-10 at 11 41 15 am

janagombitova:
Uuuu that looks much more organised and also informative in my opinion. I like. 

I assume Salim will be able to change the name in both legend items, right? 

And one thing that can use more love is alignment and scale. 
- Can the two items be better aligned? Not sure how though, but this layout is a bit off in my view. 
- And can you make the size legend a bit less prominent, smaller perhaps? It is quite in your face now. How will it work if this visualisation is added to a dashboard and does not get a lot of space on the canvas?

finnfiddle:
ok how about this now @janagombitova?
screen shot 2018-12-10 at 12 15 49 pm

janagombitova:
I think it is a good solution to get out and hear what users think! Well done 👍

Issue #1841, opened 2018-12-06
error thrown when switching from bubble chart to scatter


finnfiddle:
# Context
<img width="1440" alt="screen shot 2018-12-06 at 4 30 43 pm" src="https://user-images.githubusercontent.com/4628721/49590341-51481080-f974-11e8-9de8-d28490a23ec1.png">

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

PR #1842, opened 2018-12-07
error thrown when switching from bubble chart to scatter (Connect #1841)

finnfiddle:


finnfiddle:
this PR fixes an error that was thrown when changing from one chart type to another. When this happens the app tries to map column mappings to the new chart type. eg bucket column from one chart to bucket column of another. there was an error being thrown when changing from bubble to scatter which has been fixed at the same time as a refactor - and it can be verified to be working now with the tests that were added

kardan:
After recent changes, this PR needs some love to be able to merge.

finnfiddle:
@kardan its ready now

Issue #1858, opened 2018-12-12
Add transformation tests for all possible ops


tangrammer:
# Context
Still, we don't have test support for many ops, e.g `core/extract-multiple`

# Problem or idea
Difficult to recreate complex datasets for testing, but this situation has been improved with #1850 #1852 #1828  

# Solution or next step
Use new functionality to recreate these complex datasets easily in `test` environment and apply all available transformations

Issue #1450, opened 2018-06-08
Standardised mock dataset for testing


gabemart:
# Context
When we do feature development or manual or automated testing on Lumen, we can't use real user datasets as it would be a privacy violation.

We need mock datasets for this kind of testing.

# Problem or idea
It's easy to generate simple mock datasets with random data, but they're not very useful for testing:

These simple datasets aren't very close to our real user datasets:

* Random numbers are usually uniformly distributed. Distributions in real life are rarely uniform
* This also applies to geopoints
* They lack "naughty" values - missing values, garbled strings, typos, really long strings, strings with unusual characters, numeric values outside the expected range, etc
* For any given row, the values each column usually aren't correlated in any way (e.g. Column B value is lower when Column A valuer is higher)
* They usually don't contain FLOW specific structures like multiple option columns

Why would it be better to have a standardised mock dataset that closely resembled a real Flow dataset?

* Some bugs caught sooner - anecdotally, we currently use fairly simple datasets during feature development (I certainly do), then do enhanced testing on more complex datasets when a feature is "done". Bugs caught at this stage (or worse, only after deploying to production) take more time to fix and are worse for the user. 

* In the same way as previous point, we are likely missing out on some non-bug enhancements by not building features around realistic datasets. When we develop using a simple dataset, it is unavoidable that we to some extent optimise user experience for the simple dataset.

* Reproducing bugs between developers and other team members would be much easier if it the bug could be reproduced on the standard test dataset everyone was familiar too and had access to. It would also avoid the situation where a developer has to be given access to a production instance just to view or reproduce a bug.

* Testing visualisations in particular is hard when using simple, uniformly distributed random data. Visualisations are often used as a tool for exploring or presenting correlations and relationships between variables. If there are no correlations or relationships between values in different columns, this is harder

# What a good standard dataset would look like

This is work-in-progress, and it would be great to have input from other team members on this, particularly team members who do a lot of work with FLOW datasets

What I can think of so far:

* Wide range of column types - text, number, date, geopoint but also things like instance ID or Identifier, if possible
* **Columns**
  * Loads of columns
  * Some column names replicating the "99999999|SOME COLUMN NAME" pattern
  * Some very long column names
  * Some column names containing javascript keywords (e.g. `function`)
  * Columns which are mostly blank
  * Columns which are entirely blank
* **Text Columns**
  * Text column representations of multiple option questions
  * Text columns with typos - eg "Yes", "yes", "yse", "ye s"
[...]

Issue #1441, opened 2018-05-30
Use clojure.spec 


tangrammer:
# Context
Objectives to achieve with clojure.spec [rationale#objectives](https://clojure.org/about/spec#_objectives)
- Communication
- Unify specification in its various contexts
- Maximize leverage from specification effort
- Minimize intrusion
- Decomplect maps/keys/values
- Enable and start a dialog about semantic change and compatibility

# Problem
Problems to solve with clojure.spec [rationale#problems](https://clojure.org/about/spec#_problems)

- Docs are not enough
- Map specs should be of keysets only
- Manual parsing and error reporting is not good enough
- Generative testing and robustness
- A standard approach is needed

# Solution or next step
Having clojure 1.9 we have available clojure.spec.alpha dep so we can gradually add specifications 

## there is a couple of proof of concept branchs
https://github.com/akvo/akvo-lumen/tree/specs
https://github.com/akvo/akvo-lumen/tree/specs-lp  

# FAQs
https://gist.github.com/tangrammer/a38e1595967a5777fe368cfbe053cd3a

tangrammer:
# when do specs need to create new `src` namespaces? 
Spec ids are [qualified keywords](https://clojuredocs.org/clojure.core/qualified-keyword_q) 
So we could use existent `src` namespaces to identify our specs 
example `:akvo.lumen.transformation.engine/js-value-types` use `:akvo.lumen.transformation.engine` namespace

But there is some cases that we don't have the namespace that we need to meaningfully associate a spec
example 
`:akvo.lumen.core/int-nullable` or `:akvo.lumen.domain.core/int-nullable` 
so in these cases I've chosen to create a new `spec` namespace to allocate the expected specs
`akvo.lumen.specs.core` namespace to allocate `:akvo.lumen.specs.core/int-nullable`


# how to naming spec namespace aliases?? 
TODO: complete ....
```  
so far we have 2 kinds of nss: specs ones and normal ones 
specs ones ends in `.s`
```

tangrammer:
- [org files branch](https://github.com/akvo/akvo-lumen/tree/org-specs)  
- html doc published temporarily in https://tangrammer.github.io/akvo-lumen/backend/specs/Readme.html

PR #1863, opened 2018-12-14
Add transformation and db specs (Connect #1441 & #1450 & #1858)

tangrammer:

Relates to #1441 & #1450 & #1858 

Also conform in `test` and `dev` the new-dataset-version db call in import and transformation processes 

Thus we need some data adaptation for conforming the new-dataset-version payload, we can't use plain s/instrument facility for conforming the spec. As an alternative the PR uses robert-hooke library to check specs that need this data adaptation in `dev` and `test` environments

### DIFF
--- a/backend/src/akvo/lumen/util.clj
+++ b/backend/src/akvo/lumen/util.clj
@@ -71,12 +71,13 @@
 (defn valid-type? [s]
   (#{"text" "number" "date" "geopoint"} s))
 
-
-(defn conform
+(defn conform  
+  ([s d]
+   (when-not (s/valid? s d)
+     (throw (ex-info (str s " spec problem!")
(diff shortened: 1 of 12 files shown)


backend/src/akvo/lumen/util.clj
```diff
@@ -68,3 +71,12 @@
 (defn valid-type? [s]
   (#{"text" "number" "date" "geopoint"} s))
 
+
```
kardan: An extra new line to remove here

backend/src/akvo/lumen/util.clj
```diff
@@ -68,3 +71,12 @@
...
+(defn conform
```
kardan: I would change the order of functions follow the number of arguments.

Issue #1874, opened 2018-12-19
bubble bucket limiting is random and problematic


finnfiddle:
# Problem

When you attempt to limit the number of buckets in a bubble chart - the result shows a random selection and count of bubbles every time the vis renders.

# Solution or next step

With @kardan and @janagombitova we decided good first step is to hide the option to limit in the side bar. When we fix this we need to consider the correct ordering for limiting too.

janagombitova:
@tangrammer I believe this issue is resolved now thanks to the changes made in #2537 as we no longer show a random selection. Is that correct? Can we close it?

tangrammer:
i think so :)

PR #1875, opened 2018-12-19
bubble bucket limiting is random and problematic #1874

finnfiddle:


finnfiddle:
I just commented it out so that we can reenable this feature when we fix it properly

Issue #1877, opened 2018-12-20
Shared dashboard url does not work


janagombitova:
# Context
Why do we add this issue?

# Problem or idea

An already shared dashboard url (already generated) does not work. By opening it is says No such dashboard or visualisation.  When creating a new dashboard sharing works fine

Example: https://marsfood.akvolumen.org/s/Ea_KbD6MqUY

![screen shot 2018-12-20 at 08 36 14](https://user-images.githubusercontent.com/12456965/50270647-923f2b00-0432-11e9-82b2-7627fe3d9b42.png)

# Solution or next step
How could we solve it?

janagombitova:
After taking a closer look there seems to be something wrong with the visualisation in the dashboard 
![screen shot 2018-12-20 at 08 42 56](https://user-images.githubusercontent.com/12456965/50270908-5eb0d080-0433-11e9-9c1a-3400dda82890.png)

janagombitova:
Ok. I tried to reproduce the case and I think I know what is going on. We made a change where if a visualisation does not have variables to generate the visual it is not available to add to a dashboard. 

In this case, the dashboard holds an empty visualisation. A pie chart where only the data set is selected but no bucket column. As a result the visualisation does not generate and the dashboard does not share. 

Dashboard url: https://marsfood.akvolumen.org/s/Ea_KbD6MqUY
Visualisation: https://marsfood.akvolumen.org/visualisation/5bb2da4d-65a8-4562-9ca1-00f37f33490c

When I tried to reproduce the case I create a same pie chart (https://demo.akvolumen.org/visualisation/5c1b492b-25f6-4e89-9886-840e3043ebb9) but I cannot add it to a visualisation as it is not available to be selected in the dashboard visualisation side panel. 

By fixing the visualisation or removing it from the dashboard, the dashboard url should work properly. The question is, how do we make Lumen still work in such cases? We prevent them from happening now, but what about backwards compatibility?

kardan:
https://sentry.io/akvo-foundation/production/issues/815753145/?query=is%3Aunresolved

PR #1880, opened 2018-12-20
[#1877] Wrap fetch visualisation calls in /public/shared in a try/catch block

tangrammer:


finnfiddle:
LGTM

Issue #886, opened 2017-06-26
Documentation for Lumen Data import format and functionality


LarsSjogreen:
# Context
We need to document how data import works at the moment so that someone that needs to make data readable from Lumen knows how to do this and what functionality to expect.

# Problem or idea
We have (more or less) decided on how data import should work for the moment (CSV, available from URL). This needs to be documented. Maybe we should also write about limitations and about what's next (authorization, limited (or no) support for adding properties at the end of the CSV etc.)?

# Solution or next step
Documentation will be written and put in a .md file under /doc/integration, also linking to it from README.md in the root.

janagombitova:
@tangrammer can you check this issue if it is still relevant or not?

tangrammer:
we can close it (doing it now 😄 )

PR #1893, opened 2019-01-07
Update InputFormat.md [doc] (Connect #886)

tangrammer:
Relates #886


tangrammer:
@kardan could you review this update and evaluate if we can close  #886 issue ?
IMO after this last changes we could close it

Issue #1900, opened 2019-01-09
Bar visualisation client UI doesn't call aggregation backend API


tangrammer:
# Context
Some bar visualization options aren't sent to the backend via /aggregation API
Detected ones so far:
+ sub-bucket column value
+ truncateSize
+ sort
+ adding/removing filters don’t aggregate either

# Problem or idea
client bug

# Solution or next step

janagombitova:
Hi @finnfiddle and @tangrammer, what is the status of this issue? Is it completed, tested and can be closed or is there still more to be done?

PR #1907, opened 2019-01-11
Bar visualisation client UI doesn't call aggregation backend API #1900

finnfiddle:


finnfiddle:
I went through all the chart types not just bar

PR #1910, opened 2019-01-15
Add import and aggregation dev utils

tangrammer:
Just `dev` utils to develop/check easily import and aggregation functions.


backend/dev/src/dev/aggregation_dev.clj
```diff
@@ -0,0 +1,27 @@
+(ns dev.aggregation-dev
```
dlebrero: rename to dev.aggregation?

backend/dev/src/dev/import_dev.clj
```diff
@@ -0,0 +1,46 @@
+(ns dev.import-dev
```
dlebrero: rename to dev.import?

PR #1913, opened 2019-01-15
Speed up aggregation dataset query

tangrammer:


tangrammer:
@dlebrero  I didn't measure but should be faster as far as it didn't need so much joins and data to return

Issue #1902, opened 2019-01-10
Shared dashboards blank in Internet Explorer


janagombitova:
# Issue 

Two partners report that they can't view publicly shared dashboards in Internet Explorer 11. They are getting a blank page

# Created via Reamaze:

Link: /admin/conversations/akvolumen-support-request-number-176695-from-geert-at-akvo-dot-org

Issue #1772, opened 2018-11-14
for...in polyfill bug old firefox


finnfiddle:
# Context
Why do we add this issue?
a bug in an old version of firefox is causing the app to break: https://bugzilla.mozilla.org/show_bug.cgi?id=1101653
<img width="1440" alt="screen shot 2018-11-14 at 12 41 59 pm" src="https://user-images.githubusercontent.com/4628721/48482008-1b51b980-e818-11e8-9f48-508a7302d770.png">

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

PR #1914, opened 2019-01-15
Lumen blank internet explorer #1902 2

finnfiddle:
This will undo the fix we made for #1772. But I think #1902 is more important than #1772. So I think we should merge this and reopen #1772.


finnfiddle:
cant work out why the build failed - in the video it looks like the connection timed out or something inside the cypress test. restarting

finnfiddle:
closing because #1918 is better done

PR #1918, opened 2019-01-16
#1902 ie11 blank

finnfiddle:


finnfiddle:
the cause of the problem was that the `query-string` module was letting in uncompiled es6 syntax JS into the bundle. It had been ignored as a way to fix a separate problem. So I just replaced the `query-string` module with an alternative and problem should be solved now

Issue #1866, opened 2018-12-17
cypress tests for changing visualisation types


finnfiddle:
# Context
This is something that has proven to get broken more easily than others and it would be nice to have tests for it.

PR #1932, opened 2019-01-23
#1866 cypress test fixes

finnfiddle:


finnfiddle:
in the charts when we reformat the data we were trying to destructure values and not checking if they were objects before hand - and sometimes a null value would appear in place of the object and this was causing an exception to be thrown because you cant destructure from a null value. Not sure why some values are null in the cypress tests though

Issue #1395, opened 2018-05-01
Percantage Stacked bar charts


nadiagorchakova:
# Context
I received a suggestion from Rajashi and Guneet about stacked bar charts.
They've been working with Watershed programme and have come across a few problems when visualising project results.

# Problem or idea
Comment from Rajashi:

"Suppose we are comparing two villages (with different population sizes) and looking at different categories of residents within each village, we would like to see two stacked bars of equal size (i.e. 100%) divided by the categories (%) of the residents. Right now we have the stacked bars with the proportions displayed according to the size of the population in each village."

# Solution or next step
That's an example of a 100% stacked bar chart I found online:

<img width="590" alt="screen shot 2018-04-30 at 15 06 38" src="https://user-images.githubusercontent.com/21127166/39466006-d1dc78bc-4d26-11e8-9944-387905a4f5ce.png">

janagombitova:
Here are some more examples of the same bar charts, stacked and using percentage rather than actual values that our partners made from their data running a water quality testing proejct in Ethiopia

<img width="591" alt="screen shot 2018-10-30 at 14 37 49" src="https://user-images.githubusercontent.com/12456965/47724185-73a08d00-dc56-11e8-88ef-017d9de26e63.png">
<img width="479" alt="screen shot 2018-10-30 at 14 38 27" src="https://user-images.githubusercontent.com/12456965/47724189-74d1ba00-dc56-11e8-85d3-aff2afb7dacb.png">

janagombitova:
There is one more open issue linked to this one. But overall everything works well and we are ready.

PR #1937, opened 2019-01-30
Percentage stacked bars (Connect #1395)

finnfiddle:


finnfiddle:
looking at it I realize I still need to do the following:

- show percentage values on the axis
- implement the same for horizontal stacked bars
- make the legend hovering work again

finnfiddle:
@kardan ok its ready now

Issue #1936, opened 2019-01-29
Flow data import is not working


janagombitova:
# Context
Why do we add this issue?

We can import data in different ways to Lumen, once is directly from Flow. Salim types in his Flow url, selects the folder, survey and form he wants to import the data. 

# Problem or idea
What is the cause?

When trying to import the data from Flow after selecting the Flow instance I get a 404 error in the console. 

![screen shot 2019-01-29 at 09 06 10](https://user-images.githubusercontent.com/12456965/51893695-ee885980-23a5-11e9-84d7-401c0aec9917.png)

# Solution or next step
How could we solve it?

kardan:
![screenshot 2019-01-29 at 10 36 33](https://user-images.githubusercontent.com/31837/51898739-dd454a00-23b1-11e9-8a38-8905f4f438e2.png)

kardan:
For me it's not a 404 but still not working

kardan:
Last working commit (pre client-request-error-handling)  https://github.com/akvo/akvo-lumen/commit/411cfe60705299d8a1c9557418fc49504950be9d

PR #1938, opened 2019-01-30
Fix flow import modal (Connect #1936)

kardan:


kardan:
Closed this since I branched of master and not develop

PR #2023, opened 2019-03-05
Issue/2022 exporter feature flag

kardan:

### DIFF
--- a/exporter/index.js
+++ b/exporter/index.js
@@ -22,7 +22,7 @@ const validation = {
 };
 
 let sentryClient;
-const sentryIsEnabled = () => !(sentryClient == null);
+const sentryIsEnabled = () => sentryClient;
 
 const enableSentry = !(
   process.env.SENTRY_DSN == null
--- a/exporter/index.js
+++ b/exporter/index.js
@@ -24,11 +24,10 @@ const validation = {
 let sentryClient;
 const sentryIsEnabled = () => sentryClient;
 
-const enableSentry = !(
-  process.env.SENTRY_DSN == null
-  && process.env.SENTRY_RELEASE == null
-  && process.env.SENTRY_ENVIRONMENT == null
-  && process.env.SENTRY_SERVER_NAME == null);
+const enableSentry = process.env.SENTRY_DSN
+  && process.env.SENTRY_RELEASE
+  && process.env.SENTRY_ENVIRONMENT
+  && process.env.SENTRY_SERVER_NAME;
 
 if (enableSentry) {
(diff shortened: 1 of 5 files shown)


exporter/index.js
```diff
@@ -22,29 +22,34 @@ const validation = {
 };
 
 let sentryClient;
-const sentryIsEnabled = () => sentryClient == null;
+const sentryIsEnabled = () => !(sentryClient == null);
```
finnfiddle: splitting hairs but could be:
```const sentryIsEnabled = () => sentryClient;```

File: exporter/index.js [hunk 2]
```diff
@@ -22,29 +22,34 @@ const validation = {
...
 
-if (process.env.SENTRY_DSN == null && process.env.SENTRY_RELEASE == null
-  && process.env.SENTRY_ENVIRONMENT == null && process.env.SENTRY_SERVER_NAME == null) {
-  console.log('Sentry not configured');
-} else {
+const enableSentry = !(
```
finnfiddle: not sure this logic seems correct. should all of them be set or at least one? I think all is correct

exporter/index.js
```diff
@@ -22,29 +22,34 @@ const validation = {
...
   sentryClient = Sentry.getCurrentHub().getClient();
-  console.log('Sentry initiated');
+} else {
+  console.log('Skipping Sentry');
 }
 
 const captureException = (error, runId = '') => {
-  console.error(`Exception captured for run ID: ${runId} -`, error);
-  if (sentryIsEnabled) Sentry.captureException(error);
+  console.error(`Exceptiqon captured for run ID: ${runId} -`, error);
```
finnfiddle: another split hair

exporter/index.js
kardan: All of them should be set. Have I fooled myself?

exporter/index.js
kardan: What's the issue?

exporter/index.js
finnfiddle: I think if one is not equal to `null` then `enableSentry` will be `true`. Have _i_ fooled myself?

exporter/index.js
finnfiddle: the spelling

exporter/index.js
kardan: Yes, I think you're right. I was too focused trying to support both null and undefined & fooled myself.

exporter/index.js
kardan: :-), yes that was an alternative spelling.

Issue #2024, opened 2019-03-06
Password protect shows error but works 


janagombitova:
# Issue 

When setting up a dashboard to be password protected Lumen shows the error that it failed (in the UI) even though it worked. 

![screen shot 2019-03-06 at 08 37 26](https://user-images.githubusercontent.com/12456965/53863706-227d1d00-3feb-11e9-916c-6f3b27836636.png)

## Steps to take to recreate
1. Created a dashboard and shared it
2. Turn on password protection and immediately get the error
3. Added the password and hit save
4. Refresh shared dashboard and the password protection is enabled

# Created via Reamaze:

Link: /admin/conversations/akvolumen-support-request-number-192596-from-camille-at-akvo-dot-org

kardan:
The problem is that once the user hit the "protected switch" the client issue a PUT request to the backend. This request has a payload with {"protected" true} but no password. Hence the backend responds with a 400 error. If the user continues to add a password and hit the save button all is fine. 

We should probably not send a request from the client until the user hits the save button.

Not sure what's going on with the funky circle floating around the error message.

PR #2034, opened 2019-03-14
Fix for error protecting dashboards (Connect #2024)

kardan:
There are some outstanding issues. For example, if a password is set. And then the user tries to update the password to a too short one. He/she will get an error dialog at the bottom of the screen (the backend responded with error) but the alert in the modal will be all happy with a green succeded. But I think that is another issue than this initial one. I tried to fix that one too, but backed out of it since it looked to me like a little bit of rearchitecting is needed. @finnfiddle if you think different ping me.

### DIFF
--- a/client/src/containers/Dashboard.jsx
+++ b/client/src/containers/Dashboard.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { isEmpty, cloneDeep } from 'lodash';
 import get from 'lodash/get';
-import set from 'lodash/set';
 import { intlShape, injectIntl } from 'react-intl';
 import BodyClassName from 'react-body-classname';
 
--- a/client/src/containers/Dashboard.jsx
+++ b/client/src/containers/Dashboard.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { isEmpty, cloneDeep } from 'lodash';
 import get from 'lodash/get';
+import set from 'lodash/set';
 import { intlShape, injectIntl } from 'react-intl';
 import BodyClassName from 'react-body-classname';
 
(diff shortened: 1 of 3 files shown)


client/src/containers/Dashboard.jsx
```diff
@@ -432,8 +433,7 @@ class Dashboard extends Component {
 
   handleToggleShareProtected(isProtected) {
     this.setState({ passwordAlert: null });
-    const dashboard = getDashboardFromState(this.state.dashboard, true);
-    this.props.dispatch(actions.setShareProtection(dashboard.shareId, { protected: isProtected }));
+    set(this.state, 'dashboard.protected', isProtected);
```
finnfiddle: this doesnt look right. Maybe:
```
this.setState({ 'dashboard.protected': isProtected });
```

finnfiddle: or actually:
```
this.setState({ 
  dashboard: {
    ...this.state.dashboard,
    protected: isProtected,
  },
});

Comment (finnfiddle) [reply client/src/containers/Dashboard.jsx]: if you just set it manually (ie not using the `setState` method then the observers wont get triggered

File: client/src/containers/Dashboard.jsx [hunk 2]
```diff
@@ -432,8 +433,7 @@ class Dashboard extends Component {
 
   handleToggleShareProtected(isProtected) {
     this.setState({ passwordAlert: null });
-    const dashboard = getDashboardFromState(this.state.dashboard, true);
-    this.props.dispatch(actions.setShareProtection(dashboard.shareId, { protected: isProtected }));
```
finnfiddle: also why are we removing this API request? Isnt this how we save the config?

kardan: This get's called when we enable the radio button & yet have to type a password. The backend will respond with an error message (no password) and the client will see the error. This removes the initial request. And the user has to hit "save" once he/she entered a valid password. At that point a request with all config necessary to enable protection is available.

Issue #1993, opened 2019-02-19
Reverse geocoding fails 


janagombitova:
# Context
Why do we add this issue?

# Problem or idea

When trying to use the Reverse Geocode transformation on live production I get a 502 error 

<img width="1267" alt="screen shot 2019-02-19 at 19 14 58" src="https://user-images.githubusercontent.com/12456965/53037770-5dc00d80-347b-11e9-9fbb-5b7b8f3e46fa.png">

# Solution or next step
How could we solve it?

kardan:
I tried this to see what happened and it worked fine for me. Obviously not the same case. Would be very valuable if we could retry to make sure it's still an issue. If it's an issue it would be great with some more context.

janagombitova:
Will ask Joy to confirm if the issue is still present and then update here

kardan:
@joycarpediem are you able to check if this is still an issue?

tangrammer:
I couldn't reproduce it either, locally nor in `demo` instance

pasting here logs link related
https://console.cloud.google.com/logs/viewer?interval=PT1H&project=akvo-lumen&authuser=1&minLogLevel=0&expandAll=false&customFacets&limitCustomFacetWidth=true&scrollTimestamp=2019-02-19T23%3A43%3A09.127611621Z&timestamp=2019-02-19T23%3A50%3A52.149000000Z&dateRangeStart=2019-02-19T22%3A51%3A12.658Z&dateRangeEnd=2019-02-19T23%3A51%3A12.658Z&advancedFilter=resource.type%3D%22http_load_balancer%22%0Aresource.labels.target_proxy_name%3D%22k8s-tps-default-akvo--a537c5dbd91a049b%22%0Aresource.labels.zone%3D%22global%22%0Aresource.labels.project_id%3D%22akvo-lumen%22%0Aresource.labels.url_map_name%3D%22k8s-um-default-akvo--a537c5dbd91a049b%22%0Aresource.labels.forwarding_rule_name%3D%22k8s-fws-default-akvo--a537c5dbd91a049b%22%0Atimestamp%3D%222019-02-19T23%3A16%3A30.339548246Z%22%0AinsertId%3D%22bxvfvsf88bhf9%22

janagombitova:
I tried to recreate the data we tested with Joy on https://lumen.akvotest.org/library. Everything is the same besides the text column, but I assume that is not such a big deal. 

The steps to take are: 
[...]

PR #2035, opened 2019-03-14
Change transformation logic to be async (Connect #1993)

tangrammer:

- [X] Review transformation ops that require a different view in client side (eg: merge datasets)

Relates to #1993


tangrammer:
@finnfiddle Could you review the client part of this PR?

backend/src/akvo/lumen/lib/import.clj
```diff
@@ -52,8 +52,7 @@
                                :transformations []})
     (update-job-execution conn {:id             job-execution-id
```
kardan: Aligned values - buh hu... :-)

tangrammer: :-)

PR #2041, opened 2019-03-19
WIP Issue 1949 filter library

tangrammer:

## Thoughts to share & to agree:
we need to ensure that the user has rights to *operate* in/with a dataset and the user has rights to *view* data with dataset

### 1. where should we place dataset authentication logic?
Datasets are spread out different entities like aggregations, transformations, visualizations, collections or dashboards. 
At first glance, it seems that we should place dataset authentication logic on the endpoints in a declarative way (using `edn` and middleware). But due to the interrelations between dataset entities and the rest of the other domain entities, makes very difficult to ensure that some response carries dataset with no privileges for the current user. As a possible solution in the view, we could define some kind of entity-response-protocol that always is called in the response (and maybe the same solution could be applied to the request  🤔 )
- Putting the logic in the endpoint layer leaves to the developer the duty to take care of using lib.dataset ns without enough rights from other endpoints or layers ... (e.g: endpoint.public=> lib.public => lib.dataset)
- there is some ambiguity with the authentication behavior with `/share` endpoint, thus the authentication logic, in this case, is ignored, or displaced by a password-based protected view
- 2 different users with different rights could work in the same dataset/vis. The user with more rights could invalidate the other user access data, combining the "allowed" data with "unallowed" data.  

### 2. maybe it is worthing to grab the allowed visualizations together with allowed datasets in the middleware?
to discuss

### 3. views/ops that we need to take into account
- [x] library => dataset/all
- [x] dataset detail
- [x] visualization detail
- [x] aggregation detail
- [x] dashboard
- [ ] collection
- [x] lib.public => dataset/fetch-metadata ... allowing access thus user should have rights at the time they create the share link ... check with Daniel & Jana about this asumption
- [ ] Combined datasets: through transformations

### 4. questions: 
- auth-dataset-opt-middleware ?
- cache dataset permissions having a short expired time ?

### middleware approach using reitit to get routes (... and then matching with desired endpoints)

below routes in `dev` env

```clojure
(map (juxt first (comp #(disj % :middleware) set keys second))
                                        (rc/routes router))
```

```clojure
(["/env" #{:get}]
 ["/healthz" #{:get}]
 ["/metrics" #{:get}]
 ["/api/aggregation/:dataset-id/:visualisation-type" #{:get}]
 ["/api/collections" #{:get :post}]
 ["/api/collections/:id" #{:get :delete :put}]
 ["/api/dashboards" #{:get :post}]
 ["/api/dashboards/:id" #{:get :delete :put}]
 ["/api/datasets" #{:get :post}]
 ["/api/datasets/:id" #{:get :delete :put}]
 ["/api/datasets/:id/meta" #{:get}]
 ["/api/datasets/:id/update" #{:post}]
 ["/api/library" #{:get}]
 ["/api/transformations/:dataset-id/transform" #{:post}]
 ["/api/transformations/:dataset-id/undo" #{:post}]
 ["/api/visualisations" #{:get :post}]
 ["/api/visualisations/maps" #{:post}]
 ["/api/visualisations/rasters" #{:post}]
 ["/api/visualisations/:id" #{:get :delete :put}]
 ["/api/files" #{:post}]
 ["/api/files/:id" #{:patch :head}]
 ["/api/rasters" #{:get :post}]
 ["/api/rasters/:id" #{:get :delete}]
 ["/api/exports" #{:post}]
 ["/api/data-source/job-execution/:id/status/:status" #{:delete}]
 ["/api/job_executions/:id" #{:get}]
 ["/api/multiple-column" #{:get}]
 ["/api/split-column/:dataset-id/pattern-analysis" #{:get}]
 ["/api/resources" #{:get}]
 ["/api/shares" #{:post}]
 ["/api/shares/:id" #{:put}]
 ["/api/tiers" #{:get}]
 ["/api/admin/users" #{:get}]
 ["/api/admin/users/:id" #{:patch :delete}]
 ["/api/admin/invites" #{:get :post}]
 ["/api/admin/invites/:id" #{:delete}]
 ["/verify/:id" #{:get}]
 ["/share/:id" #{:get}]
 ["/local-development/local-server/:file" #{:get}]
 ["/spec/sample/:spec-ns/:spec-id" #{:get}]
 ["/spec/describe/:spec-ns/:spec-id" #{:get}]
 ["/spec/conform/:spec-ns/:spec-id" #{:post}])
```

### DIFF
--- a/backend/src/akvo/lumen/auth.clj
+++ b/backend/src/akvo/lumen/auth.clj
@@ -1,15 +1,15 @@
 (ns akvo.lumen.auth
   (:require [akvo.commons.jwt :as jwt]
-            [cheshire.core :as json]
-            [akvo.lumen.protocols :as p]
             [akvo.lumen.component.flow :as c.flow]
-            [hugsql.core :as hugsql]
-            [clojure.tools.logging :as log]
+            [akvo.lumen.component.keycloak :as keycloak]
+            [akvo.lumen.protocols :as p]
+            [cheshire.core :as json]
             [clj-http.client :as client]
             [clojure.set :as set]
-            [clojure.string :as string]
             [clojure.spec.alpha :as s]
-            [akvo.lumen.component.keycloak :as keycloak]
+            [clojure.string :as string]
+            [clojure.tools.logging :as log]
+            [hugsql.core :as hugsql]
             [integrant.core :as ig]
             [ring.util.response :as response]))
 
@@ -87,7 +87,7 @@
 
 (hugsql/def-db-fns "akvo/lumen/lib/dataset.sql")
 
-(defn wrap-ds-auth
+(defn wrap-auth-datasets
   "Add to the request the auth-datasets allowed to the current authenticated user,
   using flow-api check_permissions"
   [tenant-manager flow-api]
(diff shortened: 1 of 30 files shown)


backend/src/akvo/lumen/auth.clj
```diff
@@ -81,3 +84,65 @@
...
   (s/keys :req-un [::keycloak]))
+
+(hugsql/def-db-fns "akvo/lumen/lib/dataset.sql")
+
+(defn wrap-auth-datasets
+  "Add to the request the auth-datasets allowed to the current authenticated user,
+  using flow-api check_permissions"
+  [tenant-manager flow-api]
+  (fn [handler]
+    (let [auth-calls #{(with-meta ["/api/library"] {:methods #{:get}})
```
dlebrero: I guess this will be moved to the endpoints themselves?

tangrammer: I guess too 👍

tangrammer:
closing this PR thus the work is redesigned, splitted and organised on other several PRs

Issue #1949, opened 2019-02-05
Filter library based on Flow permissions


kardan:
1. Get library
2. Construct a list of forms per Flow instance 
```
{:flow-instance-a #{form-id-a form-id-b ... }
 :form-instance-b #{form-id-c form-id-d ... }}
```
3. Filter items in library that have Flow data via the Flow API (API endpoint to be designed)
4. Return library for user to see

## API Endpoint
a. get a list of all allowed forms for current user (get-list user) -> [forms]
b. post a set of forms and do the actual filtering in the Flow API (allowed-forms user #{forms}) -> filtered forms set.
c. ...

### Response
Maybe it's not enough with a list of forms but we should get more detailed permissions (view, edit, delete) on all forms?

kardan:
Ref https://github.com/akvo/akvo-flow-api/issues/146

dlebrero:
Note that the API implemented requires as input 
```
[{:instance_id "akvoflowsandbox"
:survey_id "152342023"}
{:instance_id "akvoflowsandbox"
:survey_id "148412329"}]
```

The response would be the same vector, but with only the surveys that the user is allowed to see.
The endpoint is a POST to /check_permissions. 

See https://github.com/akvo/akvo-flow-api/blob/3eea8f752469c91fd3586a3d3308b688c6dc1ca0/api/src/clojure/org/akvo/flow_api/endpoint/flumenfly.clj#L41 for an example

kardan:
Related Flow API work 
https://github.com/akvo/akvo-flow-api/issues/146
https://github.com/akvo/akvo-flow-api/pull/147
Maybe the fact that the Flow API work is already done a feature flag is not needed?

dlebrero:
We still want a safe and smooth transition, which means that we want the capability of going back to a well tested version that we know works.

Note that blue/green is not enough

tangrammer:
### conclusions

+ use a tree-uuid-auth graph (in the future cames from auth-service)
+ authentication on endpoints ids (using tree-uuid-auth)
+ filtering auth-data in model and logic layers (using tree-uuid-auth and hugsql)
+ spec based validation with binding??? https://clojurians.slack.com/archives/C1B1BB2Q3/p1554104057090200

#### miscelaneus 
+ dataset tranformation dataset-combination denormalise in other dataset.table.column (thus jsonp generic data is difficult/complex to sql) ( .... should be migrated in background process or executed on demand.... ????)

[...]

PR #2054, opened 2019-03-28
[#1949] Define a flow component

tangrammer:

Define a flow component that takes care of flow api details such as keycloak auth

Config edn is adapted to (and improved with)  this new component

### DIFF
--- a/backend/src/akvo/lumen/component/flow.clj
+++ b/backend/src/akvo/lumen/component/flow.clj
@@ -1,0 +1,56 @@
+(ns akvo.lumen.component.flow
+  "wrap flow and flow api logic and data related"
+  (:require [integrant.core :as ig]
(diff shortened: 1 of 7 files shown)


backend/src/akvo/lumen/component/flow.clj
```diff
@@ -0,0 +1,56 @@
+(ns akvo.lumen.component.flow
+  "wrap flow and flow api logic and data related"
+  (:require [integrant.core :as ig]
+            [akvo.lumen.component.keycloak :as c.keycloak]
+            [clj-http.client :as http]
+            [clojure.tools.logging :as log]
+            [clojure.spec.alpha :as s]))
+
+(defn access-token
```
tangrammer: from line 9 to line 28, this code comes from backend/src/akvo/lumen/lib/import/flow_common.clj
so just wrapping flow stuff into flow component

PR #2055, opened 2019-03-28
[#1949] Add wrap-auth-datasets middleware and DBQueryService adapter hugsql protocol 

tangrammer:
##### by matching "url" requested 
+ url matched using reitit 
```clojure
(contains? #{"/api/datasets" "/api/library"} (:template (:reitit.core/match request)))
```
##### by matching "feature-backend-flag" activated.
Logic based on  jwt-claim family_name value
```clojure
(str/includes? (:given_name (clojure.walk/keywordize-keys (:jwt-claims request))) "$auth$")
```

 ### Adding `DBQueryService` (adapter) protocol to wrap hugsql queries
The purpose of this protocol is to pass an extra parameter related to authorization, so we can add (with activated feature flag) an extra filter to the related protected SQLs (eg: https://github.com/akvo/akvo-lumen/pull/2055/files#diff-15a1115682b54d65d59a28442efe1c3dR8)

```clojure
(defprotocol DBQueryService
  (query
    [_ qqname param-data options command-options]
    [_ qqname param-data options]
    [_ qqname param-data]
    [_ qqname ]))
```

### Adding AuthorisedDBQueryService `DBQueryService` protocol impl
so, if middleware is applied we include in the request a ready to use database-query-service that will use the `{:auth-datasets [] :auth-visualisations [] :auth-dashboards [] :auth-collections []}` data in the authorized SQLs
```clojure
(defrecord AuthorisedDBQueryService [tenant-conn authorised-uuid-tree]
  p/DBQueryService
  (p/query [this fun param-data options command-options]
    (log/info :AuthorisedDBQueryService :query fun :param-data param-data :options options :command-options command-options :authorised-uuid-tree authorised-uuid-tree)
    (apply fun [(:tenant-conn this) (merge param-data authorised-uuid-tree)
                options command-options]))
  (p/query [this fun param-data options]
    (p/query this fun param-data options nil))
  (p/query [this fun param-data]
    (p/query this fun param-data nil))
  (p/query [this fun]
    (p/query this fun nil)))

(defn new-dbqs
  "`authorised-uuid-tree`: {:auth-datasets [] :auth-visualisations [] :auth-dashboards [] :auth-collections []}"
  [tenant-conn authorised-uuid-tree]
  (AuthorisedDBQueryService. tenant-conn authorised-uuid-tree))

```


backend/src/akvo/lumen/lib/auth.clj
```diff
@@ -0,0 +1,77 @@
...
+   [clojure.spec.alpha :as s]
+   [clojure.string :as str]
+   [clojure.tools.logging :as log]
+   [hugsql.core :as hugsql]
+   [integrant.core :as ig]
+   [reitit.core :as rc]))
+
+(defrecord AuthorisedDBQueryService [tenant-conn authorised-uuid-tree]
+  p/DBQueryService
+  (p/query [this fun param-data options command-options]
```
kardan: Why "fun" and not just "f"?

tangrammer: fun is more funny :)

tangrammer: and i have seen it in other code bases too

backend/src/akvo/lumen/lib/auth.clj
```diff
@@ -0,0 +1,77 @@
...
+    (p/query this fun nil)))
+
+(defn new-dbqs
+  "`authorised-uuid-tree`: {:auth-datasets [] :auth-visualisations [] :auth-dashboards [] :auth-collections []}"
+  [tenant-conn authorised-uuid-tree]
+  (AuthorisedDBQueryService. tenant-conn authorised-uuid-tree))
+
+(hugsql/def-db-fns "akvo/lumen/lib/dataset.sql")
+
+(defn match-by-jwt-family-name?
```
kardan: If we have the energy, I think we should lift this out of this namespace and do something more holistic so we can reuse for other new features. Maybe even add another claim for "staff" or "bleeding-edge".. It feels like looking for given names will be short lived.

tangrammer:
@kardan 

> The "backend/src/akvo/lumen/lib/auth.clj" namespace is very wide, I would love to see some line changes.

let's see how auth-dataset-details come out to refactor the current logic, i was trying to not overdesign more with more assumptions. Anyway, yep, I'll keep an open eye on this ns to improve its form before fully finishing #1949

backend/src/akvo/lumen/lib/auth.clj
tangrammer: resolving conversation after agreeing that we keep this family-name flag on this development level

Issue #2057, opened 2019-04-02
Inconsistency data stored through PUT api/visualisation/:id


tangrammer:
# Context
There is no validation in `spec` attribute  body payload on `PUT /api/visualisation/:id` , and current `client` side is sending any spec when we switched visualisation-type  using the horizontal scroller 
<img width="402" alt="Screen Shot 2019-04-02 at 15 38 56" src="https://user-images.githubusercontent.com/731829/55407004-7da81e00-555d-11e9-9b1c-fb5bbfc1cd7a.png">

As a result of both worlds, we're storing wrong specs in the `visualisation.spec` table

Example of `area-type-spec` with `pie` visualisation-type:
```clojure
{:name "Untitled visualisation",
                 :visualisationType "pie",
                 :type "visualisation",
                 :created 1554194706677,
                 :modified 1554200128582,
                 :datasetId "5ca2635b-d3be-4ab1-9ec8-8a8f1ad9a08a",
                 :author author,
                 :spec
                 {:showLegend true,
                  :metricColumnX "c5",
                  :metricAggregation "q1",
                  :filters [],
                  :axisLabelYFromUser true,
                  :legendPosition "left",
                  :bucketColumn nil,
                  :showLabels true,
                  :metricColumnY "c4",
                  :axisLabelX "jor",
                  :legendTitle nil,
                  :axisLabelXFromUser true,
                  :axisLabelY "Temperature - q1",
                  :version 1,
                  :sort nil},
                 :status "OK",
                 :id "5ca32112-43c1-4c91-9ca3-a79278906660"}
```

# Solution or next step
`Backend` should add specs to conform on these endpoints 
`Client` should not send wrong spec to the backend

It seems that `client` has some work done in this aspect thus `client` only calls `/api/aggregation/uuid/vis-type?query=xxxx` whith proper spec data. So not called when we switched visualisation-type  using the horizontal scroller

tangrammer:
@finnfiddle could you take a look at the `client` part of this issue?

finnfiddle:
@tangrammer yes sure

tangrammer:
#2063 depends on this issue too

janagombitova:
This was merged to develop on 22 April. I am closing the issue under the assumption that it is done. If not, @tangrammer please reopen it. Thanks

PR #2058, opened 2019-04-02
Authorize vis endpoint (Connect #1949)

tangrammer:
- [ ] Depends on #2057 

### visualisation table

```
 | Table "public.visualisation" |                          |           |          |         |
 |------------------------------+--------------------------+-----------+----------+---------|
 | Column                       | Type                     | Collation | Nullable | Default |
 | id                           | text                     |           | not null |         |
 | dataset_id                   | text                     |           |          |         |
 | name                         | text                     |           | not null |         |
 | type                         | text                     |           | not null |         |
 | spec                         | jsonb                    |           | not null |         |
 | author                       | jsonb                    |           |          |         |
 | created                      | timestamp with time zone |           |          | now()   |
 | modified                     | timestamp with time zone |           |          | now()   |


```

We should be able to authorise vizs based on dataset-id except for visualisation.type=`map`

#### Trying to select the uuid from visualisation.map.layers 
```
select  id, name, replace(((jsonb_array_elements(spec->'layers')->'datasetId')::text), '"', '')::uuid as di  from visualisation
 where type='map'


 | id                                   | name | di                                   |
 |--------------------------------------+------+--------------------------------------|
 | 5ca4c9e0-9c18-4bf2-ab45-c1b135365fb5 | n1   | 5ca4ca3f-ef0f-4970-9c27-670dc6b9f3ae |
 | 5ca4c9e0-9c18-4bf2-ab45-c1b135365fb5 | n1   | 5ca4ca3f-ef0f-4970-9c27-670dc6b9f3ae |
 | 5ca5af5c-964f-4d8c-b66f-b4664cb720b1 | n2   | 5ca4ca3f-ef0f-4970-9c27-670dc6b9f3ae |
 | 5ca5b027-267d-4d11-af56-9f579b8fc0c2 | n3   | 5ca4ca3f-ef0f-4970-9c27-670dc6b9f3ae |
 | 5ca5afb8-a773-4509-ab0f-640703625dd9 | n4   | 5ca4ca3f-ef0f-4970-9c27-670dc6b9f3ae |

```

**BUT** we can't use `di` column in where clause https://stackoverflow.com/questions/38040631/postgresql-does-not-accept-column-alias-in-where-clause
or even use the expression directly in the where/join part 

```
select  id, name, replace(((jsonb_array_elements(spec->'layers')->'datasetId')::text), '"', '')::uuid as di  from visualisation
left join dataset on dataset.id=(jsonb_array_elements(spec->'layers')->'datasetId')::text
 where type='map' 

ERROR:  argument of JOIN/ON must not return a set
LINE 2: left join dataset on dataset.id=(jsonb_array_elements(spec->...
```

**Then** we need to process the data in logic layer (clj libs nss) to ensure db authorisation working with jsonb columns


tangrammer:
WIP (tidy up this PR) on #2068

Issue #2091, opened 2019-04-23
Calling flow-api/check_permissions takes 4,5 seconds as average


tangrammer:
# Context
First call to https://lumen.akvotest.org/library that internally calls /api/library takes almost 10 times more with auth feature flag activated than without it

# Problem or idea
flow-api takes its time to answer 

# Solution or next step
@dlebrero @kardan any ideas on how to increase the performance of flow-api?

tangrammer:
After a bit of brainstorming ☔️ we agreed that we'll set up a cache in flow-api side, leaving lumen as dumb as it is and we pretend 😄it to be  
also we'll implement some way of tracking/monitoring the new-auth implementation in production without affecting old-auth and performance, that's to say running a second auth call in a
new background thread

tangrammer:
relates to https://github.com/akvo/akvo-flow-api/pull/159

tangrammer:
monitoring working too
`kubectl port-forward lumen-blue-c5956dd74-jh68w 3001:3000`

PR #2093, opened 2019-04-23
Monitoring flow-api/check-permissions call (Connect #2091)

tangrammer:

### DIFF
--- a/backend/src/akvo/lumen/lib/auth.clj
+++ b/backend/src/akvo/lumen/lib/auth.clj
@@ -3,17 +3,19 @@
    [akvo.commons.jwt :as jwt]
    [akvo.lumen.component.flow :as c.flow]
    [akvo.lumen.component.tenant-manager :as tenant-manager]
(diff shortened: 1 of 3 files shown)


backend/src/akvo/lumen/lib/auth.clj
```diff
@@ -133,47 +135,63 @@
                                   :visualisation-ids auth-visualisations
                                   :dashboard-ids auth-dashboards})))
 
+(defn- flow-check-permissions [flow-api request collector tenant data]
+  (let [t (prometheus/start-timer collector :app/flow-check-permissions {"tenant" tenant})
```
dlebrero: Use https://xsc.github.io/iapetos/iapetos.core.html#var-with-duration

See https://github.com/akvo/akvo-unified-log/blob/3f12df3f01e85efa348ed5643d65b590e6c76526/src/akvo_unified_log/db.clj#L15 is you want to get fancy.

Issue #1740, opened 2018-10-30
Visualisation get split for dashboard pdf exports 


janagombitova:
# Context
*Why do we add this issue?*

We added the option to allow exporting of dashboards and visualisations for Salim to use in reports or presentations. We aim to make these exports as useful and valueable as possible.

# Problem or idea
*What is the cause?*

In dashboards we do not have the concept of pages and how the visualisations are disctributed across the canvas is flexible. When exporting a dashboard as a pdf, the dashboard gets split into pages. This is ok. 

The problem occurs when a visualisation or a text on that dashboard is split between the pages. 

### Example1
<img width="965" alt="screen shot 2018-10-30 at 14 17 50" src="https://user-images.githubusercontent.com/12456965/47720648-ad6d9580-dc4e-11e8-8ff3-95fcee5ffd08.png">

### Example2
<img width="947" alt="screen shot 2018-10-30 at 14 19 24" src="https://user-images.githubusercontent.com/12456965/47720779-faea0280-dc4e-11e8-9451-e325bfa9fe68.png">

### Example3
<img width="958" alt="screen shot 2018-10-30 at 14 19 34" src="https://user-images.githubusercontent.com/12456965/47720780-fb829900-dc4e-11e8-978b-5af198f9bc94.png">

Here are a few examples (as not all data is public the examples are in an internal folder): https://drive.google.com/drive/u/1/folders/1z6LByKFgbagp0BT5IFc6v-9Tnc9PJcMn

# Solution or next step
*How could we solve it?*

Would it be possible to ensure that what does not fully fit on one page gets moved to the next page for pdf exports?

janagombitova:
The splitting works really well! 👍 

After a quick chat with @finnfiddle we agreed to remove the black borders from the blocks to keep the exports clean and simple. Finn's choice on removing the grey background to avoid extra ink when printing is a good one but adding the borders to ensure you can see where the blocks are feels unnecessary. So we will go with all while and see what feedback we will receive.

janagombitova:
Out in the wild and looking good.

PR #2095, opened 2019-04-24
Visualisation get split for dashboard pdf exports #1740

finnfiddle:
Ok so this turned out to be quite a tricky little task - basically a packing algorithm to pack items onto pages when exporting pdf - and moving items around creates cascading changes to subsequent pages. I wrote tests so I'm confident everything is working. Here is a pdf created with the new algorithm

[Map Dashboard (29).pdf](https://github.com/akvo/akvo-lumen/files/3111523/Map.Dashboard.29.pdf)


finnfiddle:
also its not very optimized I think but it only runs serverside so I dont think its an issue at this stage

Issue #2071, opened 2019-04-10
Category Derived columns


janagombitova:
# Context

**Why do we add this issue?** 

When researching the value of connecting Flow and Lumen, one of the questions we focused on understanding was Why use Lumen in the 1st place (and why do you not use Lumen)? In our research, we learned that Lumen's value is in making it simple to handle data. For complex tasks, you need to have a certain level of data skills and in that case, you usually have the tools you use already. But the use of data is spreading across development projects with the skills set of project officers not catching up yet. Plus working with data is an extra task on top of already a long to do list. So Lumen makes handling data easy to do and thus also saves time. 

In our research, we however also learned that our current derived column implementation does not meet the criteria of simplicity and some commonly needed and used data tasks are not yet supported with the other transformation options.  

# Problem or idea

**What is the status quo?**
Data comes in and often when visualising it Salim sees that there are too many variables making the visualisation hard to read. For example, when testing WQ the result numbers can vary but often ranges of these results are enough to indicate if the water is safe or not. Or another example would crop farmers plant. Out of the 10 different crops answered only 4 are most used and the other 6 can be grouped into one category called other and naming the answers, but keeping the visualisation with only 4 values. 

In order to make such changes to simplify the dataset and in consequence the visualisations (making them easier to read and understand) today Salim has to be familiar with javascript to use Lumen's derived column functionality. Salim, however, is not technical and does not know javascript making this either a task Akvo does for him what results in Salim not using Lumen or in Salim not using Lumen at all. 

**If this change is successful what might we observe?** 
* More partner users picking up using Lumen
* Less readership and requests about javascript Derived column
* Decrease in use of Derived column and increase in the new transformation
* Increase in visualisations 
* Increase in Lumen sales 

# Solution or next step

**How could we solve it?**
The idea is to make it easy to define new categories for existing values per column. 

**Who's behaviour will change and How?** 
*Salim*
We assume he will use Lumen more, spend more time with the dataset and transform it. We also assume he will create more visualisations (need for sharing best practices can be handy). 

We also assume we get more questions on where to do data cleaning, as the categorical column can be used for cluster-edit cases. What on one hand will prove the value of this change, but also the need to resolve the data questions across F and L. 

*Petra*
We assume Petra uses more complex formulas to track her M&E indicators thus the assumption is that this change will not be enough for her. We assume to hear more requests for expanding the functionality to fit her needs. Thus the categorical column transformation will be seen as useful, but not enough.

[...]

PR #2134, opened 2019-05-13
Add core/derive-category transformation backend op

tangrammer:
Relates to #2071 

### expected payload 
```
{
    op: 'core/derive-category',
    args : {
    source: {column: {columnName: 'c1'}},
    target: {column: {title: 'Color'}},
    derivation: {mappings: [
                           [['source-val-1'],'derived-val-1'],
                           [['source-val-2', 'source-val-3'], 'derived-val-2'],
                           [['source-val-4', 'source-val-5', 'source-val-6'],'derived-val-3']],
                 uncategorizedValue: "Uncategorised"
      }

    }
}
```

- [x] we should also add the option to search for values in the list and when creating a group  
- [x] [tracking](https://github.com/akvo/akvo-lumen/pull/2136)

### DIFF
--- a/backend/specs/akvo/lumen/specs/transformation.clj
+++ b/backend/specs/akvo/lumen/specs/transformation.clj
@@ -40,6 +40,7 @@
 				    "core/filter-column"
 				    "core/generate-geopoints"
 				    "core/merge-datasets"
+                                    "core/derive-category"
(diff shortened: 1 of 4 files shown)


tangrammer:
@finnfiddle you can use this branch to develop your client-side js code

PR #2222, opened 2019-06-19
Use Auth0 or Keycloak as authentication system

tangrammer:
- [ ] **Update release notes if necessary**

This PR offers auth0 behind a feature flag in client side. 

- [x] New ENV var `LUMEN_AUTH0_URL`  and `LUMEN_AUTH0_PUBLIC_CLIENT_ID`
- [x] client feature flag to select auth provider http://t1.lumen.local:3030?auth=auth0 or `http://t1.lumen.local:3030?auth=keycloak` ... the absense of value will select `keycloak` by default
- [ ] TODO: [refresh-token](https://github.com/akvo/akvo-lumen/pull/2222#issuecomment-505924175) auth0 technique https://github.com/akvo/akvo-lumen/issues/2228
- [x] Double check client side auth [uses](https://github.com/akvo/akvo-lumen/pull/2222#issuecomment-504454320)
- [x] Fix profile language locales
- [x] Fix profile user name
- [x] Log out `auth0Client.logout(...p) ` from [tutorial](https://auth0.com/docs/quickstart/spa/react)
- [x] Add google cloud auth0 credentials
- [x] use akvotest.eu.auth0.com instead of tangrammer.eu.auth0.com identity
- [x] feed all auth values with ENV data
- [x]  config Raven production  
- [ ] refactor and remove console.logs in client side

PR #2227, opened 2019-06-20
Issue/2212 split auth

kardan:
Since I'm off for midsummer I'll open a PR for what I had time to complete. 

### Client side errors screen
I have not had time to update the client to show different error messages if Keycloak times out. A user will get the "bad permissions" screen as before if Keycloak times out. However, the backend is responding with a 503 so I guess we can extend the client "fetchLibraryFailure".

### Feature flag
To enable the API authz add a trailing $auth to the family name.


kardan:
On why pulling in the big dependency of Aleph. Since we wanted a pool and to do it async and using the clj-http we had just throw null pointer error even for the simples readme example. So to not spend time on it I just pulled in another lib.

backend/src/akvo/lumen/auth.clj
```diff
@@ -81,3 +99,41 @@
...
+   (contains? allowed-paths tenant)))
+
+(defmethod ig/init-key :akvo.lumen.auth/wrap-authorization
+  [_ {:keys [keycloak]}]
+  (fn [handler]
+    (fn [{:keys [jwt-claims tenant] :as request}]
+      (if (api-authz? jwt-claims)
+        (try
+          (let [email (get jwt-claims "email")
+                allowed-paths (keycloak/allowed-paths keycloak email)]
```
dlebrero: allowed-paths should not run in the  `(nil? jwt-claims)` and  `:else` cases as it is making at least one http request to KeyCloak.

Either wrap it on a `delay` or move it inside the  `(if (api-tenant-admin? tenant allowed-paths)` branches

backend/src/akvo/lumen/auth.clj
```diff
@@ -81,3 +99,41 @@
...
+              (nil? jwt-claims) not-authenticated
+              (admin-path? request) (if (api-tenant-admin? tenant allowed-paths)
+                                      (handler request)
+                                      not-authorized)
+              (api-path? request) (if (api-tenant-member? tenant allowed-paths)
+                                    (handler request)
+                                    not-authorized)
+              :else not-authorized))
+          (catch Exception e
+            (log/info (.getMessage e))
```
dlebrero: Unlikely that we want to log the 401

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -1,37 +1,75 @@
...
-  [issuer]
-  (-> (client/get (format "%s/.well-known/openid-configuration" issuer))
-      :body json/decode))
+  ([issuer]
+   (let [pool (http/connection-pool {:connection-options {:raw-stream? true}})]
+     (fetch-openid-configuration issuer pool)))
+  ([issuer pool]
+   (-> @(http/get (format "%s/.well-known/openid-configuration" issuer)
+                  {:pool pool})
+       :body
```
dlebrero: Why not add a timeout here also?

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -220,11 +258,79 @@
           (println (format "Tried to remove user: %s" user-id))
           (lib/internal-server-error))))))
 
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; API Authorization
+;;;
+
+(defn- api-get-async [{:keys [timeout] :as req-opts} url]
```
dlebrero: We already have this logic in fetch-openid-configuration and request-headers. Maybe we can move it to some helper fn?

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -220,11 +258,79 @@
...
+  ([keycloak email]
+   (allowed-paths keycloak email {}))
+  ([{:keys [api-root http-pool user-id-cache] :as keycloak} email
+    {:keys [timeout] :or {timeout 10000}}]
+   (let [bare-req-opts {:pool http-pool
+                        :timeout timeout}]
+     (if-some [headers (request-headers keycloak bare-req-opts)]
+       (let [req-opts (assoc bare-req-opts :headers headers)]
+         (if-let [user-id (lookup-user-id req-opts api-root user-id-cache email)]
+           (if-let [allowed-paths (api-get-async req-opts
```
dlebrero: This exception handling is copied all over the place. Apart of the duplication, it makes the code harder to follow with so many nested ifs. This code is already non-pure and uses exceptions for flow control (for example `lookup-user-id` throws exceptions and it has no else branch in this logic), so there is really no point on returning nils and handling them

Move to helper function suggested at changes in `api-get-async`

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -270,12 +376,19 @@
...
 
-(defn- keycloak [{:keys [credentials url realm]}]
-  (map->KeycloakAgent {:issuer (format "%s/realms/%s" url realm)
-                       :api-root (format "%s/admin/realms/%s" url realm)
-                       :credentials credentials}))
+(defn- keycloak [{:keys [credentials http-pool url realm]}]
+  (map->KeycloakAgent {:api-root (format "%s/admin/realms/%s" url realm)
+                       :credentials credentials
+                       :issuer (format "%s/realms/%s" url realm)
+                       :user-id-cache (atom {})
```
dlebrero: Using an atom (instead of a size-limited cache) because we do not expect too many users?

backend/src/akvo/lumen/auth.clj
```diff
@@ -37,22 +37,40 @@
...
+
+(def internal-server-error
+  (-> (response/response "Internal server errror")
+      (response/status 500)))
+
+(defn api-authz?
+  "Feature flag predicate function. Add trailing $auth to given_name to enable
+  API authz"
+  [{:strs [family_name] :as jwt-claims}]
+  (if (nil? family_name)
```
dlebrero: maybe use `some->` to avoid the if condition?

tangrammer: `and` returns the last true value
`(and (nil? family_name) (string/ends-with? family_name "$auth"))`

example `(and 1 2)` => 2

backend/src/akvo/lumen/auth.clj
```diff
@@ -81,3 +99,41 @@
...
+  (contains? allowed-paths (str tenant "/admin")))
+
+(defn api-tenant-member?
+  [tenant allowed-paths]
+  (or
+   (contains? allowed-paths (str tenant "/admin"))
+   (contains? allowed-paths tenant)))
+
+(defmethod ig/init-key :akvo.lumen.auth/wrap-authorization
+  [_ {:keys [keycloak]}]
```
tangrammer: I'd rename `keycloak` to `authorizer`

backend/src/akvo/lumen/protocols.clj
```diff
@@ -46,6 +46,11 @@
     [this tenant-label]
     "List tenants users"))
 
+(defprotocol KeycloakAuthorization
```
tangrammer: I'd rename the protocol name to something like `AuthoriserService` ... so don't include `Keycloak` inside the name

backend/src/akvo/lumen/component/keycloak.clj
tangrammer: this is called only on system/start, 
currently for each entrypoint we defined in config.edn, so sometimes it takes a few time to make the same call several times, so i wrote it in a way it's called only once ... https://github.com/akvo/akvo-lumen/pull/2222/commits/638f0b678ebb31dd2b28239b0561a819f885e7c6

kardan:
While incorporating the feedback I'm seeing a "java.lang.RuntimeException: java.lang.IllegalStateException: already shutdown" from the timeout test. Investigating that..

backend/src/akvo/lumen/component/keycloak.clj
kardan: Without the openid-configuration we can't really operate and should exit the process (log fatal?) and let K8s do a restart, or way better have retry logic and in worst case exit.

dlebrero: Let the process die and k8s restart it until it works. We should already get an alert if this happens.

backend/src/akvo/lumen/component/keycloak.clj
kardan: Agree on the lack of beauty on this code, this should be reworked.

backend/src/akvo/lumen/component/keycloak.clj
kardan: Should change

backend/src/akvo/lumen/auth.clj
```diff
@@ -129,13 +129,13 @@
       (if (api-authz? jwt-claims)
         (try
           (let [email (get jwt-claims "email")
-                allowed-paths (delay (keycloak/allowed-paths keycloak email))]
+                allowed-paths (keycloak/allowed-paths keycloak email)]
```
dlebrero: We want a `delay`. See https://github.com/akvo/akvo-lumen/pull/2227/commits/3b2fc2e59d2993215679fa711cde601d6f440d14#r296692858

backend/src/akvo/lumen/component/keycloak.clj
tangrammer: tested that if we don't have a 20x status response then we get a proper ExceptionInfo :) 

```
ExceptionInfo clj-http: status 404 {:cached nil, :request-time 55, :repeatable? false, :protocol-version {:name "HTTP", :major 1, :minor 1}, :streaming? false, :http-client #object[org.apache.http.impl.client.InternalHttpClient 0x3ae6f731 "org.apache.http.impl.client.InternalHttpClient@3ae6f731"], :chunked? false, :type :clj-http.client/unexceptional-status, :reason-phrase "Not Found", :headers {"Connection" "keep-alive", "Content-Length" "0", "Date" "Wed, 17 Jul 2019 10:19:29 GMT"}, :orig-content-encoding nil, :status 404, :length 0, :body "", :trace-redirects []}

```

backend/src/akvo/lumen/component/keycloak.clj
tangrammer: should be solved by https://github.com/akvo/akvo-lumen/pull/2227/commits/b1910251dd800db281fc98b27594b2640e1e6d41

backend/src/akvo/lumen/component/keycloak.clj
tangrammer: should be solved by https://github.com/akvo/akvo-lumen/pull/2227/commits/1c45ebaaae64f1d83f8d65f9a6114c41c1cb2bd1

tangrammer:
@dlebrero IMO the PR is ready for a second review 🦆

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -410,14 +393,22 @@
   ::data)
 
 (defmethod ig/init-key :akvo.lumen.component.keycloak/keycloak  [_ {:keys [credentials data] :as opts}]
+  (log/info "Starting keycloak")
   (let [{:keys [issuer openid-config api-root] :as this} (keycloak (assoc data :credentials credentials))
         http-timeout 10000
-        openid-config (fetch-openid-configuration issuer {:timeout http-timeout})]
+        cm (http.conn-mgr/make-reusable-conn-manager {:timeout 2 :threads 3 :insecure? false :default-per-route 10})
```
dlebrero: `timeout` is actually the TTL of the connection. We can be more generous here. I would suggest 20 or 30 secs.

dlebrero: `threads` is actually the max number of connections for all routes, while `default-per-route` is the maximum connections for a given route. Note that route here means host, which means the keycloak server.

Given that the pool is going to be used for one route (the keycloak host), `thread` and `default-per-route` should be the same. 10 seems like a sensible value.

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -410,14 +393,22 @@
...
+        openid-config (fetch-openid-configuration issuer {:timeout http-timeout :connection-manager cm})]
     (log/info "Successfully got openid-config from provider.")
     (assoc this
+           :cm cm
```
dlebrero: Avoid abbreviations. connection-manager, http-connection-manager or http-connection-pool are more obvious names

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -51,11 +51,7 @@
                                 json/decode
                                 (get "access_token"))]
        {"Authorization" (str "bearer " access-token)
-        "Conternt-Type" "application/json"}
-       (throw
-        (ex-info "Down stream auth server timed out (Keycloak)"
-                 {:response-code 503}))))))
-
+        "Conternt-Type" "application/json"}))))
```
dlebrero: Content is mistyped

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -221,11 +239,63 @@
...
+    (swap! user-id-cache cache/hit email)
+    user-id))
+
+(defn- lookup-user-id
+  "Lookup email -> Keycloak user-id, via cached Keycloak API."
+  [cm req-opts api-root user-id-cache email]
+  (if-some [user-id (get-user-id user-id-cache email)]
+    user-id
+    (let [url (format "%s/users/?email=%s" api-root email)]
+      (let [users (-> (client/get url (assoc req-opts :connection-manager cm)) :body json/decode)]
```
dlebrero: Merge `let` with previous one

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -221,11 +239,63 @@
...
+  remove the leading /akvo/lumen and return paths as demo/admin."
+  ([keycloak email]
+   (allowed-paths keycloak email {}))
+  ([{:keys [api-root http-timeout user-id-cache cm] :as keycloak} email
+    {:keys [timeout] :or {timeout http-timeout}}]
+   (let [bare-req-opts {:timeout timeout}]
+     (if-some [headers (request-headers keycloak (assoc bare-req-opts :timeout timeout))]
+       (let [req-opts (assoc bare-req-opts :headers headers)]
+         (when-let [user-id (lookup-user-id cm req-opts api-root user-id-cache email)]
+           (when-let [allowed-paths
```
dlebrero: when-let, when-let, if-some, when-let. I wonder if we can simplify this with some macro. I kind of remember somebody publishing some library to avoid this kind of nesting.

backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -303,13 +283,16 @@
   remove the leading /akvo/lumen and return paths as demo/admin."
   ([keycloak email]
    (allowed-paths keycloak email {}))
-  ([{:keys [api-root http-timeout user-id-cache] :as keycloak} email
+  ([{:keys [api-root http-timeout user-id-cache cm] :as keycloak} email
     {:keys [timeout] :or {timeout http-timeout}}]
+   (log/info :allowed-paths-call)
    (let [bare-req-opts {:timeout timeout}]
```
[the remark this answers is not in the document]
tangrammer: using timeout and connection-timeout now... also defining a new function 
```clojure
(defn default-req-opts
  "all timeouts are in milliseconds!!
  :timeout - Time that connections are left open before automatically closing
  :connection-timeout - the default connect timeout value for connection

  https://github.com/dakrone/clj-http/issues/477
  " 
  [opts]
  (merge {:timeout 5 :threads 2 :insecure? false :default-per-route 2}
         (select-keys opts [:timeout :threads :insecure? :default-per-route])))

```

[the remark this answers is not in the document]
tangrammer: solved by https://github.com/akvo/akvo-lumen/pull/2227/commits/9b11681a55f0b73339456563a1940e95caffdcfc

backend/src/akvo/lumen/admin/add_tenant.clj
```diff
@@ -26,11 +26,11 @@
   (:require [akvo.lumen.admin.util :as util]
             [akvo.lumen.component.keycloak :as keycloak]
             [akvo.lumen.config :refer [error-msg]]
+            [akvo.lumen.http :as http]
```
dlebrero: maybe http-client is a better name?

backend/src/akvo/lumen/http.clj
```diff
@@ -0,0 +1,38 @@
...
+(defn shutdown-manager [connection-manager]
+    (http.conn-mgr/shutdown-manager connection-manager))
+
+(def default-req-opts
+  ":connection-timeout - time to establish the socket
+   :connection-request-timeout - time to have a free socket
+   :socket-timeout - time to get the response" 
+  {:connection-timeout 2000 :connection-request-timeout 50 :socket-timeout 5000})
+
+(defn get* [url & [opts]]
```
dlebrero: why `get*` and not `get`? And the same for others

backend/src/akvo/lumen/admin/add_tenant.clj
tangrammer: the alias?

tangrammer: or the namespace?

backend/src/akvo/lumen/http.clj
tangrammer: i don't want to exclude core/get so i have to remember to use core/get each time i need it inside the namespace

backend/src/akvo/lumen/admin/add_tenant.clj
tangrammer: http namespace is wrapping http-connection-manager things and http-client things ... so i think keeping generic is better

tangrammer: I didn't realize about `server` world ... I'll move the ns to http.client

tangrammer: solved by https://github.com/akvo/akvo-lumen/pull/2227/commits/781c970b52e8282a5abff5c79ce677708215cab5

Issue #2218, opened 2019-06-17
Remove Close button from Group in Category column


janagombitova:
# Context
*Why do we add this issue?*

To make the use of transformations as easy as possible and as understandable as possible. 

# Problem or idea
*What is the cause?* 

When working with Category columns for TEXT data you have the option to group values. You can select all, or select individual items and once done with defining the group there are two UI elements to close the window: an **x** and **close button**. 

The **x** works well and is clear in the UI. The **Close button** does not work - nothing happens when you click on it and is confusing. Some users who have tested out this feature have asked me what the button is supposed to do as it does not do anything and you can close the window with the **x**. 

![Screen Shot 2019-06-17 at 15.55.36.png](https://images.zenhubusercontent.com/5c8f6c3ea7f08c1a3af97cc1/12b27742-e6fd-4083-a95f-469f8526a25e)

# Solution or next step
*How could we solve it?*

Can we remove the **close button** to remove the confusion?

janagombitova:
Released. Thank you @tangrammer

PR #2235, opened 2019-07-15
[#2218] Call only once http open-id certs for each auth system

tangrammer:

Currently we are calling open-id certs for every entrypoint we declare in config.edn, 

working with current keycloak `dev` environment doesn't affect performance thus the http calls are done in local docker, but switching to other not-local auth system  (like auth0) will slow the system startup in `dev` too

### DIFF
--- a/backend/src/akvo/lumen/auth.clj
+++ b/backend/src/akvo/lumen/auth.clj
@@ -8,7 +8,9 @@
             [clojure.spec.alpha :as s]
             [akvo.lumen.component.keycloak :as keycloak]
             [integrant.core :as ig]
-            [ring.util.response :as response]))
+            [ring.util.response :as response])
+  (:import java.text.ParseException
+           com.nimbusds.jose.crypto.RSASSAVerifier))
 
 (defn claimed-roles [jwt-claims]
   (set (get-in jwt-claims ["realm_access" "roles"])))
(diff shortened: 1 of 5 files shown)


backend/src/akvo/lumen/auth.clj
```diff
@@ -57,17 +59,18 @@
...
-     (let [issuer (str url "/realms/" realm)
-           certs  (-> (str issuer "/protocol/openid-connect/certs")
-                      client/get
-                      :body)]
-       (jwt/wrap-jwt-claims handler (jwt/rsa-key certs 0) issuer))
-     (catch Exception e
-       (println "Could not get cert from Keycloak")
-       (throw e)))))
+    (let [keycloak-verifier (RSASSAVerifier. (:rsa-key keycloak))]
+    (fn [req]
```
dlebrero: indenting looks wrong in GitHub.

backend/src/akvo/lumen/auth.clj
```diff
@@ -57,17 +59,18 @@
...
-   (try
-     (let [issuer (str url "/realms/" realm)
-           certs  (-> (str issuer "/protocol/openid-connect/certs")
-                      client/get
-                      :body)]
-       (jwt/wrap-jwt-claims handler (jwt/rsa-key certs 0) issuer))
-     (catch Exception e
-       (println "Could not get cert from Keycloak")
-       (throw e)))))
+    (let [keycloak-verifier (RSASSAVerifier. (:rsa-key keycloak))]
```
dlebrero: This reminds me that in theory, for additional security, the RSA key from the OAuth server should rotate periodically. Not sure why we did not enable that feature in Keycloak, but the code as it is/was will not support the RSA key rotation.

It is ok for this pull request but please keep it in mind now that you are our security expert.

tangrammer: I'll keep it in mind 👍 !

PR #2260, opened 2019-07-26
Sign up with auth0

tangrammer:

After agreeing with Jana about sending an email with auth0 to add the specific email account to use, this PR adds specific email templates for each auth system option (keycloak and auth0) 

About keeping the verification link in auth0 version is just to keep compatibility with keycloak, that's to say that we could remove it once keycloak is removed. So far it's needed  thus keycloak users verification-email is supported by lumen and authorisation is based on keycloak users and verification-email by lumen

### DIFF
--- a/backend/resources/akvo/lumen/email/auth0/create_new_account_and_invite_to_tenant.txt
+++ b/backend/resources/akvo/lumen/email/auth0/create_new_account_and_invite_to_tenant.txt
@@ -1,0 +1,12 @@
+Hi!
+
+You have been invited to join Akvo Lumen by {{author-email}}.
+To complete the invitation process please visit:
+{{location}}/verify/{{invite-id}}?auth=auth0
+
+
+Then, using your email: {{email}} you need to click on "Sign up" or, in case that your email is managed by Google, you could also use the "Log in with Google" button
+
+Thank you
+
+The Akvo Lumen Team
--- a/backend/resources/akvo/lumen/email/auth0/invite_to_tenant.txt
+++ b/backend/resources/akvo/lumen/email/auth0/invite_to_tenant.txt
@@ -1,0 +1,11 @@
+Hi!
+
+You have been invited to join Akvo Lumen by {{author-email}}.
+To complete the invitation process please visit:
+{{location}}/verify/{{invite-id}}?auth=auth0
+
+Then, using your email: {{email}} you need to click on "Sign up" or, in case that your email is managed by Google, you could also use the "Log in with Google" button
+
+Thank you
+
+The Akvo Lumen Team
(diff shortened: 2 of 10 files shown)


tangrammer:
@dlebrero assigned the review to you thus I'd like to ensure that you agree about the PR [description](https://github.com/akvo/akvo-lumen/pull/2260#issue-301444477)

dlebrero:
So once we move to Auth0, the flow will be:

1. User receives email from Lumen to verify his email in Lumen.
1. User clicks the link in the email and lands in Lumen. His email is verified in Lumen.
2. User goes to Lumen and is redirected to Auth0
3. User sign up to Auth0
4. User is redirected to Lumen, but an error is shown because he did not verify his email with Auth0
5. User check his email again and verifies the email against Auth0
6. User goes to Lumen 
7. User is redirected to Auth0 to login
8. User log into Auth0 and it is redirection to Lumen.

Is this the expected flow?

tangrammer:
@dlebrero exactly!  that flow is the current one (with this PR)

janagombitova:
It sounds like a lot of steps, but yes. That is what we agreed on with Juan for now as:  
- handling the emails is already in Lumen anyway so we do not need to explain a change in workflow to existing users as things remain the same for adding new users 
- with sending out the email invitation we remove the need of the admin telling the new user where to go to and which email to use --> we can learn if this workflow works better in comparison to the one in Flow 
- the 1st email sent will be 'informative'. It will not hold an actionable link, nor there will be a time by which the new user needs to do something. The 2nd email, to verify the Auth0 email, will be the actionable one.

tangrammer:
@dlebrero once that we move to auth0:

1. User receives email from Lumen to know which email has been invited to Lumen, and who has invited to this new-user
2. User clicks the link in the email and lands in Lumen and is redirected to Auth0
3. User sign up to Auth0
4- User is redirected to Lumen, but an error is shown because he did not verify his email with Auth0
5. User check his email again and verifies the email against Auth0
6. User goes to Lumen
7. User is redirected to Auth0 to login
8. User log into Auth0 and it is redirection to Lumen.

tangrammer:
[the flow commented](https://github.com/akvo/akvo-lumen/pull/2260#issuecomment-515399142) by @dlebrero here is just until we removed keycloak (as authentication system)

janagombitova:
Just as a side note, Kiarii added the mock-ups for how to the workflow (without Keycloak, only Auth0) should behave here: https://github.com/akvo/akvo-platform/issues/154#issuecomment-515393651

dlebrero:
The user flow is ok as all the redirects will be automatic. I was imagining that the user had to click around a lot more than before.

Issue #2289, opened 2019-08-29
Fix auth0 user menu don't close properly


kardan:
Clicking the user button does not close an open menu, only clicking outside of the menu will.

PR #2290, opened 2019-08-29
[#2289] Make the user menu close on button click.

kardan:

### DIFF
--- a/client/src/components/workspace-nav/user-menu/UserMenu.jsx
+++ b/client/src/components/workspace-nav/user-menu/UserMenu.jsx
@@ -33,7 +33,7 @@ class UserMenu extends React.Component {
         <Manager>
           <Reference>
             {({ ref }) => (
-              <span ref={this.menuRef}>
+              <span ref={this.buttonRef}>
                 <button type="button" ref={ref} onClick={() => this.toggle()}>
                   <i className="fa fa-user-o userIcon" aria-hidden="true" />
                   {profile.username}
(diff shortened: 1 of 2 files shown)


client/src/components/workspace-nav/user-menu/UserMenu.jsx
```diff
@@ -9,6 +9,7 @@ class UserMenu extends React.Component {
   constructor(props) {
     super(props);
     this.state = { isExpanded: false };
+    this.buttonRef = React.createRef();
```
tangrammer: Didn't see that `React.createRef` call before, could you expand the reasons to use it? I'm really curious

kardan: It's because of the interaction between clicks on the button and "outside" the menu. We want to add an event listener for clicks outside of the menu and have it close. But since the button lives outside the menu it becomes "double" events being triggered. Hence we check to see if the click was outside of menu or the menu open button. https://github.com/akvo/akvo-lumen/pull/2290/files#diff-7f3fd3cbd19adb68c55c21e483c6b892R24.

tangrammer: it sounds like the same use case that we have in [other places](https://github.com/akvo/akvo-lumen/search?q=ClickAway&unscoped_q=ClickAway)...  are you familiar with current component [ClickAway](https://github.com/akvo/akvo-lumen/blob/e54fccdfc12f86225c2581cf11b13488f938f02b/client/src/components/common/ClickAway.jsx) ?

kardan: Had not seen the ClickAway but it looks like half of the issue with just clicking "somewhere" outside of the current DOM branch. I tried to cover current DOM branch and also a sister branch (the button).

tangrammer: ok!

PR #2313, opened 2019-09-11
Issue/2309 monitor authz api

kardan:
Added a new env var "API_AUTHZ_PROBABILITY" to be able to steer how much load we put on the authz API before we switch over to API onyl. Used configuration and not code to not have to deploy a new version. Config should be a flow between 0.0 and 1.0 where 0.0 is no probability and 1.0 is always.


backend/src/akvo/lumen/auth/utils.clj
```diff
@@ -0,0 +1,71 @@
...
+    :other))
+
+(defmulti api-authz?
+  (fn [{:strs [iss] :as jwt-claims}]
+    (cond
+      (string/ends-with? iss "eu.auth0.com/") :auth0
+      (or (= iss "https://kc.akvotest.org/auth/realms/akvo")
+          (= iss "https://login.akvo.org/auth/realms/akvo")) :keycloak)))
+
+#_(defmethod api-authz? :auth0
```
tangrammer: should be removed?

kardan:
@tangrammer realise things were a bit different between authentication and authorization. Hence, tired to stream line that to look more similar.

PR #2371, opened 2019-10-03
Allow numeric first character as tenant label [admin]

tangrammer:


kardan:
Fix fine, but test that checks for what we want to change needs to be removed.

Issue #2384, opened 2019-10-10
java.net.SocketTimeoutException: Read timed out


tangrammer:
https://sentry.io/organizations/akvo-foundation/issues/1243868051/?referrer=github_plugin

```
java.net.SocketTimeoutException: Read timed out
  File "client.clj", line 37, in akvo.lumen.http.client/get-STAR-
    (client/get url opts))
  File "client.clj", line 34, in akvo.lumen.http.client/get-STAR-
    (defn get* [url opts]
  File "flow_common.clj", line 22, in akvo.lumen.lib.import.flow-common/form-instances-STAR-
    (http.client/get* (merge (http.client/req-opts 20000)
  File "flow_common.clj", line 20, in akvo.lumen.lib.import.flow-common/form-instances-STAR-
    (defn form-instances* [headers-fn url]
  File "flow_common.clj", line 34, in akvo.lumen.lib.import.flow-common/form-instances
    (form-instances* headers-fn initial-url)))
...
(71 additional frame(s) were not displayed)

Read timed out
```

iperdomo:
After some investigation with @kardan we found the problem.

Currently we have a read timeout of **20 seconds** in the Lumen code, while trying to access Flow API.

Checking the logs on the Flow side we found the following:

![2019-10-10_676x150_scrot](https://user-images.githubusercontent.com/178474/66567663-ff5db180-eb67-11e9-8e20-4cf5fac97bda.png)

The majority of call are lower than 300ms but there was one that took **19.4 seconds** to finish. Really close to the read timeout on Lumen side.

Expanding this particular call we found GAE auto-scaling functionality was spinning up a new _instance_ to cope with the current load.

> This request caused a new process to be started for your application, and thus caused your application code to be loaded for the first time. This request may thus take longer and use more CPU than a typical request for your application.

Starting a new instance causes this particular call takes _long_ time, which causes a read timeout in the Lumen side.

kardan:
Flow API returns 500 at one page and we have no retry logic for imports.

kardan:
Beside us having retry there was a 500 error on Flow and there is an issue opened there https://github.com/akvo/akvo-flow-api/issues/184

janagombitova:
@kardan Is his issue is now resolved? Can we close it?

PR #2387, opened 2019-10-10
[#2384] Use defaults for Flow form instance import

kardan:


backend/src/akvo/lumen/lib/import/flow_common.clj
```diff
@@ -1,53 +1,66 @@
...
-            [clojure.java.jdbc :as jdbc]
-            [clojure.string :as str])
+  (:require
+   [akvo.commons.psql-util :as pg]
+   [akvo.lumen.http.client :as http.client]
+   [cheshire.core :as json]
+   [clojure.java.jdbc :as jdbc]
+   [clojure.string :as str]
+   [clojure.tools.logging :as log]
+   [diehard.core :as dh])
```
iperdomo: Why this re-indentation is required?

kardan: It's not, it happened while playing with diehard and pulling in a temporary http client.

Issue #2437, opened 2019-11-13
Flow questions can have same value but different group


tangrammer:
# Context
Flow allow the final user to define questions with the same "name" but different groups. So the question identity  depends on the context of the group

# Problem or idea
Lumen doesn't expect to have duplicated column names. One direct consequence is js-transformation, that uses that name to derive new values

# Solution or next step
🤔

tangrammer:
After talking with @Kiarii about how better handle this question duplicated problem, we agree on adding the difference as a suffix in the question title
something like that... "example-question-duplicated-X", "example-question-duplicated-Y"
Now we need to decide the suffix value:
should we use just a numeric index? or should we add the flow group name there?
I've just checked that lumen receives the flow group in the import/update processes so we are able to use it

janagombitova:
My initial hunch was to go with the simple numeric index. But if we have the question group info coming from Flow, it might be more informative to use the group index (G1, or G3) if we have that data.

Kiarii:
I had the feeling that using the group name or number (as a prefix or suffix) for repeating questions only will make it look weird; it was, it was the reason we had settled for numbering for now; so that if there is a question "example-question-duplicated-X" the user can expect a Y or more of the same.. once we visit RQGs this might need to change..

janagombitova:
With Juan we spoke yesterday and agreed to handle this issue to: 
* [ ] add Flow's question groups to the data table as step 1. Not only will this bring the Flow user view of data closer in Lumen, but it also solves the issue of duplicated column headers without expanding the column headers. (the next step would be to see how we can make use of the question groups in question dropdowns across Lumen)
* [ ] when creating a derive column with javascript, warn the user that she selected a column which is duplicated and tell her to refer to the question group too 
* [ ] when updating a Flow dataset with an existing transformation, fail the update because oof a duplicated column derivation and inform in the error message that the update failed because this transformation holds a duplicated column and you need to change it to refer to the group.

tangrammer:
working in akvotest
<img width="1277" alt="Screen Shot 2019-12-16 at 16 12 15" src="https://user-images.githubusercontent.com/731829/70918225-14026e00-201f-11ea-8352-3ddcbea4ef1f.png">

tangrammer:
also working in demo.akvolumen
getting the proper error with js transformations that should refer to the group to avoid data inconsistency references

[...]

Issue #862, opened 2017-06-19
Show more characters to support long COL titles


Kiarii:
# Context
We are at the moment supporting about 15 chars for col titles which makes it difficult to understand the data since a mouseover to reveal truncated text is necessary for most cols. In this example all col titles are truncated :( 
<img width="1265" alt="screen shot 2017-06-19 at 11 46 35" src="https://user-images.githubusercontent.com/12692444/27276814-0e630a84-54e5-11e7-83a6-f834602856aa.png">

# idea
- Reduce col title font size
- Remove bold styling
- Support 2 lines ~ 30 chars before truncating

![image](https://user-images.githubusercontent.com/12692444/27277079-c198d2dc-54e5-11e7-841e-c272e9f1d96c.png)

This would also make it easy for the user to determine whether to rename a col (if need be) when they are able to see much if not all of the title.. other functionalities also require understanding the title so the more its visible, the easier it is for the user to determine their course of action..

Kiarii:
![image](https://user-images.githubusercontent.com/12692444/27277171-1618ebf8-54e6-11e7-8778-b71a96750dde.png)

Kiarii:
In the wake of thw new data type - Geopoint - it may be best to adjust the placement of the type label in the col header to accommodate the longer type, for now:

![image](https://user-images.githubusercontent.com/12692444/28824838-98cd2d5e-76cc-11e7-869e-ca0cab8a330c.png)

janagombitova:
This issue to seems seems to address the problem described here: https://github.com/akvo/akvo-lumen/issues/868. To ensure we handle the problem in one issue I am referencing this one and closing it.

janagombitova:
Closed the wrong issue, so reopening this one. To clarify the task is to:
* [ ] Adjust column header styling 
* [ ] Support 2 lines ~ 30 chars before truncating
* [ ] Change how data type is displayed

Kiarii:
Following the discussion and considering the UX side of things, the last proposal made might improve the scannability and readability of both col **data types** and **names** as illustrated below. This would also mean that the col titles can use a lesser font weght and font and thereby reduce visual distraction and enable long titles to be easily readable..

<img width="1092" alt="Screenshot 2019-09-06 at 12 20 25" src="https://user-images.githubusercontent.com/12692444/64416959-d6e22380-d0a0-11e9-9cd8-83d922af3520.png">

muloem:
@tangrammer is this fully completed?

[...]

PR #2449, opened 2019-11-22
Duplicated questions and long column names

tangrammer:
# WIP

relates #862 #2437 

how about using column groups https://schrodinger.github.io/fixed-data-table-2/example-column-groups.html

then, instead of changing the column name if we have duplicates we could just "adjust" the js "problematic" code

also we could use column-groups just to split current functionality
eg: column title and functions in "first" row, column type in second row 🤔 

Screen Shot 2019-11-22 at 19 44 34


janagombitova:
@loicsans and @Kiarii can you take a look at Juan's implementation and suggestion? Thanks

PR #2452, opened 2019-11-27
Add group column to lumen dataset view (Only flow datasets)

tangrammer:
Relates #2437 

This PR only does what the title states. Following PRs will contain the logic to avoid collisions on duplicated questions names and javascript execution 

Pay attention that we add 2 groups besides the flow questions groups. One called "Survey details" containing : `'identifier', 'instance_id', 'display_name', 'submitter', 'submitted_at', 'surveyal_time', 'device_id'.` The other called 'Transformations' that will contain the columns resulted of a lumen transformation

PS: We need still to agree on having these 2 new groups @janagombitova  @kardan 

Screen Shot 2019-11-28 at 09 11 53

Screen Shot 2019-11-28 at 09 12 03


iperdomo:
Hi,

>  "Survey details"

Why *Survey* if the user imported a *Form* ?

tangrammer:
@iperdomo good point!

janagombitova:
`Transformations` are brilliant. `Form details` can be deceiving as is it the details of the form/questionnaire or of the submissions. We use a lot of confusing language in Flow/Lumen. But in Flow data exports we call these first columns 'Metadata' as it is the metadata about the submissions. So let's use that for consistency.

kardan:
As "Form metadata" ?

janagombitova:
No just `Metadata`

janagombitova:
We can always reiterate later

Issue #2451, opened 2019-11-27
Make use of question groups in question dropdowns


janagombitova:
In issue #2437 we added Flow question groups to the Lumen data view. This means, that Salim sees how the columns are grouped. The implementation is basic to solve the mentioned issue of handling duplicated column header in derive column transformation. 

This issue wants to see how we can take the information on question groups further. 

Across Lumen we have places where Salim works with questions. He selects them in transformations, in visualisations, in dashboard filters. We know Flow datasets can have hundreds of columns so working with the question/column dropdowns can be painful. The question/column groups help to break down the dataset structure and keep things organised. 

Now that Lumen has the data, can we make more use of them? Can they be implemented in the dropdowns somehow?

tangrammer:
https://github.com/akvo/akvo-lumen/pull/2454

PR #2454, opened 2019-11-29
Make use of question groups in question dropdowns

tangrammer:
Relates #2451

# TODO / WIP
- [x] **adapt cypress tests**
- [x] **Adapt Select menu to other type of entities (so far only column entities works as expected)**
- [x] **Review CSS: I've removed a couple of hard coded css references and metric columns on bar chart seems to have a problem**
- [x] Review the way that merged dataset columns appear in current dataset
- [x] Improve category css in select

# changes landscape
- updated react-select version from "^1.0.0-rc.2" to "3.08"
- created a utilities/column.js namespace to include the column (and group column) utils
- move utilities/utils/filterColumns function to utilities/column.js
- adapt select options to the new group behavior


client/src/components/visualisation/configMenu/BarConfigMenu.jsx
```diff
@@ -160,6 +161,8 @@ class BarConfigMenu extends Component {
           columnOptions.find(item => item.value === value).title : null,
       }, spec, onChangeSpec, columnOptions)}
     />) : '';
+    const columnsBucketColumn = Immutable.fromJS(filterColumns(columnOptions, ['number', 'text']));
```
kardan: These are an array of columns, should be plural (columnBucketColumns)?

tangrammer: I tried to mean that `columnsBucketColumn ` represents the "columns of bucketColumn"

client/src/components/common/SelectMenu.scss
```diff
@@ -16,13 +16,22 @@
     .Select, .Select-control, .Select-input {
         height: 100%;
     }
-
+    .css-18ng2q5-group {
```
kardan: I'm worried about these styles. Are we sure those names are not might be renamed at some point? An alternative is to pass in styles to the react-select component.

Screenshot 2019-12-05 at 10 09 02

Issue #2476, opened 2019-12-17
Use Authz service instead of Flow API


dlebrero:
For Lumen to follow Flow permissions, Flow API is too slow.

Use the new Authz service and check what its performance is

PR #2477, opened 2019-12-17
Use Authz service instead of Flow API

dlebrero:
Created new component and moved the authz specific logic to it

We will still leave the authorisation code with the existing feature flags to check the performance in production

See [#2476]

### DIFF
--- a/backend/src/akvo/lumen/lib/auth.clj
+++ b/backend/src/akvo/lumen/lib/auth.clj
@@ -1,7 +1,7 @@
 (ns akvo.lumen.lib.auth
   (:require
    [akvo.commons.jwt :as jwt]
-   [akvo.lumen.component.flow :as c.flow]
+   [akvo.lumen.component.authz-service :as c.authz]
    [akvo.lumen.component.tenant-manager :as tenant-manager]
    [akvo.lumen.db.collection :as db.collection]
    [akvo.lumen.db.dashboard :as db.dashboard]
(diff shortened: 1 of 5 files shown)


backend/src/akvo/lumen/lib/auth.clj
```diff
@@ -213,16 +217,16 @@
           (handler (assoc request
                           :auth-service (new-auth-service auth-uuid-tree))))))))
 
-(defmethod ig/init-key :akvo.lumen.lib.auth/wrap-auth-datasets  [_ {:keys [tenant-manager flow-api monitoring] :as opts}]
-  (wrap-auth-datasets tenant-manager flow-api (:collector monitoring)))
+(defmethod ig/init-key :akvo.lumen.lib.auth/wrap-auth-datasets  [_ {:keys [tenant-manager authz-service-client monitoring] :as opts}]
+  (wrap-auth-datasets tenant-manager authz-service-client (:collector monitoring)))
 
-(s/def ::flow-api ::c.flow/config)
+(s/def ::authz-service-client any?)
```
tangrammer: trying to avoid this `any?` spec I've set up this PR https://github.com/akvo/akvo-lumen/pull/2478

dlebrero: Merged, but honestly, I wonder if adding one protocol and one record for that is worth it.

tangrammer: yep, I thought the same, that's the reason that I requested your review :) 

anyway we'll have 2 things now, we are sure that our init(iated) system will have an authentication-client-protocol there and we could also mock the component in other envs

PR #2483, opened 2019-12-18
Set pod resources requests/limits

orifito:


kardan:
The original changes should be modified and applied again.

Issue #2486, opened 2019-12-18
GroupName and GroupId should not exist in derived transformation columns


tangrammer:
# Context
Backend uses the "original column" to derive the new ones. But GroupId and GroupName are only allowed/expected in flow columns

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

tangrammer:
closed by https://akvo.slack.com/archives/CCH3D1YUT/p1576688902169500

tangrammer:
https://demo.akvolumen.org/dataset/5dfa3a7d-4268-477a-8ab0-1a6214fbce20
<img width="1194" alt="Screen Shot 2019-12-18 at 18 10 21" src="https://user-images.githubusercontent.com/731829/71107597-ad698580-21c1-11ea-9bc1-a7d8856548da.png">
working now :) @muloem

PR #2487, opened 2019-12-18
Don't include groupName nor GroupId on new derived columns

tangrammer:
Relates #2486


kardan:
Let's merge after deployments work

Issue #2489, opened 2019-12-19
Don't hit keycloak-authorization-service if user exists in the cache


tangrammer:
# Context
We get 500 status response trying to edit dashboards with many vizs, thus lumen tries to auth each one through keycloak/allowed-paths
https://github.com/akvo/akvo-lumen/blob/69b171ee712cec7ff795a8972fff95b9b9ad3432/backend/src/akvo/lumen/component/keycloak.clj#L445-L456

Cache is working fine but ... we always call request-header function  
https://github.com/akvo/akvo-lumen/blob/69b171ee712cec7ff795a8972fff95b9b9ad3432/backend/src/akvo/lumen/component/keycloak.clj#L453

so at the end we reach the http.client connection manager limits getting a `Timeout waiting for connection from pool`

# Problem or idea
https://console.cloud.google.com/logs/viewer?interval=NO_LIMIT&project=akvo-lumen&authuser=1&minLogLevel=0&expandAll=false&timestamp=2019-12-19T14%3A36%3A43.021000000Z&customFacets&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22akvo-lumen%22%0Aresource.labels.cluster_name%3D%22production%22%0Aresource.labels.pod_name%3D%22lumen-blue-749d954957-g8xhd%22%0Aresource.labels.container_name%3D%22lumen-backend%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.location%3D%22europe-west1-d%22%0Atimestamp%3D%222019-12-19T14%3A35%3A08.272994021Z%22%0AinsertId%3D%22rylxs3a5j4iqddzhk%22&scrollTimestamp=2019-12-19T14%3A35%3A08.272994021Z

# Solution or next step
Avoid keycloak call until necessary

dlebrero:
Even if this change is ok, note that the issue is really on how we actually load the visualisations, which is causing a "stampede" to the backend, which is not scalable here, neither it was on the authorisation side. 

@kardan has been working on fixing how we load the dashboards.

tangrammer:
tested here 
https://lumen.akvotest.org/dashboard/5dfbb67f-6c2f-4f6c-bdaf-120110821132

tangrammer:
and here
https://demo.akvolumen.org/dashboard/5dfbb7ff-e627-4cfc-afcf-c87021d9ed41
and here
[...]

PR #2490, opened 2019-12-19
[#2489] Avoid request-header keycloak call if data exists in cache

tangrammer:
Relates #2489

### DIFF
--- a/backend/src/akvo/lumen/component/keycloak.clj
+++ b/backend/src/akvo/lumen/component/keycloak.clj
@@ -414,11 +414,13 @@
 (defn- lookup-paths [req-opts api-root user-id jwt-paths-cache paths-cache-seconds iat]
   (if-let [paths (get-paths jwt-paths-cache user-id iat)]
     paths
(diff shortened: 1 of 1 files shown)


backend/src/akvo/lumen/component/keycloak.clj
```diff
@@ -450,10 +450,9 @@
   remove the leading /akvo/lumen and return paths as demo/admin."
   [{:keys [api-root user-id-cache jwt-paths-cache paths-cache-seconds connection-manager monitoring] :as keycloak} {:keys [email iat]}]
   (prometheus/with-duration (registry/get (:collector monitoring) :app/auth-allowed-paths {})
-    (when-let [headers (request-headers keycloak)]
```
dlebrero: We are not checking now that the result of `request-headers` is nil. Is this on purpose?

tangrammer: now it's better, tx!

Issue #2500, opened 2020-01-10
CI don't capture E2E tests logs


kardan:
# Context
In out CI build pipeline we run the end to end tests (e2e). In this step among other things, we enforce Clojure specs checks.

# Problem or idea
In SemaphoreCI the logs of the e2e tests are lost. Making any spec problem really difficult to figure out.
https://akvo.semaphoreci.com/jobs/32ef441b-cae8-4019-bc49-02c2fc9e8e80#L443

# Solution or next step
Fix the issue..

## Ref
https://github.com/SeleniumHQ/docker-selenium/issues/87
https://forums.balena.io/t/dbus-failed-to-connect-to-socket-host-run-dbus-system/3107/14

tangrammer:
@kardan IMO this could be closed by https://github.com/akvo/akvo-lumen/pull/2507

PR #2506, opened 2020-01-13
[#2500] Use latest akvo/akvo-lumen-e2e-test

iperdomo:
* We remove the `ELECTRON_ENABLE_LOGGING=true` to avoid a misleading
  dbus error:
  `Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory`

* More info at:
  https://github.com/cypress-io/cypress/issues/4925


tangrammer:
Great!

Issue #2510, opened 2020-01-15
Search library on 'enter' key press


tangrammer:
# Context
<img width="1137" alt="Screenshot 2020-01-15 at 11 13 23" src="https://user-images.githubusercontent.com/731829/72425462-24c21300-3788-11ea-8f38-dad6e22976c8.png">

as a final user, we miss this click enter behavior

PR #2511, opened 2020-01-15
[#2510] Search library on click 'enter' key

tangrammer:

### DIFF
--- a/client/src/components/library/LibrarySearch.jsx
+++ b/client/src/components/library/LibrarySearch.jsx
@@ -24,7 +24,7 @@ export default class LibrarySearch extends Component {
           className="search"
           onChange={evt => this.setState({ searchText: evt.target.value })}
           value={this.state.searchText}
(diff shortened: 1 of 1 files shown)


client/src/components/library/LibrarySearch.jsx
```diff
@@ -24,6 +24,7 @@ export default class LibrarySearch extends Component {
           className="search"
           onChange={evt => this.setState({ searchText: evt.target.value })}
           value={this.state.searchText}
+          onKeyDown={o => (o.key === 'Enter' && this.props.onSearch(this.state.searchText.trim()))}
```
iperdomo: While I understand that we want a one-line here, why not `evt` for the local variable instead of `o`? What does `o` means? ... `evt` is widely use as variable for _event_, a [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) in this case?

tangrammer: fixed! 👍

Issue #1656, opened 2018-09-14
Add filters to dashboards 


janagombitova:
# Context
*Why do we add this issue?*

Dashboards hold multiple visualisations and text to showcase a situation of a project, tell a story, to showcase the situation on the ground to eventually drive decisions. 

# Problem 

You have a dataset for an entire country and create different visualisations. The dashboard holds these visualisations and you see the visualisations being based on the entire country. Then you want to see what the situation is in one region (a value in the dataset). So you select that region in your filter and the visualisations adjust to be based on the filtered out data. Or you want to see the status for all water points built by a specific organisation (another variable in the dataset). 

This is not possible today in Lumen. 

## Current status quo
Today users have to make a separate dashboard with the exact same visualisation for each district. However, for more than 10 districts making a dashboard for each separately is cumbersome and you loose overview in the bar graphs.

# Solution or next step

The idea is to be able to filter an existing dashboard. I view a shared dashboard and then select to show me the chart based on a certain variable. I expect the created visuals to adapt to my filter. 

Example holding open data: http://213.136.89.20/akvo/ethiopia-water-points-functionality/

**To keep in mind**: Filtering assumes you are working with one dataset. But in Lumen, you can create one dashboard that holds visualisations from many datasets or you only select visualisations from one dataset. 

Some initial sketches from 2017 can be found here: https://projects.invisionapp.com/share/9TCLA2ORF#/screens/243782421

janagombitova:
I spoke with Irene today to understand a bit more on the use of dashboards with filters, what they expectations are, etc. 

* dashboards are used for monitoring a project and its impact and results. The filters would then help to see the differences and results per different regions that are part of the project, or for a specific timeframe, or component
* today, for example, she created three dashboards for Football for Water with the 'same' visuals per different area: 
   * https://footballforwater.akvolumen.org/dashboard/5d15d9c4-3550-4564-b49b-989676a78d69
   * https://footballforwater.akvolumen.org/dashboard/5d15d32f-a4af-47a8-a8ec-75e0c8d3c6c7
   * https://footballforwater.akvolumen.org/dashboard/5d14c364-97fd-45c8-b01e-44a064dcd835
* the person creating the dashboard is the one who will also decide on which columns the viewer can filter the visualisations on the shared dashboard 
* the expectation is that once the viewer selects her filter, the visualisation will adjust based on that data
[...]

PR #2512, opened 2020-01-15
[#1656] Filtered Dashboard: Add feature flag based on auth0 jwt metadata

tangrammer:
path to flag > 'lumen' > 'features' > 'filteredDashboard'

Relates #1656


iperdomo:
@tangrammer can you show a token containing that metadata? Is not the case we didn't want to use extra features from Auth0 and stick to OIDC?

tangrammer:
@iperdomo it's an internal and provisional feature-flag ... no need to think in using standards/specs 
don't you think the same? 🤔

tangrammer:
Also, the feature flag is to be used in `test` and `prod` environments so we know that we'll have auth0 meta-data there

Issue #2534, opened 2020-01-31
Exporter timeout in maps with lot of data


tangrammer:
# Context
We received a reamaze issue about problem with exporting vizs.( In link provided there are 2 issues, just pay attention to the viz related problem)
https://flowhelp.reamaze.com/admin/conversations/probleme-lumen-33c0bd8249ea1802

# Problem or idea
🤔 no clear idea yet, just merged/deployed to test (and today reverted) a change to specify timeout in puppeteer waitFor selector ... anyway found that there is something weird in the processes launched and timeout relations. Also we should to take care of sync timeout client expectations with timeout backend expectations 

# Solution or next step
...

tangrammer:
example logs trying to export this map viz: https://lumen.akvotest.org/visualisation/5d137ab0-0d12-4153-b978-348e1983ed55
start date : `2020-02-03 08:23:26.558` 
end date: `2020-02-03 08:29:17.891`

attached json logs in the txt file

[null__logs__2020-02-03T09-29.txt](https://github.com/akvo/akvo-lumen/files/4146884/null__logs__2020-02-03T09-29.txt)

tangrammer:
locally tested with the same data and it works fine
<img width="778" alt="Screenshot 2020-02-03 at 10 56 40" src="https://user-images.githubusercontent.com/731829/73644161-1c694380-4675-11ea-89aa-132f83fb8bd7.png">



![Untitled visualisation (2)](https://user-images.githubusercontent.com/731829/73644150-17a48f80-4675-11ea-96a6-5ab8b92bd001.png)
![Untitled visualisation (1)](https://user-images.githubusercontent.com/731829/73644151-183d2600-4675-11ea-8384-e3d86c31864c.png)

tangrammer:
included viz on akvotest https://lumen.akvotest.org/visualisation/5e393b63-7144-453a-be41-76d2b2fff4d9

tangrammer:
After adding more logs and more timeout, and doing in the local browser the same exporter http call ... finally reached the final entwined reasons 😄 !

## exporter/index.js expectations
from the point of view of the exporter, the logic to know if a graph is properly loaded is to wait for a HTML div class name,  something like `.render-completed-XXXXXXXXXX`, being the `XXXXXXXXXXX` the UUI of the visualization. So far, so good! 

## Leaftlet behaviour
[...]

PR #2535, opened 2020-01-31
[#2534] Add specific wait for selector timeout [exporter]

tangrammer:


tangrammer:
merging it to try it in `test` env with `real` map data

PR #2600, opened 2020-03-09
Issue/2509 user feedback

tangrammer:

### DIFF
--- a/client/src/index.ejs
+++ b/client/src/index.ejs
@@ -7,6 +7,7 @@
     <link rel="shortcut icon" href="/favicon.ico">
     <script src="https://use.fontawesome.com/51ad00bb8c.js"></script>
     <script async data-cfasync='false' src='https://d29l98y0pmei9d.cloudfront.net/js/widget.min.js?k=Y2xpZW50SWQ9NzQ3Jmhvc3ROYW1lPWFrdm9sdW1lbi5zdXBwb3J0aGVyby5pbw=='></script>
+    <script async defer data-bugyard='5e65daa3f74c520014b05cfa' src='https://widget.bugyard.io/bugyard.min.js'></script>
   </head>
   <body>
     <div id="root"></div>
--- a/client/src/index.ejs
+++ b/client/src/index.ejs
@@ -7,7 +7,9 @@
     <link rel="shortcut icon" href="/favicon.ico">
     <script src="https://use.fontawesome.com/51ad00bb8c.js"></script>
     <script async data-cfasync='false' src='https://d29l98y0pmei9d.cloudfront.net/js/widget.min.js?k=Y2xpZW50SWQ9NzQ3Jmhvc3ROYW1lPWFrdm9sdW1lbi5zdXBwb3J0aGVyby5pbw=='></script>
-    <script async defer data-bugyard='5e65daa3f74c520014b05cfa' src='https://widget.bugyard.io/bugyard.min.js'></script>
+    <script type="text/javascript">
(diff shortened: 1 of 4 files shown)


client/src/index.ejs
```diff
@@ -7,7 +7,9 @@
     <link rel="shortcut icon" href="/favicon.ico">
     <script src="https://use.fontawesome.com/51ad00bb8c.js"></script>
     <script async data-cfasync='false' src='https://d29l98y0pmei9d.cloudfront.net/js/widget.min.js?k=Y2xpZW50SWQ9NzQ3Jmhvc3ROYW1lPWFrdm9sdW1lbi5zdXBwb3J0aGVyby5pbw=='></script>
-    <script async defer data-bugyard='5e65daa3f74c520014b05cfa' src='https://widget.bugyard.io/bugyard.min.js'></script>
+    <script type="text/javascript">
+      !function(){if("function"!=typeof window.bugyard){var a=function(){a.c(arguments)};a.q=[],a.c=function(b){a.q.push(b)},window.bugyard=a;var b=document.createElement("script");b.setAttribute("data-bugyard","5e65daa3f74c520014b05cfa"),b.setAttribute("async","async"),b.setAttribute("defer","defer"),b.setAttribute("src","https://widget.bugyard.io/bugyard.min.js"),document.getElementsByTagName("head")[0].appendChild(b)}}();
```
iperdomo: It would be interesting to understand why the change? Why do we need a script to include an external `.js`
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
````js
!(function() {
  if ("function" != typeof window.bugyard) {
    var a = function() {
      a.c(arguments);
    };
    (a.q = []),
      (a.c = function(b) {
        a.q.push(b);
      }),
      (window.bugyard = a);
    var b = document.createElement("script");
    b.setAttribute("data-bugyard", "5e65daa3f74c520014b05cfa"),
      b.setAttribute("async", "async"),
      b.setAttribute("defer", "defer"),
      b.setAttribute("src", "https://widget.bugyard.io/bugyard.min.js"),
      document.getElementsByTagName("head")[0].appendChild(b);
  }
})();

````

File: client/src/utilities/feedback.js [hunk 2]
```diff
@@ -1,14 +1,13 @@
...
 
 export const init = (state) => {
-  if (feedbackInited || !state.profile) {
+  if (!state || !state.profile) {
     return;
   }
-  feedbackInited = true;
   window.bugyard('setUser', {
     email: 'devops@akvo.org',
+    name: state.profile.sub || 'Unknown',
```
iperdomo: `name` is not a required field, and we already passing an `id`

client/src/utilities/feedback.js
```diff
@@ -1,14 +1,13 @@
 /* eslint-disable no-undef, no-use-before-define, no-underscore-dangle */
 
-let feedbackInited;
 
 export const init = (state) => {
-  if (feedbackInited || !state.profile) {
+  if (!state || !state.profile) {
     return;
   }
-  feedbackInited = true;
```
iperdomo: My understanding of the flag `feedbackInitiated` is for calling `window.bugyard` only once as the `init` function is hooked to the store changes:
* https://github.com/akvo/akvo-lumen/blob/2bb84614edb64c3a9c233a95830784a41e942409/client/src/app.jsx#L32
* https://redux.js.org/api/store/#subscribelistener

client/src/index.ejs
iperdomo: I thinkk that using HTML5 `async` and `defer` for the script tag has the same effect as injecting it via a `document.createElement("script")` as we're not including the same script more than once -  https://stackoverflow.com/a/14666847

client/src/utilities/feedback.js
tangrammer: the thing is that user details are not yet working and i was trying to pass another piece of data to understand the problem better, but no idea yet

client/src/utilities/feedback.js
tangrammer: yep, you're right but user details not working yet so was trying things to know more about the problem

Issue #2596, opened 2020-03-09
Provide a way for gather user feedback


iperdomo:
# Context
We're already capturing _exceptions_ in the client and server, but we want to collect _feedback_ and bugs/errors from the users.

# Solution or next step
Select and integrate a user feedback service in Lumen

tangrammer:
we should keep plugin on shared dashboard and remove it from exporter version

tangrammer:
same with vizs shared and exported

tangrammer:
@iperdomo should we close this issue?

PR #2607, opened 2020-03-11
[#2596] Adapt to public and export views

tangrammer:


client/src/index.ejs
```diff
@@ -7,15 +7,20 @@
     <link rel="shortcut icon" href="/favicon.ico">
     <script src="https://use.fontawesome.com/51ad00bb8c.js"></script>
     <script async data-cfasync='false' src='https://d29l98y0pmei9d.cloudfront.net/js/widget.min.js?k=Y2xpZW50SWQ9NzQ3Jmhvc3ROYW1lPWFrdm9sdW1lbi5zdXBwb3J0aGVyby5pbw=='></script>
-    <script>
+    <script type="text/javascript">
+      const url = new URL(window.location.href);
```
iperdomo: There is no need to create a new `URL` ... `location` is already an object, that object has the `pathname` property.
* https://developer.mozilla.org/en-US/docs/Web/API/Window/location

tangrammer: changed!

PR #2608, opened 2020-03-13
Issue 2606 wip

iperdomo:
Delay the call to `invalidateSize()` so the parent has the new dimension.

When we trigger the `renderLeafletMap` on `componentDidMount` and
`UNSAFE_componentWillReceiveProps` the dimensions of the parent container
element are the *old* ones.

Found it via:

    let parent = this.leafletMapNode.parentElement;
    console.log(parent.clientHeight, parentclientWidth);

More info:

    PaulLeCam/react-leaflet#40


tangrammer:
@iperdomo i copied the commit description to the PR description too

Issue #2569, opened 2020-02-19
Fully upgrade react to 16.8.0 (hooks version)


tangrammer:
# Context
We need to fully upgrade client code to use current react 16.8.0 version. That means to follow new lifecycle approach and remove the UNSAFE_  prefixes already added for compatibility 

Also remove warnings
<img width="1077" alt="Screenshot 2020-02-19 at 09 25 36" src="https://user-images.githubusercontent.com/731829/74815691-ff31a780-52f9-11ea-88eb-2b6674d4dcfe.png">

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

janagombitova:
@tangrammer considering we flipped a lot of changes to this to production today, is there more to be done here or can we close the issue?

tangrammer:
still we need to take care of migrating 
`react-dom.development.js?9f84:89 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

PR #2652, opened 2020-04-14
[#2569] Use react hooks in MapVisualisation [poc]

tangrammer:

Using react hooks improves the current design. 
It could be better if we could use https://react-leaflet.js.org/ instead of plain leaflet but then we have to take care of current leaflet.utfgrid wrapper https://github.com/akvo/akvo-lumen/blob/issue-2569-try-hooks/client/src/vendor/leaflet.utfgrid.js#L1

### DIFF
--- a/client/src/components/charts/MapVisualisation.jsx
+++ b/client/src/components/charts/MapVisualisation.jsx
@@ -234,6 +234,7 @@ PopupContent.propTypes = {
 
 const MapVisualisation = (props) => {
   const [hasRendered, setHasRendered] = useState(false);
+  // eslint-disable-next-line no-unused-vars
   const [hasTrackedLayerTypes, setHasTrackedLayerTypes] = useState(false);
   const hasAddedLayers = useRef(false);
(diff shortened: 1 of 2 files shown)


client/src/components/charts/MapVisualisation.jsx
```diff
@@ -439,147 +403,180 @@ export default class MapVisualisation extends Component {
...
+  }, [hasTrackedLayerTypes]);
+
+  useEffect(() => {
+    if (!hasAddedLayers.current) {
+      const loadInterval = setInterval(() => {
+        const checks = Object.values(mapp.current._layers).map((l) => {
+          try {
+            return l.isLoading();
+          } catch (error) {
+            // eslint-disable-next-line no-console
```
iperdomo, line 471: We should remove this. We don't want `console.log` in production.

iperdomo: If is an error we can fix (act upon), send it to Sentry

PR #2669, opened 2020-05-04
Remove sampling in bubble/line/scatter vizs and warn the user if results are greater than 2.5K

tangrammer:
- [ ] review current existing charts that could return more than 2.5K and notify the user about them

### DIFF
--- a/backend/src/akvo/lumen/endpoint/aggregation.clj
+++ b/backend/src/akvo/lumen/endpoint/aggregation.clj
@@ -27,10 +27,6 @@
             (log/error e)
             (http/bad-request
              {:message (format "JSON parse exception: %s" (.getMessage e))}))
-          (catch java.sql.SQLException e
-            (log/error e)
-            (http/bad-request
-             {:message  (format "Database query exception: %s" (.getMessage e))}))
           (catch Exception e
             (log/error e)
             (http/internal-server-error {:message (.getMessage e)})))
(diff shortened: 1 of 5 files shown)


backend/src/akvo/lumen/endpoint/aggregation.clj
```diff
@@ -23,7 +24,16 @@
...
                              (json/parse-string query keyword))
           (catch JsonParseException e
-            (http/bad-request {:message (.getMessage e)})))
+            (log/error e)
+            (http/bad-request
+             {:message (format "JSON parse exception: %s" (.getMessage e))}))
+          (catch java.sql.SQLException e
+            (log/error e)
+            (http/bad-request
+             {:message  (format "Database query exception: %s" (.getMessage e))}))
```
iperdomo, line 33: I guess we expose this message to the user? I don't think is _useful_ to export database errors but meaningful ones

backend/src/akvo/lumen/endpoint/aggregation.clj
```diff
@@ -23,7 +24,16 @@
                              visualisation-type
                              (json/parse-string query keyword))
           (catch JsonParseException e
-            (http/bad-request {:message (.getMessage e)})))
+            (log/error e)
+            (http/bad-request
+             {:message (format "JSON parse exception: %s" (.getMessage e))}))
+          (catch java.sql.SQLException e
```
iperdomo, line 30: We can `(import java.sql.SQLException)` in the namespace declaration, and use `SQLException`

iperdomo:
As agreed, we need to be _pro-active_ and check how many visualizations will be affected by the change.

Issue #2537, opened 2020-02-03
Scatter visualization returns unstable/random results 


tangrammer:
# Context
https://flowhelp.reamaze.com/admin/conversations/lumen-bug

# Problem or idea
Reproducible example:
just reload this vis to get random results
https://akvointernal.akvolumen.org/visualisation/5e342bb6-1a91-4b9d-b5f2-c2ad6e267ec5

<img width="775" alt="Screenshot 2020-02-03 at 11 57 13" src="https://user-images.githubusercontent.com/731829/73647778-89cca280-467c-11ea-87a9-4d4d2a1025b7.png">
<img width="776" alt="Screenshot 2020-02-03 at 11 57 04" src="https://user-images.githubusercontent.com/731829/73647779-89cca280-467c-11ea-9c75-f40dd860e12a.png">
<img width="801" alt="Screenshot 2020-02-03 at 11 56 57" src="https://user-images.githubusercontent.com/731829/73647780-89cca280-467c-11ea-8fbf-b365520c9d91.png">

After analyzing different request payloads the visualization specification keeps stable so it's not related to that part

# Solution or next step
How could we solve it?

tangrammer:
looking into the scatter aggregation query details I realised that we are using a `max-points:2500` default value with an `ORDER BY random()` Postgres function.
These 2 options working together make indeterministic the query/graph result on datasets with rows above 2500 rows as the example provided in the previous issue description

Also found that this approach is the same as we have with `line` and `bubble` graphs

@muloem what do think we should proceed?

janagombitova:
Considering our partner datasets will have more than 2500 rows, this needs to be handled to ensure partners can rely on the visualisations.

iperdomo:
>  Does there need to be the limit?

Yes. We need to define a limit. Also make sure that we notify the user if a dataset backing the visualization is larger than the defined limit.

iperdomo:
Just to clarify my previous comment, even if we define that we support 1M rows, we need to define and enforce that limit. It can't be _boundless_.

janagombitova:
500k rows, would that we a good boundary? We do not have a dataset this big, yet so it gives us a buffer.

iperdomo:
Checking in local development the count **match** but the result of 2 runs is different:

```
lumen_tenant_1=> select count(*) from ds_35433b39_23ed_48f6_8180_f74b29035b3e;
 count 
-------
  5000
(1 row)

[...]

PR #2672, opened 2020-05-05
[#2537] Remove sampling in bubble/line/scatter v2 spec vizs and warn the user if results are greater than 2.5K 

tangrammer:

- add backward compatibility with v1 (the sampling one)


tangrammer:
@iperdomo 
I've created a new PR with almost the same contents as https://github.com/akvo/akvo-lumen/pull/2669 

the difference is backward compatibility. Older vizs that weren't edited will work the same as they do but any modification will increment the spec.version and will show the warning prompt to the user in case the results are greater than 2.5K

PR #2682, opened 2020-05-11
Delivery metrics support

### LINKED ISSUE #2681: Delivery metrics support
Record if a release is due to a hotfix or not

dlebrero:
Fixes #2681

### DIFF
--- a/ci/promote-test-to-prod.sh
+++ b/ci/promote-test-to-prod.sh
@@ -80,7 +80,7 @@ git log --oneline $NEWEST_VERSION_IN_PROD..$TEST_LIVE_VERSION | grep -v "Merge p
 TAG_NAME="promote-$(TZ=UTC date +"%Y%m%d-%H%M%S")"
 
 echo ""
-read -r -e -p "Does this deployment contain a hotfix, rollback or fix-forward for a previous deployment? [yn] " FIX
+read -r -e -p "Does this deployment contain a hotfix, rollback or fix-forward for a previous deployment? [Y/n] " FIX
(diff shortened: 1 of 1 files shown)


ci/promote-test-to-prod.sh
```diff
@@ -72,15 +72,24 @@ else
...
 log "Commits to be deployed:"
+echo ""
 git log --oneline $NEWEST_VERSION_IN_PROD..$TEST_LIVE_VERSION | grep -v "Merge pull request" | grep -v "Merge branch"
 
 "${DIR}"/helpers/generate-slack-notification.sh "${NEWEST_VERSION_IN_PROD}" "${TEST_LIVE_VERSION}" "Promoting Lumen to dark prod cluster" "good"
 
 TAG_NAME="promote-$(date +"%Y%m%d-%H%M%S")"
 
+echo ""
+read -r -e -p "Does this deployment contain a hotfix, rollback or fix-forward for a previous deployment? [yn] " FIX
```
iperdomo, line 83: When presenting a prompt `yes/no`, `y/n` usually cli tools _take a side_ ... either `Y` or `N` ... `[Yn`] or `[yN]` and is expressed with a capital letter. What will be the default value if the user hits `<ENTER>`. That's _UX_ at cli level :wink: 

Based on the following test, it seems that we default to `Y` because the user needs to explicitly answer `n` to consider it a `REGULAR_RELEASE`. So we should change the prompt to `[Yn]` or better `[Y/n]`

https://stackoverflow.com/questions/29436275/how-to-prompt-for-yes-or-no-in-bash/29436423#29436423

Issue #2127, opened 2019-05-06
Support Flow repeated question groups


janagombitova:
# Context
This issue https://github.com/akvo/akvo-lumen/issues/529 has given a start to this topic but only a little bit has been touched upon in the issue as it derailed to also redesigning the data grid. 

## Why do we add this issue?
We are working on connecting Flow and Lumen and in this process, we have decided that we need to first handle the current gaps between Flow and Lumen. Lumen needs to be able to handle all Flow data types and data structures. Initially, Lumen was built not knowing too much about Flow, but with the change to move towards one product, Lumen needs to handle all 'special' Flow data types in a simple and user-friendly way. 

# Problem or idea

## What are repeated question groups? 

Repeated question groups are our alternative to table questions. Often a group of questions needs to be asked multiple times when filling in the questionnaire. For example, if you want to ask the same questions to each member of a household, you will add the questions to one question group and click on the repeat option. This way you do not need to pre-define in the survey the questions for each member of the household, as that is an impossible task. 

Repeated question groups are, simply said, a group of questions (of any kind) that the respondent can fill in more than once. 

## Repeated question group data 
Using repeated question groups results in one submission holding more than one answer to one or more questions. The number of repetitions can differ per submission. 

When handling repeated question group data in a table with Flow exports we learned a lot (see this issue: https://github.com/akvo/akvo-product-design/issues/227) on what way of presenting data works and what does not work for our users: 

### 1) All data in one sheet with the repetitions in separate rows
![Screen Shot 2019-05-13 at 11 05 22](https://user-images.githubusercontent.com/12456965/57609378-09ef2d00-756f-11e9-8aee-7ea54c3c1478.png)

This data structure, where all data was in one sheet and the repetitions were in separate rows has proven not useful to our users. They complained about the empty cells, it was hard to see which column is repeated, complexity in having to analyse data that was not repeated this way. Thus we changed the export options in Flow to explore other solutions. 

This structure is still supported in Flow's Data analysis exports as we aimed to keep all the data in one sheet. However, we still receive requests on changing this structure to either have separate sheets or columns. 

### 2) Repeated group data split into a separate sheet with data point metadata copied over
![Screen Shot 2019-05-13 at 11 07 21](https://user-images.githubusercontent.com/12456965/57609514-53d81300-756f-11e9-8979-1dd74c3952e3.png)

[...]

PR #2693, opened 2020-05-14
[#2127] Add column metadata

tangrammer:
Metadata will be used so far to store RQG columns metadata

the future api response with rqg metadata will be more or less like ...

Screenshot 2020-05-13 at 13 51 47


tangrammer:
> So the extra metadata will only be relevant to columns of a RQG? The rest of the columns dont need this metadata?

yep!

PR #2695, opened 2020-05-14
Import one RQG column in flow import/update processes (with feature-flag)

tangrammer:
- [ ] **Update release notes if necessary**

The review should be better if you go through the commits in order 🤓

PR #2698, opened 2020-05-14
Import one RQG column in flow import/update processes (with feature-flag)

tangrammer:


tangrammer:
@iperdomo merging these changes as are the same as approved ones #2695

PR #2700, opened 2020-05-15
Issue/2127 rqg js

iperdomo:
- Fixes spec definition (metadata is a `vector` not a `map`)
  - An error in this problematic spec was hidden behind a `(catch Exception _)` . The function instrumentation was complaining and reporting a misleading error :cry: 

- The main change is to help a user write less code when by converting `rqg` columns to a JS object automatically, __before__ executing the user's code. We do it by enhancing the _adapter_ that was converting column names -> column title and adding a key `__rgq__` to the `row` object.  This key has an array of all columns of type `rqg`. The transformation function loops that array (if present) and calls `JSON.parse()` on those column's values.

- The `row` object passed to the JS context was previously a Clojure immutable datastructure, because we're mutating those `rqg` keys, we convert that object in to a mutable `HashMap` before passing it to the JS code.

### DIFF
--- a/backend/specs/akvo/lumen/specs/import/values.clj
+++ b/backend/specs/akvo/lumen/specs/import/values.clj
@@ -37,7 +37,7 @@
                      string?
                      #(s/gen #{"c1" "c2" "c3"})))
 
-(s/def ::metadata (s/nilable map?))
+(s/def ::metadata (s/nilable vector?))
 
 (s/def ::title (s/with-gen
                         string?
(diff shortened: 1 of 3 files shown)


tangrammer:
@iperdomo BTW, do we have new tests for this new js functionality extensión?

iperdomo:
@tangrammer we don't have tests. no

tangrammer:
> @tangrammer we don't have tests. no

https://github.com/akvo/akvo-lumen/pull/2703 one here

tangrammer:
we can merge the unit test later if we think so

PR #2703, opened 2020-05-18
[#2127] Add unit test for derive rqg columns as js object

tangrammer:

### DIFF
--- a/backend/test/akvo/lumen/lib/transformation/derive/js_engine_test.clj
+++ b/backend/test/akvo/lumen/lib/transformation/derive/js_engine_test.clj
@@ -1,15 +1,71 @@
 (ns akvo.lumen.lib.transformation.derive.js-engine-test
   (:require
+   [akvo.lumen.lib.transformation.derive :as derive]
    [akvo.lumen.lib.transformation.derive.js-engine :as js-engine]
    [clojure.edn :as edn]
    [clojure.string :as string]
-   [clojure.test :refer [deftest is]]))
+   [clojure.test :refer [deftest is]]
+   [clojure.walk :as walk]))
 
 
 (deftest javascript-logic
   (is (= (js-engine/eval* "1 + 1") 2))
   (is (= (js-engine/eval* "'akvo'.toUpperCase();") "AKVO"))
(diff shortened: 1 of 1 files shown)


backend/test/akvo/lumen/lib/transformation/derive/js_engine_test.clj
```diff
@@ -1,16 +1,72 @@
...
+                     "title" "ID number",
+                     "groupId" nil,
+                     "metadata" nil,
+                     "groupName" nil}]},
+                  "title" "Repeating Biodata",
+                  "type" "rqg",
+                  "hidden" false,
+                  "columnName" "c567919211"}]
+        adapter (comp (js-engine/rqg columns)
+                      (js-engine/column-name->column-title columns))
```
iperdomo, line 56: While this is how we're using it, is not the _unit_ to test? we only want/need to test the `js/rqg` _adapter_.

backend/test/akvo/lumen/lib/transformation/derive/js_engine_test.clj
```diff
@@ -1,16 +1,72 @@
...
+             :c567919211 "[{\"Name\": \"once\", \"ID number\": 65894632, \"Age repeated\": 21.0},
+                           {\"Name\": \"twice\", \"ID number\": 66999888452, \"Age repeated\": 23.0},
+                           {\"Name\": \"three timea\", \"ID number\": 67302878563, \"Age repeated\": 21.0}]"}
+
+        columns-groups (derive/columns-groups (walk/keywordize-keys columns))
+        row-extended (derive/extend-row row adapter columns-groups)
+        row-fn (js-engine/row-transform-fn {:adapter adapter
+                                            :code "row[\"Repeating Biodata\"].filter(a => a.Name === \"once\")"})]
+    (is (= (row-fn row-extended)
+           {"0" {"Name" "once", "ID number" 65894632, "Age repeated" 21}}))))
```
iperdomo, line 67: This is a _correct_ response from evaluating the code, but we only support _Date_, _Text_, or _Number_ ... I would suggest to add `.length` and the end and expect a Number?

PR #2711, opened 2020-06-11
[#2127] React hooks approach on Dataset view

tangrammer:

Hooks approach are better to use so let's take this client RQG work as an opportunity

Also work on https://github.com/akvo/akvo-lumen/issues/2712 by https://github.com/akvo/akvo-lumen/commit/c80193cdfa95c969bcfb6d251d9fc05588210a61

### DIFF
--- a/client/src/components/common/PendingSaving.jsx
+++ b/client/src/components/common/PendingSaving.jsx
@@ -1,0 +1,65 @@
+import { useState, useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { SAVE_COUNTDOWN_INTERVAL, SAVE_INITIAL_TIMEOUT } from '../../constants/time';
+
+export default function PendingSaving(props) {
+  const isMountedFlag = useRef(false);
+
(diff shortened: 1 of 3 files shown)


client/src/components/common/PendingSaving.jsx
```diff
@@ -0,0 +1,65 @@
+import { useState, useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { SAVE_COUNTDOWN_INTERVAL, SAVE_INITIAL_TIMEOUT } from '../../constants/time';
+
+export default function PendingSaving(props) {
```
marvinkome, line 5: Is `PendingSaving` a hook or a component? It looks like a component but it doesn't return any jsx?

tangrammer: it's a hook designed to be composed

any ideas to improve the syntax so it looks like a hook?

marvinkome: To make it a hook just rename it to `usePendingSaving`

tangrammer: done :)

PR #2716, opened 2020-06-15
[#2127] Add backend APIs to support groups

tangrammer:

- [x] `/dataset/<id>/group/metadata` to get metadata group 
- [x] `/dataset/<id>/group/transformations` to get transformations group 
- [x] `/dataset/<id>/group/main` to get main group (in the case of csv )
- [x] `/dataset/<id>/groups` to get meta with groups 
- [x] Tests!

Anyway these "static" groupIds will be returned by backend call `/dataset/<id>/groups` returning this kind of [data](https://github.com/akvo/akvo-lumen/issues/2127#issuecomment-644232253) 

so far both groups ^^ are calculated on client-side based on domain conditions such as column names

example json responses
`/dataset/<id>/groups` https://gist.github.com/tangrammer/046a954beba038d34c1f61cbe67cbd22#file-fetch-groups-json

`/dataset/<id>/group/metadata` https://gist.github.com/tangrammer/046a954beba038d34c1f61cbe67cbd22#file-fetch-metadata-group-json

### DIFF
--- a/backend/src/akvo/lumen/lib/import/csv.clj
+++ b/backend/src/akvo/lumen/lib/import/csv.clj
@@ -63,7 +63,7 @@
         column-types))
 
 (defn valid-column-name? [column-name]
-  (boolean (re-find #"^c\d+$" column-name)))
+  (boolean (re-matches #"c\d+$" column-name)))
 
 (defn data-records [column-spec rows]
   (for [row rows]
--- a/backend/src/akvo/lumen/lib/import/csv.clj
+++ b/backend/src/akvo/lumen/lib/import/csv.clj
@@ -62,6 +62,9 @@
         column-titles
         column-types))
 
+(defn valid-column-name? [column-name]
+  (boolean (re-find #"^c\d+$" column-name)))
+
 (defn data-records [column-spec rows]
   (for [row rows]
     (apply merge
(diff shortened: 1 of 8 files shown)


tangrammer:
example json responses
`/dataset/<id>/groups` https://gist.github.com/tangrammer/046a954beba038d34c1f61cbe67cbd22#file-fetch-groups-json

`/dataset/<id>/group/metadata` https://gist.github.com/tangrammer/046a954beba038d34c1f61cbe67cbd22#file-fetch-metadata-group-json

backend/src/akvo/lumen/lib/import/csv.clj
```diff
@@ -62,6 +62,9 @@
         column-titles
         column-types))
 
+(defn valid-column-name? [column-name]
+  (boolean (re-find #"^c\d+$" column-name)))
```
iperdomo, line 66: > ;; The distinction is that re-find tries to find _any part_ of the string
> ;; that matches the pattern, but re-matches only matches if the _entire_
> ;; string matches the pattern.

https://clojuredocs.org/clojure.core/re-matches

tangrammer: no idea about it :) I've used same approach that we have in akvo.lumen.lib.transformation.engine https://github.com/akvo/akvo-lumen/blob/dataset-by-group-api/backend/src/akvo/lumen/lib/transformation/engine.clj#L112

iperdomo: the current code works because we're limiting the behaviour via the regex itself `#"^c\d+$"`, it must start with `c` and then one or more digits, it most end with a digit

tangrammer: changed!

PR #2720, opened 2020-06-19
[#2127] Use react hooks

tangrammer:
Preserving functionality using react hooks

This step could be considered a first step towards a better design to not have 600 LOC in DatasetTable(V2) logic 😅 

after this [commit ](https://github.com/akvo/akvo-lumen/pull/2720/commits/5856ce327d0804f0ca5b77a081843f94dbd0b2a6): 600 => 488 LOC

 
### DIFF
--- a/client/src/components/dataset/DatasetTableV2.jsx
+++ b/client/src/components/dataset/DatasetTableV2.jsx
@@ -65,8 +65,10 @@ function DatasetTable(props) {
   };
 
   const handleResize = () => {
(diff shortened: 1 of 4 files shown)


tangrammer:
@marvinkome could you expand the reasons for the NewDatasetWrapper function? I still don't get it 🤔

marvinkome:
Well, Initially I wanted to do the feature flag split in the DatasetTable component. But now that we're doing it in the Dataset component there's no need for the wrapper

PR #2733, opened 2020-06-24
[#2127] Revert #2698

tangrammer:

Remove all changes related with importing RQG in one json column:
- rqg column type
- metadata column attribute
- conditional execution based on feature-flag


iperdomo:
@tangrammer what will happen with the instance that has the feature flag enabled? :thinking:

tangrammer:
@iperdomo  I thought to remove the `demo` datasets that uses previous approach

PR #2736, opened 2020-06-25
[#2127] Add groupId and groupName to column [backend][db]

tangrammer:

So far we could have 2 **static** possible groupIds: `main`, `metadata` or **dynamic** ones that currently should relate to [flow-group-question-id](https://github.com/akvo/akvo-lumen/blob/master/backend/src/akvo/lumen/lib/import/flow_common.clj#L88)

PS: `transformations` is not going to be a group... but in the mid/long term maybe will be a `subgroup`

### DIFF
--- a/backend/specs/akvo/lumen/specs/import/column.clj
+++ b/backend/specs/akvo/lumen/specs/import/column.clj
@@ -41,7 +41,8 @@
 
 (s/def ::column-header (s/keys :req-un [::v/title]
                                :opt-un [::v/id
-                                        ::v/metadata]))
+                                        ::v/groupId
+                                        ::v/groupName]))
 
 (s/def ::c.text/header* (s/keys :req-un [::c.text/type] :opt-un [::v/key]))
 (s/def ::c.text/header (s/merge ::column-header ::c.text/header*))
--- a/backend/specs/akvo/lumen/specs/import/values.clj
+++ b/backend/specs/akvo/lumen/specs/import/values.clj
@@ -37,11 +37,13 @@
                      string?
                      #(s/gen #{"c1" "c2" "c3"})))
 
-(s/def ::metadata (s/nilable map?))
-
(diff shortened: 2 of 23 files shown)


backend/specs/akvo/lumen/specs/import/column.clj
```diff
@@ -41,7 +41,8 @@
 
 (s/def ::column-header (s/keys :req-un [::v/title]
                                :opt-un [::v/id
-                                        ::v/metadata]))
```
tangrammer, line 44: this line relates to https://github.com/akvo/akvo-lumen/pull/2733

backend/specs/akvo/lumen/specs/import/values.clj
```diff
@@ -37,12 +37,14 @@
                      string?
                      #(s/gen #{"c1" "c2" "c3"})))
 
-(s/def ::metadata (s/nilable map?))
```
tangrammer, line 40: also relates to https://github.com/akvo/akvo-lumen/pull/2733

backend/src/akvo/lumen/lib/import/flow_v3.clj
```diff
@@ -18,20 +18,19 @@
                                              (assoc :type "GEO-SHAPE-FEATURES")
                                              (assoc :multipleType "geo-shape-features")
                                              (assoc :multipleId (:id i))
-                                             (assoc :ns (:ns i))
```
tangrammer, line 21: this `ns` shouldn't appear here "yet"

iperdomo:
> So far we could have 3 static possible groupIds: main, transformations, metadata or dynamic ones that currently should relate to flow-group-question-id

Here we're making the `transformations` a first class citizen. I think we're not fully decided on this, right?

tangrammer:
@iperdomo removed `transformations` as a first class citizen group

Issue #2748, opened 2020-07-02
Importing new datasets does not recognise repeated groups


janagombitova:
When testing today on dark-demo, I noticed that a new dataset that I import from Flow does not recognise that there is a repeated group in it. Existing datasets, that I used the days before are fine. I am not sure if this is a new issue introduced today or it was here longer. 

**Example 1**  
* Flow: Dataset: uat1 > _ _ Lumen test > Example dataset (for testing and demos) > form ID: 554339123
* repeated group is `Family`
* Lumen: https://dark-demo.akvolumen.org/dataset/5efdb44b-c50a-4431-8ffc-53f6b1f42fdb

**Example 2**
* Flow: uat1 > _ _ Lumen test >Imported data - larger - with RQGs > Form ID: 214279115 
* repeated group is the second group
* Lumen: https://dark-demo.akvolumen.org/dataset/5efdb2d9-17e3-472d-91fa-f999d6fe57a0

I also noticed that when looking at the datasets in the Library, they do not indicate the origin `from Flow`. Could this be the problem? But then they do show the group navigation, so Lumen knows it is from Flow as with non-Flow datasets we do not show the data group navigation.

iperdomo:
> I noticed that a new dataset that I import from Flow does not recognise that there is a repeated group in it.

@janagombitova what do you mean by **does not recognise** ?

We removed the current implementation of RQG, the one that places everything in one cell, as per PR: https://github.com/akvo/akvo-lumen/pull/2733

In the comments I asked about the existing imported datasets:
https://github.com/akvo/akvo-lumen/pull/2733#issuecomment-648938109 https://github.com/akvo/akvo-lumen/pull/2733#issuecomment-649254485

@tangrammer said that he was going to delete those:
https://github.com/akvo/akvo-lumen/pull/2733#issuecomment-648946798

janagombitova:
I see. So then the following question is, why does the UI not mark the groups as repeated with the icon? @marvinkome

iperdomo:
> I see. So then the following question is, why does the UI not mark the groups as repeated with the icon?

I think because the client _doesn't know_ that is repeated. I'm checking if the metadata about the group is there.

janagombitova:
Thank you!

PR #2750, opened 2020-07-02
[#2127] When `rgq` flag is on, add `ns` and `repeatable` keys

iperdomo:
The client needs to know which data groups are repeatable. For that,
we client can rely on the metadata in the column definition.

After importing a Flow form, we store as part of the columns
definition the `ns` and `repeatable` keys in the columns and expose
them as part of the endpoint `/dataset/<id>/groups`.

### DIFF
--- a/backend/src/akvo/lumen/lib/import.clj
+++ b/backend/src/akvo/lumen/lib/import.clj
@@ -21,7 +21,8 @@
 
 (defn- successful-execution [conn job-execution-id data-source-id table-name columns {:keys [spec-name spec-description]} claims]
   (let [dataset-id (util/squuid)
-        imported-table-name (util/gen-table-name "imported")]
+        imported-table-name (util/gen-table-name "imported")
+        {:strs [rqg]} (env/all conn)]
     (db.dataset/insert-dataset conn {:id dataset-id
                           :title spec-name ;; TODO Consistent naming. Change on client side?
                           :description spec-description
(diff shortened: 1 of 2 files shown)


marvinkome:
What does the `ns` column mean? And currently testing, In the metadata group `repeated` is set to `true`

muloem:
`ns` from what I understood of Ivan's explanation is the way to define to which data group these columns belong.  it uses the `groupId`

janagombitova:
I assume this change is to handle: #2748 . Is that so? Is it done? @marvinkome

marvinkome:
Both that and the transformation issue

marvinkome:
There's an implementation issue with the PR. Metadata group has `repeatable` set as `true` when It's not a `rqg`

janagombitova:
As Ivan will not be around to make the changes, can one of you handle it?

tangrammer:
@marvinkome I can't reproduce the problem described here https://github.com/akvo/akvo-lumen/pull/2750#issuecomment-654137880 

I have imported a flow dataset with feature-flag rqg=true, then `metadata` columns have `repeatable`=false
Screenshot 2020-09-02 at 15 12 56

Are you aware that ns and repeatable column fields are filled on the `imported` process so you need to import a new flow dataset or update an existing one to test this functionality?

marvinkome:
Ahh okay. I'll try it again

tangrammer:
PS: ns means namespace, and should be renamed to it in the following backend PRs
PS2: ns=`null` and ns=`main` are equivalents

Issue #2753, opened 2020-07-03
Bug:: With the new data group, transformation takes user back to Metadata group


marvinkome:
# Context
Why do we add this issue?
According to @Kiarii 
>  If I go to a data group, and sort A/D, I get some weird behavior: grid applies the sort > reloads back to metadata data group and closes the data group (sidebar), navigator

janagombitova:
Tested on lumen.akvotest.org works well 👍 good job

PR #2754, opened 2020-07-03
[#2753] - Stop transformation from taking user to metadata

marvinkome:
This commit contains some changes

- First, Refactored the effect for dataset groups
- Check if a group is already selected then re-fetch the selected group
data
- Stop the sidebar from automatically closing on every component update


marvinkome:
We're re-fetching to get the transformation alongside the group object

Issue #2749, opened 2020-07-02
Add walkthrough to new dataset UI


janagombitova:
# Context
## Why do we have this issue? What are we trying to solve? 

We want our products to make it easy to work with data. This also means that it should be easy to work with the products and thus easy to get started with different product features. Having a way to explain what different buttons do and how they behave is a nice way to get started with the tool. 

## How will this benefit the users?
* easier pick up of Flumen 
* less guessing when working with the data grid 

## How will this benefit Akvo? 
* streamline the communication - we communicate where it is relevant 
* ensure the product is explained in a unified and consistent way - by the product itself

# Opportunity 
## Current status quo
Currently, the only way users learn about the product pieces (or changes to them) is by exploring and reading the support documentation. It is ok but there are more friendlier ways to introduce features and changes to users. 

A while back we decided to use https://introjs.com/example/hello-world/index.html in Flow to walk users through new features. 

## The idea 
The idea is to do the same for Lumen. To use IntroJs and to walk the users through the dataset view. Why? With supporting RQGs we are changing the data grid quite a lot (view, navigation, how derived columns behave). So this is a good opportunity to explain the changes to existing users and explain the data grid to new users. 

<img width="1266" alt="Screen Shot 2020-07-06 at 11 06 23" src="https://user-images.githubusercontent.com/12456965/86576237-c0cf1900-bf78-11ea-8edf-b5980ac062b2.png">

* the change should be part of the RQG feature flag. 
* we should introduce the different parts of the grid and what to expect (for example with `Transformations` you create a new column and this will land at the end of the data group)

## Data set view components 
* dataset header (user defined) 
* autosave update
* transfor action 
* transformation log 
* visualise action 
* data groups 
* column header data type (can be changed by clicking on it) 
* column header actions 

# Next steps 
* [x] add support for IntroJs
* [x] define introduction points 
* [x] define copy
* [x] Have Georgia improve copy (she has time this week for us) 
* [x] perhaps end by pointing at the `Help` button: If you ever need help, this is where to go.i

Kiarii:
I think with this for now we should prioritize one main item
- the communication around splitting a (flow form) dataset into data groups - to be shown when the user's dataset has more than one data group (else the messaging would be meaningless since there would be no sidebar/datagroup nav)

- The comms around Repeated data group and Derived col indication are secondary and relevant when those two elements are in view..

[...]

PR #2773, opened 2020-07-21
[#2749] - only show intro once

marvinkome:

Using local storage we can track when a user has seen the data group walkthrough so we know not to show the user again. The only limitation is the user sees it If he logs in from another device.

### DIFF
--- a/client/src/components/dataset/DatasetTableV2.jsx
+++ b/client/src/components/dataset/DatasetTableV2.jsx
@@ -81,6 +81,10 @@ function DatasetTable(props) {
   };
 
   useEffect(() => {
+    if (window.localStorage.getItem('done-intro')) {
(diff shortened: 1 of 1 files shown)


tangrammer:
@janagombitova do you agree with the limitation that is described above? 
I agreed with @marvinkome on that but just realised that we forgot to include you in the decission

janagombitova:
I trust your judgement. 

Just to make sure I understand: the user will **not** see the walkthrough for the 2nd time if she:
- Clears her browser history
- Opens a new tab
- Opens a new window
- Opens another Lumen tenant

But **will** see if she:
- Logs in using a different user account
- Logs in on a different device

Is this correct?

tangrammer:
@janagombitova IMO yep!

PR #2781, opened 2020-07-22
[#2127] Add namespace column to dataset_version

tangrammer:
So far dataset_version identity is based on `dataset-id` and `version``

but with RQG approach we'll have `N` dataset_version for each dataset 

`namespace` column is added to identity together with `dataset-id` and `version`

This PR changes the `columns` value as identity prop and proposes `namespace` thinking in backwards compatibility and to don't need querying the `columns` jsonb value to derive which "data-group" is related with, in other words it should simplify the sql query and domain logic 

By default this column will have `main` as default value


tangrammer:
@dlebrero, `ns` is the short version of `namespace`, do you think still we should rename it or just document the PR?

dlebrero:
Either rename or use the long version. I do not have enough domain knowledge to judge if namespace is the best name.

PR #2782, opened 2020-07-23
[#2127] Adapt sql dataset_version queries to main ns column value

tangrammer:
Relates https://github.com/akvo/akvo-lumen/pull/2781

In relation to dataset or dataset-version, remove direct sql references and use clj namespaces for it 

Use default `main` dataset_version.namespace value in updated queries 

### DIFF
--- a/backend/src/akvo/lumen/lib/transformation.sql
+++ b/backend/src/akvo/lumen/lib/transformation.sql
@@ -23,7 +23,7 @@ SELECT id FROM dataset WHERE id = :id
 SELECT id, table_name AS "table-name", imported_table_name AS "imported-table-name", columns, version, transformations
   FROM dataset_version
  WHERE dataset_id = :dataset-id
-   AND ns = :ns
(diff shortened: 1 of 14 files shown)


backend/src/akvo/lumen/lib/transformation.sql
```diff
@@ -18,45 +18,52 @@ UPDATE dataset
...
   FROM dataset_version
  WHERE dataset_id = :dataset-id
+   AND namespace = :namespace
    AND version = (SELECT MAX(v.version)
                     FROM dataset_version v
                    WHERE v.dataset_id = :dataset-id);
 
--- :name latest-dataset-versions-by-dataset-ids :? :*
--- :doc Returns the most recent dataset version for a given dataset id
+-- :name db-latest-dataset-versions-by-dataset-ids :? :*
```
valllllll2000, line 31: Just a doubt I have why is it ids and not id?

Issue #2763, opened 2020-07-15
Dashboard not loading


janagombitova:
From support: https://akvoo.reamaze.com/admin/conversations/akvolumen-support-request-number-366224-from-boroma-at-akvo-dot-org

<img width="1247" alt="Screen Shot 2020-07-15 at 09 03 51" src="https://user-images.githubusercontent.com/12456965/87514757-0b3c4e00-c67b-11ea-92dc-4d175bb8df51.png">
 
On mdi.akvolumen.org the dashboard: "Riposte Au COVID-19 Au Mali - Effets À Court Terme Des Activités De Communication" is not loading in the editor or when shared. On the `Console` I get errors pointing to maps and then a 504.

janagombitova:
More details here: https://docs.google.com/document/d/19TW6KwhqkMi48DCR2u-FMYvrjjq3As7IQYAM8CgHrIE/edit#

dlebrero:
The issue is not the number of visualisations, but the number map visualisations, due to the "stampede" on the number of request to load those maps.

For example, https://demo.akvolumen.org/dashboard/5dfbb7ff-e627-4cfc-afcf-c87021d9ed41 has 35 visualisations, of which 8 are maps. 

Loading the public dashboard, which contains all the data for all visualisation but the maps, takes 500 ms. This is reasonable.

But after loading the dashboard, the 8 maps make ~240 concurrent request to the backend to draw the maps. This is what is causing the DB to choke. Note that this is just for one single user.

I will look at what is the bottleneck on those requests, but it would be good to change the FE code to lazily load the maps as they the user scrolls down the dashboard.

janagombitova:
@marvinkome can you take a look at Dan's suggestion?

dlebrero:
Further looking into this, I think I have been looking at the wrong problem as there seems a bunch of conflated issues.

I see in the Lumen-Maps metrics and logs nothing that could explain why the DB was 100% CPU for several hours.

After broadening my search, I see something very suspicious in the Lumen-Backend usage of the DB connections: 
![Screenshot 2020-07-28 at 09 32 27](https://user-images.githubusercontent.com/2147539/88633979-0b464000-d0b6-11ea-846c-512210b799b8.png)

So basically, mostly for the mdi tenant, sometimes the DB connection is in use for more than 2 hours. My expectation for a healthy service would be of at most a few seconds. 

[...]

PR #2797, opened 2020-08-04
[#2763] Run VACUUM after "apply/undo transformation" and after "update dataset"

tangrammer:

Relates #2763 

#### About updating dataset:
Thus VACUUM can **not** be run inside a transaction **SO FAR** we need to wait for the full dataset update that also processes the transformations related instead of calling VACUUM after every transformation as should be the ideal implementation


tangrammer:
@dlebrero still we have one dataset_version for each dataset 
so we don't need to take care of RQG boundaries yet

Issue #625, opened 2017-03-01
Custom order of items in legend


mtwestra:
### Context
Currently, items in a legend are shown in alphabetical order. However, this does not always make sense, especially with ordinal data types, as shown below.

<img width="200" alt="screen shot 2017-03-01 at 10 02 58" src="https://cloud.githubusercontent.com/assets/478954/23452987/c8c98810-fe66-11e6-9f15-f5fefe9d3aa4.png">

### Proposed solution
Proposed solution is to allow users to change the order of the items in the legend, for example as part of the UI where they specify the colours for each item.

janagombitova:
I have changed the title to not limit the improvement to only Maps. It might be valuable to change the order of items in the legend for any visualisation option. 

In the image Mark shares above ideally the order of things would be from the highest risk to the lowest (or another way around). 

![screenshot 2019-03-04 at 11 40 20](https://user-images.githubusercontent.com/12456965/53728517-3b62c280-3e73-11e9-9d11-a9786a3230e6.png)

In this case, the lowest range should go first and the highest last to make it easier to read the legend and associated colours.

Kiarii:
Based on the assumption that there are a ltd number of categories on which to apply a custom order, we could 
- introduce a legend-category-order-editor where the user opts for a custom order ad can use a drag and drop to arrive at the desired order as illustrated.. 
- determine whether we refer to this as categories or values and word the UI accordingly
![image](https://user-images.githubusercontent.com/12692444/64532644-8e359f00-d31a-11e9-81e4-28b29cc100a4.png)

janagombitova:
@tangrammer I checked all our visualisations where we have a legend and it is useful for all cases: Bar, Bubble, Map, Pie, Polar, Donut, Scatter. 

I propose we start with Pie  (as there is where you started) > push to production under the flag > test internally > once confirmed that all is ok add to the other visualisation types, in the same way, > push to all users without feature flag

janagombitova:
Found one little bug with the current implementation: 

<img width="1250" alt="Screen Shot 2020-08-11 at 13 36 46" src="https://user-images.githubusercontent.com/12456965/89892945-c5ba6480-dbd7-11ea-9fc5-226ff09a514b.png">

It shows also for `polar chart` what is fine, but the order is not applied - so it does not actually work. 

For `pie` and `donut` where it also shows, it works like a charm

tangrammer:
@janagombitova @Kiarii 
[...]

PR #2803, opened 2020-08-07
Add custom order editor functionality to legend items in Bar chart

tangrammer:
- [ ] Add i18n

Relates #625 

- Uses `orderedLegend`  feature-flag
- Add https://github.com/clauderic/react-sortable-hoc and https://github.com/sindresorhus/array-move client libs 
- lumen backend changes only related to specs, just to define the data API we expect in visualisation.spec.legend

```js
{visualisation:
 {spec:
  {legend:
   {order:
    {mode:"custom/auto",
     list: ["v1", "v2", "v3"]}}}}}

```
 
# TODO
- [x] client needs to load legend values 
- [x] client needs to send order selection
- [ ] improve css styles

# Current state

Screenshot 2020-08-10 at 17 12 37


client/src/components/charts/LegendsSortable.jsx
```diff
@@ -0,0 +1,115 @@
...
+  return legend;
+};
+
+export const LegendsSortable = ({
+  onChangeSpec,
+  colors,
+  specLegend,
+  visualisation,
+}) => {
+  const sortable = (get(specLegend, 'order.mode') || 'auto') !== 'auto';
```
muloem, line 64: Why not the positive case here? i.e. 

```suggestion
  const sortable = (get(specLegend, 'order.mode') || 'auto') === 'custom';
```

client/src/components/charts/LegendsSortable.jsx
```diff
@@ -0,0 +1,115 @@
...
+    onChangeSpec({ legend });
+  };
+
+  const getColor = (key, index) => (colors && colors[key]) || palette[index];
+
+  // eslint-disable-next-line new-cap
+  const SortableList = SortableContainer(({ legendItems }) =>
+  (
+    <div style={{ marginBottom: '5px' }}>
+      {legendItems.map((value, index) =>
```
muloem, line 80: The way you have structured this gives the impression that within the same list its possible to find sortable and non-sortable items. Is that the case?

tangrammer: good catch!

tangrammer: this should solve it https://github.com/akvo/akvo-lumen/pull/2803/commits/7f0d069d2fbce9d403154988f61d3e6c657b5556

PR #2806, opened 2020-08-11
Aggregate on OPTION column types

tangrammer:
Before this PR, we had "OPTION" values represented as:
`TEXT` in `client`
`TEXT` in `backend`
`TEXT` in `db`

This PR changes the way "OPTION" is represented in client and backend but db keeps the same
`OPTION` in `client`
`OPTION` in `backend`
`TEXT` in `db`

Changes will be behind feature-flag `optionColumnType` 

Screenshot 2020-08-11 at 14 27 01
Screenshot 2020-08-11 at 14 25 37

- [ ] Update fr, and es i18n client versions

To test it, we need to update the flow dataset or import a new one

### DIFF
--- a/backend/src/akvo/lumen/lib/import/flow_common.clj
+++ b/backend/src/akvo/lumen/lib/import/flow_common.clj
@@ -78,6 +78,7 @@
     "GEOSHAPE" "geoshape"
     "GEO-SHAPE-FEATURES" "multiple"
     "CADDISFLY" "multiple"
+    "OPTION" "option"
     "text"))
 
 (defn questions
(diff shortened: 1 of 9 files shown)


backend/src/akvo/lumen/lib/import/flow_common.clj
```diff
@@ -70,15 +70,21 @@
...
-  [question]
-  (condp = (:type question)
-    "NUMBER" "number"
-    "DATE" "date"
-    "GEO" "geopoint"
-    "GEOSHAPE" "geoshape"
-    "GEO-SHAPE-FEATURES" "multiple"
-    "CADDISFLY" "multiple"
-    "text"))
+  [environment question]
```
tangrammer, line 73: code related to feature-flag

PR #2815, opened 2020-08-13
[#625] Add more legend order support

tangrammer:

Add support for the following types:
- Polar chart
- Horizontal Simple Bar
- Vertical Simple Bar
- Bubble chart

Also needs feature flag 'optionColumnType'


tangrammer:
added bubble chart support

PR #2817, opened 2020-08-17
Add ordered legend support to scatter and map vizs

tangrammer:

Also fulfilling  https://github.com/akvo/akvo-lumen/issues/625#issuecomment-674765942


client/src/components/charts/ScatterChart.jsx
```diff
@@ -160,20 +164,10 @@ class ScatterChart extends Component {
     };
   }
 
-  getColor(category) {
```
tangrammer, line 163: Using same logic as we have in the other vizs

tangrammer:
Last 3 commits in relation to https://github.com/akvo/akvo-lumen/issues/625#issuecomment-674765942

Issue #2847, opened 2020-08-25
Use Kaocha test library


kardan:
# Context
At the moment we use Leiningen as a build tool and eftest (https://github.com/weavejester/eftest) as test runner.

# Problem or idea
Switching to use tools.deps as build system requires changes to the setup. Kaocha (https://github.com/lambdaisland/kaocha) seems to be one of the most active and used modern test runners.

# Solution or next step
Switch to use Kaocha test runner.

PR #2854, opened 2020-08-25
[#2847] Use kaocha for running tests (pt 2)

kardan:
Seperate step for enable so we can disable if needed.
Depends on https://github.com/akvo/akvo-lumen/pull/2853


iperdomo:
Close stalled PR

PR #2857, opened 2020-08-27
Add Google cloud sql

dlebrero:
Adding Google SQL Proxy but not using it yet. 

To use it, we will switch the JDBC url configuration

Part of #2855

### DIFF
--- a/ci/k8s/config-test.yaml
+++ b/ci/k8s/config-test.yaml
@@ -1,0 +1,7 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: akvo-lumen
+  namespace: default
+data:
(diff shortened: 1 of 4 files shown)


ci/k8s/config-test.yaml
```diff
@@ -0,0 +1,7 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: akvo-lumen
+  namespace: default
+data:
+  google-sql-db-instance: shared-test-database
```
tangrammer, line 7: why shared-test-databaset instead of lumen-test-database?

tangrammer (changes requested):
Just send a question related to naming

tangrammer (approved):

Issue #2861, opened 2020-09-01
Adapt transformations to work with N dataset_versions


tangrammer:
Relates #2127 and  https://github.com/akvo/akvo-lumen/pull/2856

## Approach
[Choosing](https://github.com/akvo/akvo-lumen/issues/2804#issuecomment-682384390) easy impl so far, that's to say we store all txs in `main` namespace

## adapt transformations to work with N dataset_versions
Combine transformation example [changes](https://github.com/akvo/akvo-lumen/pull/2856/files) related:
- backend/specs/akvo/lumen/specs/transformation.clj 
- backend/src/akvo/lumen/lib/transformation.clj
- backend/src/akvo/lumen/lib/transformation.sql
- backend/src/akvo/lumen/lib/transformation/combine.clj
- backend/src/akvo/lumen/lib/transformation/combine.sql
- backend/src/akvo/lumen/lib/transformation/engine.clj
- backend/src/akvo/lumen/db/transformation/combine.clj
   
## TODO 
- adapt all transformations types

PR #2878, opened 2020-09-09
[#2861] Rename column.ns to column.namespace

tangrammer:
these changes are included in #2861 but we want to be sure that the client part is ready for it


tangrammer:
@marvinkome could you ensure that the client-side keeps working with this renamed column property?

tangrammer:
@marvinkome in case client side needs changes could you add them to this branch? thanks!!

iperdomo:
Closing as we agreed on a different design

Issue #2880, opened 2020-09-09
Improve impl of getting data-groups 


tangrammer:
# Context
Relates #2858 
Improve impl (add all derived columns in transformations group) and add unit tests for https://github.com/akvo/akvo-lumen/blob/issue-2861/backend/src/akvo/lumen/db/dataset.clj#L42

# Problem or idea
the transformations group doesn't work fine if transformations come from several dataset-version-tables

# Solution or next step
TODO

Issue #2862, opened 2020-09-01
Adapt update process to work with N dataset-versions


tangrammer:
Relates #2127 and  https://github.com/akvo/akvo-lumen/pull/2856

Depends on #2861

# TODO
define problems to solve

PR #2887, opened 2020-09-14
[#2880] Improve fetch dataset 

tangrammer:
relates to #2880
 and also:
 -> [#2861] Add namespace on derived columns
-> [#2862] Fix retrieving namespace from columns


kardan (approved):
LGTM

Issue #2895, opened 2020-09-15
Remove table-name arg from map-config/build


tangrammer:
# Context
https://github.com/akvo/akvo-lumen/blob/master/backend/src/akvo/lumen/lib/visualisation/maps.clj#L126

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

PR #2896, opened 2020-09-15
[#2895] Remove table-name not used arg

tangrammer:

### DIFF
--- a/backend/src/akvo/lumen/lib/visualisation/map_config.clj
+++ b/backend/src/akvo/lumen/lib/visualisation/map_config.clj
@@ -143,7 +143,7 @@
     :else []))
 
 (defn shape-aggregation-sql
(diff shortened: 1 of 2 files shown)


backend/src/akvo/lumen/lib/visualisation/map_config.clj
```diff
@@ -272,26 +272,23 @@
...
 
 (defn get-sql
-  [tenant-conn columns table-name geom-column popup-columns point-color-column
+  [tenant-conn columns geom-column popup-columns point-color-column
    where-clause {:keys [aggregationDataset aggregationColumn
                         aggregationGeomColumn layerType]
                  :as layer} layer-index]
   (cond
 
-    (= layerType "raster")
```
tangrammer, line 281: if raster type we don't call this get-sql  function

kardan (approved):
👍

Issue #2901, opened 2020-09-17
CSV based dataset transformation columns are not shown in the new grid


iperdomo:
# Context
When a user uses the new grid and applies a transformation, that new column is created but is not visible. The user needs to switch to _classic_ grid view to be able to see the column.

# Problem or idea
* _classic_ grid and the new grid must show the derived (transformations) columns

# Solution or next step
?

iperdomo:
Agreed with @marvinkome that we're going to show the _Transformations_ data group in CSV based datasets, only if the dataset has transformations.

PR #2903, opened 2020-09-18
[#2901] - show transformations on CSV datasets

marvinkome:
When doing transformations on CSV based datasets, show the sidebar when there's a transformation group available and remove it when there's no transformations


iperdomo (changes requested):
The behaviour of CSV datasets is correct. Showing a _Transformations_ data group if present, but now sorting of data groups in Flow based dataset is different with the most recent changes.

iperdomo (approved):

Issue #2741, opened 2020-06-29
Allow to select all columns in the Transformations dropdown


janagombitova:
# Context
With supporting RQG data in Lumen (#2127) we changed how we show data in the data grid. We added a navigation option via the data groups (based on the form question groups) (#2705). 

# Issue
With this UI change, when creating a transformation you can now **only** select the columns from the group in view. Though it might seem handy, @Kiarii suggest that from a UX point of view we should not limit the options when transforming data to what is in view in the grid. 

The user should see all the groups and columns in the transformation dropdowns (as it worked before, or as it works for visualisations)

![transformation](https://user-images.githubusercontent.com/12456965/86013992-11e28880-ba20-11ea-9288-a3bdac4ef7a0.gif)

## ...but...

For transformations when 2 or more columns are needed, when Salim selects a column from a RQG, we should allow him to select the second column only from the same group. If he selects the first column from a non-RQG, he should be able to select the 2nd column from any group besides the RQGs. 

The transformations in mind are: 
* Combine column
* Generate geopoint

Merge and reverse geocode will be handled separately. 

# Next steps 
* [x] allow Salim to select from all data groups and columns when using Transformations 
* [x] for `combine column` and `generate geopoints` do not allow to select the combination of a column from a RQG and a non-RQG

janagombitova:
@marvinkome I did some testing on dark-demo and found 2 issues: 

# Data group order not consistent 

I can select any column with the transformations. That is great! But the order of the groups is not consistent with their order in the dataset. The order should not change. 

![order of columns](https://user-images.githubusercontent.com/12456965/86213171-9694ea00-bb79-11ea-87d4-a3b9becad9d7.gif)

# Can only select group column when `combining` or `generating geopoints`

The second change was to create a limitation when using columns from a **repeated** group. The need is that when you `combine columns` or `generate geopoints` you to work with two columns and these two need to have the same amount of rows. So, when one selects a column from a **repeated** group he needs to work with the 2nd column only from this group, as other columns from other groups will not match the number of rows. 

However, now the limitation is implemented on any group. Now, when I go to `combine` or `generate geopoints` I can only select columns from the group that I have open. Regardless if this one is repeated or not. But this should not be limited. I should be able to combine two columns from different groups if both groups are not repeated. 

[...]

PR #2914, opened 2020-09-21
[#2741] - allows to select all column during combine columns and generate geopoints

marvinkome:

### DIFF
--- a/client/src/components/dataset/DatasetTableV2.jsx
+++ b/client/src/components/dataset/DatasetTableV2.jsx
@@ -309,7 +309,7 @@ function DatasetTable(props) {
             hideSidebar();
           });
         },
-        columns: props.columns,
+        columns,
       });
     } else if (menuItem === 'extractMultiple') {
       handleSidebarProps({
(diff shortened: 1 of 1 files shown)


iperdomo:
@marvinkome did you checked the expected logic explained in the issue?

> For transformations when 2 or more columns are needed, when Salim selects a column from a RQG, we should allow him to select the second column only from the same group. If he selects the first column from a non-RQG, he should be able to select the 2nd column from any group besides the RQGs.

For RQG, we filter by columns in that group. For non-RQG all columns in non-RQG data groups.

marvinkome:
I thought the logic is only needed when we have support for RQG?

iperdomo:
> I thought the logic is only needed when we have support for RQG?

That it's an option, but why wait? or are we going to log a different issue to support that logic? It feels that we can fix it right away? What do you think?

marvinkome:
Right now we don't know if it's a RQG or not

janagombitova:
Just chipping in. I agree with the notion `why wait`. But if we are waiting because more work needs to be done to know which group is repeated (as we do not have that data yet) then I suggest we push out this change. And in another effort ensure we know which group is repeated and enforce the limitation for RQGs.

janagombitova:
@marvinkome and @iperdomo what is the status of this change?

PR #2929, opened 2020-09-24
[#2928] Adapt caddisfly count test expectations

### LINKED ISSUE #2928: Caddisfly schema errors in production
# Context
We have added a new Caddisfly test, we need to adapt our expectations. We have now 114 tests.

# Problem or idea
Error in the logs

# Solution or next step
* Change the log to `warning`
* Adapt the number of expected tests

tangrammer:
Closes #2928


iperdomo (changes requested):
Tests failing

Issue #2945, opened 2020-10-05
Cannot change colours on scatter plots 


janagombitova:
https://demo.akvolumen.org/visualisation/5f7ac727-381d-4b21-9384-7c28ea52a5bd

Steps to take to recreate: 
1. Select to create a new scatter plot. 
2. Define the x and y axis. 
3. Click on a point to change colour --> **does not work**
4. Define the category bucket - the points are now coloured based on the values. 
5. Click on a point to change colour --> **does not work**

janagombitova:
@marvinkome what is the status of this issue?

marvinkome:
It works when you select the category bucket column. But to support changing the default blue requires API changes. And also there's a bug when you create a new scatter plot the colors are not consistent with the label config. This is due to the way the scatter plot is created. I'd have to refactor for it to work.

janagombitova:
Thanks Marvin for the update. Let's keep the issue open then for now and see how to plan things in once we finish the work on RQGs

tangrammer:
Parking the issue a while until we found time to refactor this client related code 

Relates to https://github.com/akvo/akvo-lumen/issues/2979#issue-724525082

PR #2948, opened 2020-10-06
[#2945] - fix legend color on scatter plots

marvinkome:
this only fix when bucket column is selected

### DIFF
--- a/client/src/components/charts/ScatterChart.jsx
+++ b/client/src/components/charts/ScatterChart.jsx
@@ -167,12 +167,17 @@ class ScatterChart extends Component {
     if (!itsSet(category)) {
       return color;
     }
+
+    // if color is selected
     if (colorMapping[category]) {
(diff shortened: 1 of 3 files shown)


client/src/components/charts/ScatterChart.jsx
```diff
@@ -249,6 +254,7 @@ class ScatterChart extends Component {
   handleShowColorPicker(isPickingColor) {
     const { interactive, print, edit } = this.props;
     if (!interactive || print) return;
+    console.log(edit, isPickingColor);
```
tangrammer, line 257: remove logs

tangrammer:
@marvinkome there is an inconsistency between the chart and the chart config

(look at the colors 👀 
Screenshot 2020-10-07 at 13 32 23
)

tangrammer:
can't reproduce again the problem, merging now

PR #2953, opened 2020-10-08
[#2952] Guard against a potential null value

### LINKED ISSUE #2952: Lumen dashboard not loading due to pie charts
An issue reported via support. This dashboard: https://pipep-rca.akvolumen.org/s/Ojje88GFLBA loads and then disappears. From the console it seems that some pie charts are causing the problem. 

Screen Shot 2020-10-08 at 12 03 50

iperdomo:
Closes #2952


kardan (approved):
LGTM

Issue #2954, opened 2020-10-08
Removing a dashboard filter makes an invalid GET request


iperdomo:
# Context
When removing a filter from a Dashboard definition, the client makes an invalid GET request with a `null` path leading to an HTTP 500 response.

# Problem or idea
![Peek 2020-10-08 17-01](https://user-images.githubusercontent.com/178474/95477124-4b645180-0988-11eb-9cb1-4e214790fd4c.gif)

# Solution or next step
* Verify from where that `null` value is coming, make sure that we don't send invalid requests

janagombitova:
@iperdomo can we close this issue?

marvinkome:
Yes

PR #2956, opened 2020-10-09
[#2954] - remove invalid fetchColumn call in dashboard filters

marvinkome:
Removing the fetchColumn call when a dashboard filter is removed, because it's being called with a null value


tangrammer (approved):
good catch!

PR #2962, opened 2020-10-13
[#1098] Setup local-admin using an @akvo.org email address

iperdomo:
The scripts creates a new "Local Admin"  user in Keycloak, with the email address passed as argument. It configures the user to be *admin* of `t1` tenant.

Usage:

      ./backend/dev_helpers/local-admin.sh "<email>"

Example:

    ./backend/dev_helpers/local-admin.sh "bob@akvo.org"

### DIFF
--- a/backend/dev_helpers/local-admin.sh
+++ b/backend/dev_helpers/local-admin.sh
@@ -82,7 +82,7 @@ cat <<EOF > /tmp/new-user.json
 EOF
 
 email="${1}"
-sed -i "s/xxxxx/${email}/" /tmp/new-user.json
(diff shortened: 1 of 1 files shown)


iperdomo:
@kardan @tangrammer it would be nice to test on macOS ... i'm not sure about the `sed -i` call ... i think in macOS there is need for an extra parameter

iperdomo:
```
docker-compose up --force-recreate --no-deps --detach keycloak  # force recreate container and remove local state

./backend/dev_helpers/local-admin.sh "<email>@akvo.org"         # change <email> with yours

```

tangrammer (approved):
works like a breeze in osx, only had to `chmod +x local-admin.sh`

PR #2965, opened 2020-10-14
Issue/2925 data group query

iperdomo:
* Support for bar charts aggregation with multiple *iterations*

### DIFF
--- a/backend/src/akvo/lumen/lib/aggregation/commons.clj
+++ b/backend/src/akvo/lumen/lib/aggregation/commons.clj
@@ -73,16 +73,16 @@
      :from (mapv :table-name adapted-dgs)}))
 
 (defn data-groups-sql
-  [template]
+  [{:keys [select from]}]
   (str/join " "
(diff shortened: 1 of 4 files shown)


backend/src/akvo/lumen/lib/aggregation/commons.clj
```diff
@@ -50,3 +51,40 @@
...
+                                                   (or (= "instance_id" columnName)
+                                                       (= "rnum" columnName)))
+                                                 columns))}))
+                     []
+                     data-groups)]
+    {:select (mapv :columnName (flatten (map :columns adapted-dgs)))
+     :from (mapv :table-name adapted-dgs)}))
+
+(defn data-groups-sql
+  [template]
```
kardan, line 76: Is this really a template? Is it not the other way around that this function that kinda is the template and that what we pass in is some kind of "data"?

kardan (approved):
LGTM

Issue #2964, opened 2020-10-14
Filtering out empty rows in visualisations not working 


janagombitova:
![Lumen bug filter](https://user-images.githubusercontent.com/12456965/95955665-7c67da80-0dfd-11eb-8b25-b10bd77b558f.gif)

Irene found a bug in the visualisations. It seems that the filters work if you specify a value, but not when you want to filter out the empty rows 
https://flowhelp.reamaze.com/admin/conversations/akvolumen-support-request-number-410009-from-irene-at-akvo-dot-org

janagombitova:
Tested. Released.

PR #2966, opened 2020-10-14
[#2964] Filter isEmpty option column as text type

tangrammer:
Option column types are stored as text, so empty value in option column types means "" instead of NULL

Relates #2964


kardan (approved):
Approved, but a test would have got you a gold star!

PR #2967, opened 2020-10-14
Issue/2707 nice tooltips

martinchristov:


client/src/components/dataset/ColumnHeader.jsx
```diff
@@ -119,11 +129,43 @@ export default class ColumnHeader extends Component {
...
+const ConditionalTooltip = ({ children }) => {
+  const [useTooltip, setUseTooltip] = useState(false); // eslint-disable-line
+  const handleRef = (ref) => {
+    if (!ref) return;
+    if (ref.parentElement.offsetWidth < ref.offsetWidth + 40 && !useTooltip) {
+      setUseTooltip(true);
+    } else if (ref.parentElement.offsetWidth >= ref.offsetWidth + 40 && useTooltip) {
+      setUseTooltip(false);
+    }
+  };
```
marvinkome, line 141: Why not `useRef`?

martinchristov: There is no need for it in this case. A ref handler function is sufficient.

marvinkome: Okay 👍🏾

Issue #2950, opened 2020-10-07
Limit combine columns and generate geopoints to work with repeated data groups 


janagombitova:
This issue is a follow up of: #2741 

When working with data **transformations** you normally generate a new column out of **one** original column, or you adjust the values in one column. But we have two transformations currently where **two columns** are used: 
* Combine columns 
* Generate geopoints 

When you combine columns or generate geopoints you to work with two columns and these two need to have the same amount of rows. This becomes tricky when a dataset holds **repeated data groups** as the row count differs. So, when one selects a column from a repeated group he needs to work with the 2nd column only from this group, as other columns from other groups will not match the number of rows.

The expectation is: 
* I should be able to combine two columns from different groups if both groups are not repeated.
* If I select the 1st column from a repeated group, I can select the 2nd column only from the same repeated group 
* If I select the 1st column from a non-repeated group, I should not be able to select the 2nd column from any existing repeated group, but I can select from any non-repeated group.

janagombitova:
Can we close this issue?

PR #2968, opened 2020-10-14
[#2950] - Limit columns when working with RQG

marvinkome:
When doing Combine columns or generate geopoint transformations,
This commit add the columns logic when working with RQG or not.
The logic can be found in the connected issue


client/src/components/dataset/sidebars/GenerateGeopoints.jsx
```diff
@@ -54,6 +55,27 @@ export default class GenerateGeopoints extends Component {
     this.handleChangeColumnTitleGeo = this.handleChangeColumnTitleGeo.bind(this);
   }
 
+  getColumns(latOrLong) {
```
iperdomo, line 58: I guess this logic can be unit tested? it is possible?

marvinkome: I'd have to extract it out to a separate function to unit test it. Or else I'd be testing the whole component together

client/src/utilities/column.js
```diff
@@ -97,3 +97,22 @@ export const columnSelectOptions = (intl, columns) => {
   const res = extractColumnOptions(columns);
   return isImmutable(res) ? res.toArray() : res;
 };
+
+// For combine columns and generate geopoints column select
```
tangrammer, line 101: will we need this comment on mid-long term?

client/src/utilities/column.js
```diff
@@ -97,3 +97,22 @@ export const columnSelectOptions = (intl, columns) => {
...
+export const columnBasedOnOppositeColumn = (columns, oppositeColumnName) => {
```
tangrammer, line 102: perhaps we could find a better name for this function 🤔 . The logic inside doesn't seem to be represented by the name
My proposal and refactor will be:
```js
export const filterColumnsByDataGroupDimension = (columns, columnName) => {
  // if column is not selected the return all columns
  if (!columnName) {
    return columns;
  }

  const column = columns.find(col => col.get('columnName') === columnName);

  // if column is not repeatable return all not repeatable columns
  if (!column.get('repeatable')) {
    return columns.filter(col => !col.get('repeatable'));
  }

  // if column is repeatable, return only columns from the same group
  return columns.filter(col => col.get('groupName') === column.get('groupName'));
};

```

tangrammer (changes requested):
do you aggree with my proposal
refactor?

tangrammer (approved):

PR #2972, opened 2020-10-15
Issue/2805 show dataset size

martinchristov:


marvinkome:
Did you test this across visualizations and dashboard headers? They both use the Entity header

martinchristov:
@marvinkome thanks for the reminder. Indeed I had not tested but I did so now and it all seems good for both Dashboards & Visualisations

marvinkome (approved):

Issue #2707, opened 2020-05-27
Change data type indication from text to icons


janagombitova:
# Context

As part of the work to support RQGs in Lumen we decided to improve the dataset UI. To make the change doable, small and iterative, this issue handles only one small piece of the puzzle and can be addressed and released independently from the other changes.

## Why do we have this issue? What are we trying to solve? 
To simplify to UI and clean it up a bit, easily distinguish between a standard element and a data value. 
 
## How will this benefit the users & Akvo? 
* cleans up the dataset UI making it more appealing and easy to orientate in 

Plus as this change is part of a larger UI improvement of the dataset editor to support RQGs so we can assume that the entire change will:

* make Lumen more attractive for Flow users that have RQG data and thus strengthen our value (as they can make use of the existing Lumen features as transformation log, easy transformations, custom visuals, direct Flow update, etc)
* by making it easy to work with RQG data we strengthen our market position as competitors do not have an easy way to handle the data

# Opportunity 

## Current status quo
* We have these datatypes now:
   * Number
   * Text
   * Date
   * Multiple
   * Geopoint
   * Geoshape

These are indicated as text next to the column header. 

<img width="1203" alt="Screen Shot 2020-05-27 at 14 43 09" src="https://user-images.githubusercontent.com/12456965/83030811-1b995a80-a034-11ea-8df2-aa47e11af061.png">

## The idea 

The idea is to change all datatypes to icons. On mouseover, the data type should be shown. On clicking on the type icon, the user should have the option to change the data type, as it works today.  
<img width="878" alt="Screen Shot 2020-05-27 at 13 07 45 2" src="https://user-images.githubusercontent.com/12456965/83030825-1f2ce180-a034-11ea-8d8a-0f521fae433c.png">

# Next steps  
* Use the icons as shown in the mockups 
* Kiarii can provide icons where not available

janagombitova:
Just to update the current status: We have these datatypes now:
* Number
* Text
* Date
* Multiple
* Geopoint
* Geoshape
* Option

Kiarii:
@janagombitova aren't Option and Multiple are Text data types?

janagombitova:
@Kiarii In theory they are, but in practice they are automatically given to Flow option questions, geoshapes and Caddisfly and specific transformations only work for them. So they need to be shown as Option and Multiple so Salim see why Lumen behaves differently with these columns.

Kiarii:
[...]

PR #2975, opened 2020-10-16
[#2707] Require svg images and adapt "multiple" i18n

tangrammer:

no more nginx config needed just adding the required webpack & jsx logic to include and read the files

### DIFF
--- a/client/src/translations/en.json
+++ b/client/src/translations/en.json
@@ -222,7 +222,7 @@
   "min_value": "Min: {val}",
   "min": "Min",
   "misc": "Other options",
-  "multiple": "Multiple",
+  "multiple": "Multiple values",
   "must_choose_bucket_column_first": "Must choose bucket column first",
   "must_choose_x_axis_column_first": "Must choose X-axis column first",
   "name": "Name",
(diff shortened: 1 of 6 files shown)


dlebrero (changes requested):
The problem is that the images are not in the container:

```
Daniels-Mac-mini:akvo-authorization dlebrero$ for i in `kubectl get pods -l "app=lumen" -o jsonpath="{.items[*].metadata.name}"`; do kubectl exec -c lumen-client $i find . | grep type-text.svg ; done
Daniels-Mac-mini:akvo-authorization dlebrero$
```

The nginx config is not required as the "location / {" is already doing the same.

tangrammer:
So adding another fix approach

marvinkome (approved):
Awesome

client/src/translations/en.json
```diff
@@ -222,7 +222,7 @@
   "min_value": "Min: {val}",
   "min": "Min",
   "misc": "Other options",
-  "multiple": "Multiple",
+  "multiple": "Multiple values",
```
kardan, line 225: I suggest we change the key to "multiple_values".

kardan: Was meant as a suggestion, I could still merge the fix if we want it this Friday afternoon!

tangrammer:
@kardan not so straightforward to change the i18n key thus we use the column type to point it, so to make this change it'll imply to create a new i18n key for each backend column key eg: multiple, multiple_type, text, text_type .... 🤔 so merging now the PR and we can think better on this improvement in a different issue

PR #2994, opened 2020-10-26
[#2988] - set a 500ms timeout before making changes in visualization

### LINKED ISSUE #2988: Attempt to immediately persist changes in the backend leads to errors
# Context
When a user is editing a visualization, we don't need to *immediately* persist the changes in the backend. 

Peek 2020-10-22 16-54

# Problem or idea
We send a `PUT` request immediately after every key stroke when changing labels. I would guess that the behaviour is similar in other parts of the application.

# Solution or next step

We need to *buffer* changes and send fewer requests after some time of user inactivity.

### LINKED ISSUE #2990: A transient error when persisting leads to inconsistent state in the client
# Context

This is related to #2988 .

When we attempt to save all the changes in the backend and we have a *transient* error (e.g. one of the `PUT` request fails), the client reaches an inconsistent state, even if a subsequent request succeeds.

The user visible error is a browser alert: *You have unsaved changes, are you sure you want to leave this page?*
2020-10-26_1256x727

In the screenshot you see one HTTP 500 in the middle of several HTTP 200

# Problem or idea

* Attempt to save too many changes at once could lead to *some* HTTP requests errors
* After a successful save the client should *clear* that error state

# Solution or next step
* Clear the error state after a successful save to the backend

marvinkome:
Using the lodash.debouce function. I made the visualizationSpec save
callback only run after 500ms

Closes #2988
Closes #2990 

### DIFF
--- a/client/e2e-test/cypress/integration/index.spec.js
+++ b/client/e2e-test/cypress/integration/index.spec.js
@@ -50,10 +50,14 @@ describe('Akvo Lumen e2e tests', () => {
         force: true,
       });
 
+      cy.waitForAutoSave();
+
       cy.get('[data-test-id="dataset-menu"] .SelectMenu input')
         .first()
         .click({ force: true })
         .type(`${datasetName}{enter}`, { force: true });
+
+      cy.waitForAutoSave();
 
       // open column menu
       cy.get('[data-test-id="column-select"] .SelectMenu input', {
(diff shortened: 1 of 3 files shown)


iperdomo (changes requested):
@marvinkome the builds are failing. I tried to *rerun* them and still fail. Not sure if an actual problem or not.

marvinkome:
Yeah, the problem doesn't seem related to the changes. I'm trying it locally now

iperdomo (approved):

Issue #3002, opened 2020-10-27
Existing visualisations do not show selected data in dropdowns


janagombitova:
All existing visualisations do not show the values selected in the dropdowns. Only bar charts work fine

Here is a pivot table example: 
<img width="1215" alt="Screen Shot 2020-10-27 at 15 20 06" src="https://user-images.githubusercontent.com/12456965/97314354-e9508b00-1867-11eb-9faa-af4a339369cb.png">
https://demo.akvolumen.org/visualisation/5f7abf61-0526-492a-b8f2-7db566cc71c5

When I open an existing pivot table the dropdowns for columns and rows are empty despite having the pivot table generated. This happens for existing visualisation of this type: 
* Pivot table
* Bubble 
* Scatter
* Line / Area 
* Pie / Donut
* Map

iperdomo:
Together with @tangrammer and @marvinkome we diagnosed and discovered the problem. Basically we *forgot* to update some of the `Select` components (the ones affected) with some new functionality that is *aware* of the changes in the internal structure of columns definition.

As commented with @marvinkome this is another symptom of the change we introduced in the columns definition structure and lead  to not be able to apply transformations because the drop-downs where empty.

IMHO the root case of this is the *API break* (the contract between the Web client and the backend) we introduced with that columns definition structure.

marvinkome:
I wouldn't say the API changed caused this bug. The `columnSelectOptions` function changes the shape of the `options` props. But I didn't change the `value` prop to match this new shape.

iperdomo:
> I wouldn't say the API changed caused this bug. The columnSelectOptions function changes the shape of the options props. But I didn't change the value prop to match this new shape.

I'm not talking about this particular case. My point is that the change of the initial data structure `[{col1}, {col2}, ..., {coln}]` to `["group1", [{col1-1}, ..., {col-1-n}], "group2", [{col2-1}, ..., {col2-n}]]` has been the source of **many** bugs.

marvinkome:
Closing this. Reopen if bug shows up again

PR #3004, opened 2020-10-28
[#3002] - show selected option in data groups dropdown

marvinkome:

### DIFF
--- a/client/src/translations/en.json
+++ b/client/src/translations/en.json
@@ -255,6 +255,7 @@
   "pdf": "PDF",
   "pending": "Pending",
   "percentage_stack_bars": "Percentage Stack Bars",
+  "pick_a_default_value": "Pick a default value",
   "pie": "Pie",
   "pivot table": "Pivot table",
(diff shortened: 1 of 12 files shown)


client/src/translations/en.json
```diff
@@ -255,6 +255,7 @@
   "pdf": "PDF",
   "pending": "Pending",
   "percentage_stack_bars": "Percentage Stack Bars",
+  "pick_a_default_value": "Pick a default value",
```
iperdomo, line 258: When adding a new translation key, we need to take care of the `es.json` and `fr.json` files too

marvinkome: What're the translations for Spanish and french?

kardan: I have started to add keys: use Google translate to get something (that hopefully will trigger a native speaker to react) and then I assign someone that knows the languages as reviewers. Not sure they love this practice but.. seems to work :-)

iperdomo (approved):

PR #3016, opened 2020-11-02
Issue 2963/hide introjs in visualization editor

marvinkome:

- Stop showing introjs step in visualization
- Render introjs only once

### DIFF
--- a/client/src/components/dataset/DatasetTableV2.jsx
+++ b/client/src/components/dataset/DatasetTableV2.jsx
@@ -28,16 +28,106 @@ function formatCellValue(type, value) {
   }
 }
 
+function useIntroJs(props, isMounted, sidebarProps) {
+  const introStarted = useRef(false);
+
+  useEffect(() => {
+    if (window.localStorage.getItem('useDataGroupsIntroDone')) {
+      return undefined;
+    }
+
+    const intro = introJs();
+
+    intro.onbeforechange(() => {
+      const currentStepIdx = intro._currentStep;
+      const currentStepDynamic = !!intro._options.steps[currentStepIdx].dynamic;
+
+      if (currentStepDynamic) {
+        const step = intro._options.steps[currentStepIdx];
+        const element = document.querySelector(step.element);
+
+        if (element) {
+          const introItem = intro._introItems[currentStepIdx];
+          introItem.element = element;
+          introItem.position = step.position;
+        }
+      }
+    });
+
+    intro.onexit(() => {
+      window.localStorage.setItem('useDataGroupsIntroDone', true);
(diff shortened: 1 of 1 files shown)


tangrammer:
@marvinkome could you point the lines changed that are related to:

1- Stop showing introjs step in visualization
2 -Render introjs only once

In relation to point 1, does it mean that if a user clicks in this link (https://lumen.akvotest.org/visualisation/5fa11b66-bb83-4581-8c45-4c85bbc8ab32) is not going to have the introjs? (so trying to edit directly the viz without going through library step/screen )

tangrammer:
PS: I like the new hook design! 🚀 
PS2: do we really need these lines https://github.com/akvo/akvo-lumen/blob/af9b8a0e21a3659f6a93d1aacc238dfa9e592f50/client/src/components/dataset/DatasetTableV2.jsx#L41-L55 ? they looks like very introjs internal details 🤔

marvinkome:
> @marvinkome could you point the lines changed that are related to:
> 
> 1- Stop showing introjs step in visualization
> 2 -Render introjs only once
> 
> In relation to point 1, does it mean that if a user clicks in this link (https://lumen.akvotest.org/visualisation/5fa11b66-bb83-4581-8c45-4c85bbc8ab32) is not going to have the introjs? (so trying to edit directly the viz without going through library step/screen )

Introjs was showing in visualization before it was called twice. So when you close the first one (it destroys the first instance of introjs, but other instances are still loaded). So when you go to visualization screen and press ENTER, it continues the intro of other instances

marvinkome:
> PS: I like the new hook design! 🚀
> PS2: do we really need these lines?
> 
> https://github.com/akvo/akvo-lumen/blob/af9b8a0e21a3659f6a93d1aacc238dfa9e592f50/client/src/components/dataset/DatasetTableV2.jsx#L41-L55
> 
> they looks like very introjs internal details 🤔

Yes. IntroJS doesn't work with React out of the box. So with the onbeforechange we dynamically point to the element when it's present in the DOM

tangrammer:
> > PS: I like the new hook design! 🚀
> > PS2: do we really need these lines?
> > https://github.com/akvo/akvo-lumen/blob/af9b8a0e21a3659f6a93d1aacc238dfa9e592f50/client/src/components/dataset/DatasetTableV2.jsx#L41-L55
> > 
> > they looks like very introjs internal details 🤔
> 
> Yes. IntroJS doesn't work with React out of the box. So with the onbeforechange we dynamically point to the element when it's present in the DOM

then why don't we move this logic to a proper file instead of being inside DatasetTableV2 ?

Issue #3026, opened 2020-11-04
Add tests to maps


kardan:
# Context
We don't have much tests around maps. 

# Problem or idea
We're about to add maps from the new dataset version 2 implementation. Would be great to have tests to get some confidence we don't break anything.

# Solution or next step
Write tests.

janagombitova:
@kardan what is the status of this one?

PR #3041, opened 2020-11-11
[#3026] Enhance generated Flow dataset with random geopoints

kardan:
To test, run dev/new-flow-dataset at the repl.

```Clojure
dev> (new-flow-dataset "Flow dataset 1")
"5fab8fb5-c1f7-43c0-b9af-d33bfee9fa9f"
```

This should generate a new dataset with a geopoint column in the "Non repeatable Group". This can be used to create a map from.


tangrammer (approved):
wow! happy to see that you are getting familiar with specs and generators

PR #3046, opened 2020-11-11
[#3045] Hide `Instance id` column from non-metadata groups

### LINKED ISSUE #3045: Hide `Instance id` column from non metadata groups
# Context
The `instance_id` column (title: "Instance id") is available in all data groups. After a conversation with @janagombitova, @Kiarii and other team members, we have decided to *hide* that column from non-metadata groups.

# Problem or idea
The `Instance id` column is a *special* column that serves as *link* (internally a foreign key) for all data groups to the *parent* one (metadata). The user must not be able to *transform* this column. Initially we're going to hide it.

# Solution or next step
- Hide the column in non-metadata groups
- Collect feedback from users about the need/no-need of having that columns available in all data groups

iperdomo:
The `Instance id` column is a special column that serves as link
(internally a foreign key) for all data groups to the parent one
(metadata). The user must not be able to transform this
column. Initially we're going to hide it.

Since the column already has a property `hidden` we just use it to
include or not the column.

Closes #3045

### DIFF
--- a/client/src/containers/Dataset.jsx
+++ b/client/src/containers/Dataset.jsx
@@ -297,6 +297,7 @@ function Dataset(props) {
     return <LoadingSpinner />;
   }
 
+
   return (
     <NavigationPrompt
       shouldPrompt={pendingSaving.savingFailed}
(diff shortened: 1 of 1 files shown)


tangrammer (changes requested):
did you check that hidden columns don't appear in dropdowns?

iperdomo:
> did you check that hidden columns don't appear in dropdowns?

I didn't, the change was to *hide* the column from the data-grid so the user is not able to transform it

iperdomo:
Closing in favour of filtering columns in the backend

Issue #3036, opened 2020-11-09
Support for 'to uppercase' transformation using data groups


kardan:
# Context
To support repeated question groups, on the backend we have a new version of datasets called "dataset_version_2" which includes the concepts of "data groups". We need to add the transformation facilities to the new dataset_version_2.

# Problem or idea
It's a deep subject and we need to get a proper understanding of the problems. 

Some random issues might be:
1. Old Flow data without namespaces
2. How do we handle derive transformations?
3. How do we handle combine columns

# Solution or next step
How could we solve it?

kardan:
From @iperdomo on Zulip https://docs.google.com/spreadsheets/d/1HiwOe8Io7vnuN62yNi_g7NYyzvJuBufWBv6OAP8KLls/edit#gid=1409395453

iperdomo:
> It's a deep subject and we need to get a proper understanding of the problems.
> 
> Some random issues might be:
>
>    Old Flow data without namespaces
>    How do we handle derive transformations?
>    How do we handle combine columns

We have changed the scope of this issue and we're only going to deal with one transformation **without** thinking in backward compatibility.

tangrammer:
we need to take care of vacuum if data-groups https://github.com/akvo/akvo-lumen/blob/issue%2F3036-to-uppercase/backend/src/akvo/lumen/lib/transformation.clj#L54

iperdomo:
* As part of this changes we removed the `repeatable` attribute from columns definition. There is still client code that relies on this property
* We want to filter out `hidden` columns from the client - See #3045

janagombitova:
What is the status of this issue?

iperdomo:
As described in one of the comments, we support `core/to-uppercase` without backward compatibility.
I'm going to close this issue and open a new one when we tackle the backward compatibility.

PR #3047, opened 2020-11-12
[#3045] Hide `Instance id` in the backend

### LINKED ISSUE #3045: Hide `Instance id` column from non metadata groups

iperdomo:
We have explored the idea of hiding the the `Instance id` for
non-metadata groups in the client. The result is that some code uses
the Redux store (a central place for state) but other parts, like the
Grid, uses the direct response from the XHR call.

Hiding only visually in the Grid and Dropdowns makes us wonder how the
grid will behave if the server is responding more data than the
visible one. How this will play out when transformations are added.

For now it seems *safer* to filter the `hidden` columns in the
backend.

Closes #3045 #3036

### DIFF
--- a/backend/src/akvo/lumen/lib/dataset.clj
+++ b/backend/src/akvo/lumen/lib/dataset.clj
@@ -98,7 +98,11 @@
       (let [dataset* (-> (select-keys (first data-groups)
                                       [:name :modified :created :updated :source :transformations])
                          (assoc :status "OK"))]
-        (lib/ok (assoc dataset*  :id id :groups (mapv (juxt :id :columns :repeatable) data-groups))))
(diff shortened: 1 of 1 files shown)


kardan (approved):
LGTM

PR #3058, opened 2020-11-27
[#3036] Add support for transformations that produce derived column(s)

tangrammer:


iperdomo:
The PR looks fine, just for my understanding. I didn't had to do *anything* about columns definition to support `core/combine` locally. Just enable the transformation and make sure we pick 2 columns from the same data group. 😕

tangrammer:
> The PR looks fine, just for my understanding. I didn't had to do _anything_ about columns definition to support `core/combine` locally. Just enable the transformation and make sure we pick 2 columns from the same data group. 😕

yep, but before this PR the code couldn't create unique derivation column names across all data-groups

iperdomo (approved):

PR #3060, opened 2020-11-30
[#3036] Preserve columns order after transformation

tangrammer:

Using set/difference we lost the original columns order

previous result (derived column is "test combine column")
Screenshot 2020-11-30 at 10 57 15

result with this PR changes (order is preserved and new derived column is places at the end)
Screenshot 2020-11-30 at 11 47 13


kardan (approved):
LGTM

PR #3062, opened 2020-11-30
Add data-group derive and reverse-geocode transformation support

tangrammer:


kardan:
Meh, was going to add my review but by trigger happy finger merged. But it was the end goal anyway, so all fine.

Issue #3061, opened 2020-11-30
Columns included in a transformation with data groups can only belong to one data-group


tangrammer:
# Context
Relates #3036

We should restrict columns included in the transformation payload also in `client` side 

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

tangrammer:
merged backend side

janagombitova:
Great! I live the error I get when I try to combine two columns from different groups. 

But indeed we should also limit on the client side. So when Salim selects one column, in the 2nd dropdown only columns from that group should show. For derive javascript transformation I guess there is nothing to do on the client side at the moment, right?

tangrammer:
> But indeed we should also limit on the client side. So when Salim selects one column, in the 2nd dropdown only columns from that group should show. 
ok!

> For derive javascript transformation I guess there is nothing to do on the client side at the moment, right?
yep, although we could do some tricky thing (reuse backend logic in client side) it would be fairly more complicated

PR #3070, opened 2020-12-01
[#3061] Restrict dropdown select columns to current data-groups in client side

tangrammer:
UI parts to be accomplished:

- [x] combine-columns transformation
- [ ] extract-multiple transformation
- [ ] split-column transformation
- [ ] derive-category transformation
- [ ] merge-datasets transformation
- [ ] generate-geopoints transformation
- [ ] reverse-geocode transformation

### DIFF
--- a/client/src/components/dataset/DataTableSidebar.jsx
+++ b/client/src/components/dataset/DataTableSidebar.jsx
@@ -57,5 +57,7 @@ DataTableSidebar.propTypes = {
     ]
   ).isRequired,
   onClose: PropTypes.func.isRequired,
+  env: PropTypes.object,
+  group: PropTypes.object,
   intl: intlShape,
 };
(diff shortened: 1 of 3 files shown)


janagombitova:
@tangrammer why are we making this restriction? 

I believe we said we should **restrict from one data group** but not from the data group in view. 

We said this restriction should only apply on transformations that need more than one column. Some of the transformations you listed need just one column. For example with the `derive category column` there is no data group in view as we bring the user to a new screen. 

Secondly, I might know already what I am going to do with the dataset once I open it, so I go directly to the transformation I need. But with this restriction, I will first have to go to my data group and then to the transformation. 

And last but not least, I remember @Kiarii and Marvin implemented / were talking about implementing a change where the data group in view was shown in the dropdown of the transformation 'first' (auto scrolled to). This way the user sees the content she is on, but she is not limited to what she can work with as it would require another action from her.

tangrammer:
perfectly clear!

PR #3071, opened 2020-12-02
[#3036] Fix reverse-geocode transformation with data-groups

tangrammer:

Use same approach that we use in merge-dataset transformation


tangrammer:
closing PR thus reverse geocode source columns will be always in the same data-group

PR #3078, opened 2020-12-12
Issue/3026 map test restart

kardan:
So we have something when we start with data groups for maps. Though there were issues with the builds, hence re-ran the builds a couple of times.

### DIFF
--- a/backend/specs/akvo/lumen/specs/import/values.clj
+++ b/backend/specs/akvo/lumen/specs/import/values.clj
@@ -1,9 +1,17 @@
 (ns akvo.lumen.specs.import.values
   (:require [clojure.spec.alpha :as s]
             [akvo.lumen.util :refer (squuid)]
-            [akvo.lumen.specs :as lumen.s :refer (sample)]
+            [akvo.lumen.specs :as lumen.s]
             [akvo.lumen.lib.import.csv :as csv]))
 
+(defn round-coordinate
+  "Round coordinate to 6 digits but go by bigdec rounding to remove trailing
+   zeros. This to be able to compare facts to values from Postgis"
+  [coordinate]
+  (-> coordinate
+      bigdec
+      (.setScale 6 java.math.RoundingMode/HALF_EVEN)
+      double))
 
 (defn rand-lng
   []
@@ -19,7 +27,8 @@
 (defn rand-point
   "Random geopoint with 6 decimals resolution"
   []
-  (format "POINT(%.6f %.6f)" (rand-lng) (rand-lat)))
+  (format "POINT(%s %s)"
+          (round-coordinate (rand-lng)) (round-coordinate (rand-lat))))
 
 (def point "POINT(10 10)")
 
--- a/backend/specs/akvo/lumen/specs/import/values.clj
+++ b/backend/specs/akvo/lumen/specs/import/values.clj
@@ -2,7 +2,8 @@
   (:require [clojure.spec.alpha :as s]
             [akvo.lumen.util :refer (squuid)]
             [akvo.lumen.specs :as lumen.s]
-            [akvo.lumen.lib.import.csv :as csv]))
+            [akvo.lumen.lib.import.csv :as csv])
+  (:import java.math.RoundingMode))
 
 (defn round-coordinate
   "Round coordinate to 6 digits but go by bigdec rounding to remove trailing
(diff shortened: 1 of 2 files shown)


backend/specs/akvo/lumen/specs/import/values.clj
```diff
@@ -1,9 +1,17 @@
...
+            [akvo.lumen.specs :as lumen.s]
             [akvo.lumen.lib.import.csv :as csv]))
 
+(defn round-coordinate
+  "Round coordinate to 6 digits but go by bigdec rounding to remove trailing
+   zeros. This to be able to compare facts to values from Postgis"
+  [coordinate]
+  (-> coordinate
+      bigdec
+      (.setScale 6 java.math.RoundingMode/HALF_EVEN)
```
iperdomo, line 13: we can `(: import java.math.RoundingMode)` in the namespace declaration and the constant reads shorter? `RoundingMode/HALF_EVEN`

backend/test/akvo/lumen/visualisation/map_test.clj
```diff
@@ -0,0 +1,177 @@
...
+      (jdbc/db-do-commands *tenant-conn* [drop-table-seq
+                                          drop-table
+                                          drop-imported-table-seq
+                                          drop-imported-table])
+      (dataset/delete *tenant-conn* *dataset-id*))))
+
+
+(use-fixtures :once
+  system-fixture data-groups-future-fixture tenant-conn-fixture
+  error-tracker-fixture tu/spec-instrument dataset-data-fixture
```
iperdomo, line 72: I'm interested to hear why we're using *fixtures* for setting up the test? 🤔 
Other functional tests prepare the dataset/visualization/etc as part of the test and *not* in a fixture. Are we going to reuse them? If so, is not the `akvo.lumen.fixtures` the place for this code?

backend/test/akvo/lumen/visualisation/map_test.clj
```diff
@@ -0,0 +1,177 @@
...
+(defmulti verify-val (fn [_ f]
+                       (cond
+                         (inst? f) :inst
+                         (map? f) :geopoint ;; a bit simplistic :-)
+                         :else :default)))
+
+(defmethod verify-val :inst [v f]
+  ;; Since we convert datetime with timezone to text in the mapconfig sql
+  (let [custom-formatter (f/formatter "yyyy-MM-dd HH:mm:ssZ")]
+    (is (= (inst-ms (.toInstant (f/parse custom-formatter v)))
```
iperdomo, line 136: I think it should be possible to use the JDK api [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) for this checks instead of using an external library.

iperdomo (changes requested):

backend/test/akvo/lumen/visualisation/map_test.clj
kardan: The thought was to prepare the dataset, write on test for existing maps, and then clean up the data after the test is done. This could be handled without the fixture but since I knew we wanted to write a test to verify the functionality between current maps and the future implementation of dataset-version-2 maps I put it in a fixture so we can do another deftest but use the same data. 

This said, maybe fixtures should live in fixtures.clj. I was not totally sure about the approach and kept the implementation as isolated as possible.

iperdomo: > The thought was to prepare the dataset, write on test for existing maps, and then clean up the data after the test is done. 

The cleanup is done by the *outer* database connection and system fixture, right?

I also don't have a strong opinion but is *different* than the rest of functional tests

iperdomo: It seems to be that the joggling with those [DDL](https://www.postgresql.org/docs/9.5/ddl.html) is already handled by migrating/rollback approach

backend/test/akvo/lumen/visualisation/map_test.clj
tangrammer: value `'2018-10-01 00:00:00+00'` received doesn't compliant with Instant/parse expectations so we're keeping current parse approach here

PS: value should has `T` instead of blank space

tangrammer (approved):

PR #3084, opened 2020-12-28
Add data-group support to update dataset logic 

tangrammer:
- [ ] **Update release notes if necessary**

Add test flow `uat1/_ _Lumen tests/Lumen Testing/2 Lumen RQG Testing/New form`

```clojure
(tu/read-edn-flow-dataset "uat1" "638889127" "638879132")
```

Screenshot 2021-01-07 at 12 54 14

PR #3085, opened 2020-12-31
Utils to write/read flow dataset to/from edn file

tangrammer:
So we can use these edn files to import/update flow dataset in dev/tes envs without connecting akvo-flow-api

Problems to solve:
- [X] read edn with [tagged literals](https://clojure.org/reference/reader#tagged_literals) eg: java.util.Instant
- [ ] added support for lumen geo types
- [x] always output to a dev directory ignored(included in gitignore)

## Used this uat1 form to generate edn files
Screenshot 2020-12-31 at 18 03 53

## Files generated:
Screenshot 2021-01-04 at 12 25 12

## Dataset generated using these edn files
Screenshot 2020-12-31 at 18 03 15


backend/test/akvo/lumen/test_utils.clj
```diff
@@ -92,7 +92,13 @@
   (let [spec {"source" (with-meta {"kind" kind
                                    "hasColumnHeaders" (boolean has-column-headers?)}
                          {:data data})}
-        [tag {:strs [updateId] :as res}] (update/update-dataset tenant-conn caddisfly {} error-tracker dataset-id data-source-id spec)]
+        [tag {:strs [updateId] :as res}] (update/update-dataset tenant-conn caddisfly
```
tangrammer, line 95: this change is related with #3084  to data-groups update logic

backend/dev/src/dev.clj
```diff
@@ -24,7 +26,10 @@
             [integrant.core :as ig]
             [integrant.repl :as ir]
             [integrant.repl.state :as state :refer (system)])
-  (:import [org.postgresql.util PSQLException PGobject]))
+  (:import [org.postgresql.util PSQLException PGobject]
+           [java.time Instant]))
+
+(def dev-flow-datasets-dir "flow-datasets")
```
kardan, line 32: I would rather remove this indirection, why not have it inline?

kardan:
Should we add a function in dev that have a hardcoded reference to the filename? Then it would be easy to import as part of CI or just in general if you don't want to eval code but play with Lumen and Flow data locally?

kardan (approved):
Wonderful!

Issue #3092, opened 2021-01-12
Capitalized group names affects to derive-js transformation 


tangrammer:
# Context
right now we capitalize group names by default in dataset view data-group
![Screenshot 2021-01-12 at 10 52 44](https://user-images.githubusercontent.com/731829/104298490-53221d80-54c4-11eb-8b92-ee419ea090ee.png)

# Problem or idea

trying js transformation `row["Non repeatable"]["Family name"]+" welcome"`  **FAILS**
but `row["Non Repeatable"]["Family name"]+" welcome"`  **SUCCEED**

![Screenshot 2021-01-12 at 10 53 51](https://user-images.githubusercontent.com/731829/104298778-ad22e300-54c4-11eb-97be-8d075b8aa981.png)
![Screenshot 2021-01-12 at 10 53 57](https://user-images.githubusercontent.com/731829/104298789-af853d00-54c4-11eb-881a-24968c24ebe5.png)

# Solution or next step
How could we solve it?

tangrammer:
@janagombitova @kardan should we adapt the backend to be case insensitive or just don't change letter case in frontend

PS: changing behaviour in client is just removing css class, backend changes should have more changes and tests and perhaps more unknown consequences (thinking in undo with existent datasets with 2 groups having different letter case)

tangrammer:
Relates https://github.com/akvo/akvo-lumen/blob/de85cd70ad8dcd2c60e9de97d00f479ace9a4806/client/src/components/dataset/sidebars/GroupsList.scss#L21

janagombitova:
@tangrammer I would say we show things are they are. So do not change letter case in the frontend. It is less work and the users see the data, titles, etc as it is put in.

kardan:
Agree

tangrammer:
finally we need to adapt "Metadata", "Main" and "Transformations" in backend side thus I don't think we want to show these values in lower case 🤔

janagombitova:
Agree.

PR #3093, opened 2021-01-12
[#3092] Adapt backend to uppercase default groups

tangrammer:
so backend could work with:  
`Main` or `main`
`Transformations` or `transformations`
`Metadata` or `metadata`


kardan (approved):
LGTM

Issue #3031, opened 2020-11-05
Adapt admin scripts to new Google SQL


tangrammer:
# Context
Relates to #2855 & a new request to add a tenant.

# Problem or idea
We can't create new instances after the switch to Google SQL due to differences in how we connect and role permissions.

# Solution or next step
Use kubectl to call current scripts but we need to adapt some config details

PR #3095, opened 2021-01-21
[#3031] New instance bash

tangrammer:

### DIFF
--- a/ci/new-instance.sh
+++ b/ci/new-instance.sh
@@ -15,7 +15,7 @@ fi
 
 log "CLUSTER: $CLUSTER"
 
-read -r -p "lumen url:" URL
+read -r -p "lumen url (should include https://):" URL
 
 if [ -z "${URL}" ]; then
     log "Nothing typed"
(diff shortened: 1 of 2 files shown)


ci/new-instance.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+
+log "URL: $URL"
+log "ORGANISATION $ORGANISATION"
+log "EMAIL $EMAIL"
+
+gcloud container clusters get-credentials "$CLUSTER" --zone europe-west1-d --project akvo-lumen
+
+export POD_NAME=$(kubectl get pods -l "run=lumen-blue-green,color=$(./live-color.sh)" -o jsonpath="{.items[0].metadata.name}")
```
dlebrero, line 46: export is unnecessary.

ci/new-instance.sh
```diff
@@ -0,0 +1,50 @@
...
+
+log "POD $POD_NAME"
+
+kubectl exec "$POD_NAME" -c lumen-backend -- java -cp akvo-lumen.jar clojure.main -m akvo.lumen.admin.add-tenant "$URL" "$$ORGANISATION" "$EMAIL"
```
dlebrero, line 50: $$ORGANIZATION <-- is it $ or $$?

ci/new-instance.sh
```diff
@@ -0,0 +1,50 @@
...
+
+read -r -p "user email:" EMAIL
+
+if [ -z "${EMAIL}" ]; then
+    log "Nothing typed"
+    exit 2
+fi
+
+
+log "URL: $URL"
```
dlebrero, line 40: Do you want to validate the url? I see in the doc that there are some requirements. 

Or maybe at least put that requirement in the "read URL" in line 18

ci/new-instance.sh
dlebrero, line 46: The "./live-color.sh" path means that the expectation is that the script is run from within the `ci` directory. 

Either:
1. Add this to the documentation.
2. Check that the POD_NAME is not empty
3. Avoid the issue as per https://github.com/akvo/akvo-lumen/blob/4b1ac11ac88c8869fa189f8fcd141e1b84843333/ci/flip-blue-green-deployment.sh#L35

dlebrero (changes requested):
The new tenant.md documentation should be updated to point to this file.

ci/new-instance.sh
tangrammer: included

ci/new-instance.sh
tangrammer: added requirement in the read URL step

ci/new-instance.sh
tangrammer: fixed

ci/new-instance.sh
tangrammer: removed

PR #3122, opened 2021-05-24
Add data-groups merge-dataset support

tangrammer:

### DIFF
--- a/backend/src/akvo/lumen/lib/data_group.clj
+++ b/backend/src/akvo/lumen/lib/data_group.clj
@@ -26,7 +26,8 @@
                                                  columns))}))
                      []
                      data-groups)]
-    {:select (mapv :columnName (flatten (map :columns adapted-dgs)))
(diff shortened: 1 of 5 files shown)


backend/src/akvo/lumen/lib/data_group.clj
```diff
@@ -26,7 +26,8 @@
                                                  columns))}))
                      []
                      data-groups)]
-    {:select (mapv :columnName (flatten (map :columns adapted-dgs)))
+    {:select (cond-> (mapv :columnName (flatten (map :columns adapted-dgs)))
+               (= 1 (count data-groups)) (conj "rnum"))
```
tangrammer, line 30: only csv datasets have only 1 data-group

tangrammer: not really, if you have a csv dataset with a merged data-group

Issue #3125, opened 2021-06-02
Configure JVM heap dump collection


iperdomo:
# Context
We need to debug recent JVM OutOfMemory errors

# Problem or idea

We're unable to debug the problem as we don't have access to the heap.

# Solution or next step

Use the same approach as we do in Flow API. Configure the JVM to
generate a heap dump and upload it to a known bucket.

PR #3128, opened 2021-06-03
[#3125] Increase JVM heap to 75% and enable UseContainerSupport

kardan:


backend/run-server.sh
```diff
@@ -16,4 +16,6 @@ java -XshowSettings:vm \
      -XX:+UseG1GC \
      -XX:+HeapDumpOnOutOfMemoryError \
      -XX:HeapDumpPath="/dumps/${dump_filename}" \
+     -XX:+UseContainerSupport \
```
iperdomo, line 19: UseContainerSupport is *true* by default

iperdomo (changes requested):
Let's remove `UseContainerSupport` as is `true` by default

iperdomo (approved):

PR #3129, opened 2021-06-09
fixed color picker positioning

melowsoft:

- fixed the colour picker not appearing in the right place

### DIFF
--- a/client/src/components/common/ColorPicker.jsx
+++ b/client/src/components/common/ColorPicker.jsx
@@ -9,7 +9,7 @@ import LegendShape from '../charts/LegendShape';
 const ColorPicker = ({ title, left = 0, top = 0, placement, style, hideArrow, ...rest }) => (
   <Popover
     left={left}
-    top={top}
+    top={0}
     placement={placement}
     hideArrow={hideArrow}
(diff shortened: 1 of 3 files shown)


client/src/components/common/ColorPicker.jsx
```diff
@@ -9,7 +9,7 @@ import LegendShape from '../charts/LegendShape';
 const ColorPicker = ({ title, left = 0, top = 0, placement, style, hideArrow, ...rest }) => (
   <Popover
     left={left}
-    top={top}
+    top={0}
```
kardan, line 12: This component is used for several types of charts (see https://github.com/akvo/akvo-lumen/search?q=ColorPicker&type=code) I would guess this change will be a bit off for some chart types, or was that taken into consideration and it all works fine?

melowsoft: Okay, I would effect it, just for the simple bar chart and see

melowsoft: I have made the change just for the Simple bar chart component

kardan: Cool!

PR #3137, opened 2021-06-17
Issue/2794 dashboard visualization error

melowsoft:

- Added Error screen for dashbaord error
- Fixed undefined error

### DIFF
--- a/client/src/actions/collection.js
+++ b/client/src/actions/collection.js
@@ -85,10 +85,10 @@ export function addEntitiesToCollection(entityIds, collectionId) {
     dispatch(showNotification('info', `Added to ${collection.title}`, true));
   };
 }
-export function addEntityToCollection(entityId, collectionId) {
+export function addDatasetToCollection(datasetId, collectionId) {
   return (dispatch, getState) => {
       const collection = getState().library.collections[collectionId];
-      collection.datasets.push(entityId)
+      collection.datasets.push(datasetId)
       dispatch(editCollection(collection));
       dispatch(showNotification('info', `Added to ${collection.title}`, true));
   };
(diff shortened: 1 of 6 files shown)


client/src/actions/collection.js
```diff
@@ -73,16 +73,29 @@ export function deleteCollection(id) {
 export function addEntitiesToCollection(entityIds, collectionId) {
   return (dispatch, getState) => {
     const collection = getState().library.collections[collectionId];
+   
+    console.log(entityIds, collection, "for need purpose")
     const visualisations = uniq([...entityIds.visualisations, ...collection.visualisations]);
+    console.log( "i got here finally, where the bug is")
```
iperdomo, line 79: Please remove those debugging logging

iperdomo (changes requested):
I think this branch has a mixture of 2 different problems:

* Adding a dataset within an existing collection
* Showing an error message when the dashboard doesn't exist

Is this case? 🤔 If so, let's split the changes in 2 branches, one for the dataset problem, other for the visualization error

iperdomo:
I would also suggest to fix your committer settings, right now is using a non recognized email address.

https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address

iperdomo:
@melowsoft this branch still has changes that are *not* related to the issue. Can you update/fix this branch so we have only the necessary changes for the issue at hand?

melowsoft:
> @melowsoft this branch still has changes that are _not_ related to the issue. Can you update/fix this branch so we have only the necessary changes for the issue at hand?

okay, I will fix this

PR #3142, opened 2021-06-21
Issue/2539 creating visualization and dashboard

melowsoft:

- Fixed creating visualisation on collection
- Fixed creating dashboard on collection

### DIFF
--- a/client/src/actions/collection.js
+++ b/client/src/actions/collection.js
@@ -94,6 +94,23 @@ export function addDatasetToCollection(datasetId, collectionId) {
   };
 }
 
+export function addVisualisationToCollection(datasetId, collectionId) {
+  return (dispatch, getState) => {
+    const collection = getState().library.collections[collectionId];
+    collection.visualisations.push(datasetId);
+    dispatch(editCollection(collection));
+    dispatch(showNotification('info', `Added to ${collection.title}`, true));
+  };
+}
+export function addDashboardToCollection(datasetId, collectionId) {
+  return (dispatch, getState) => {
+    const collection = getState().library.collections[collectionId];
+    collection.dashboards.push(datasetId);
+    dispatch(editCollection(collection));
+    dispatch(showNotification('info', `Added to ${collection.title}`, true));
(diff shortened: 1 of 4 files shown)


client/src/actions/collection.js
```diff
@@ -94,6 +94,23 @@ export function addDatasetToCollection(datasetId, collectionId) {
...
 
+export function addVisualisationToCollection(datasetId, collectionId) {
+  return (dispatch, getState) => {
+    const collection = getState().library.collections[collectionId];
+    collection.visualisations.push(datasetId);
+    dispatch(editCollection(collection));
+    dispatch(showNotification('info', `Added to ${collection.title}`, true));
+  };
+}
+export function addDashboardToCollection(datasetId, collectionId) {
```
iperdomo, line 105: change parameter name to `dashboardId`

client/src/actions/collection.js
```diff
@@ -94,6 +94,23 @@ export function addDatasetToCollection(datasetId, collectionId) {
   };
 }
 
+export function addVisualisationToCollection(datasetId, collectionId) {
```
iperdomo, line 97: change parameter name to `visualisationId`

client/src/actions/collection.js
iperdomo: now that we have 3 copies of the same function, the only part that changes is
```
collection.<entityType>.push(entityId)
```
Does it make sense to refactor that and create a single function:

```js
addEntityToCollection(entityType, entityId, collectionId) {
    const collection = getState().library.collections[collectionId];
    collection[entityType + "s"].push(datasetId);
    dispatch(editCollection(collection));
    dispatch(showNotification('info', `Added to ${collection.title}`, true));
  };
}
```

When using it

```js
dispatch(addEntityToCollection("dataset", dataset.id, collectionId))

iperdomo (changes requested):
What do you think on refactoring that copied code and use only one function?

iperdomo (approved):

PR #3144, opened 2021-06-23
Issue/2539 nav return to collection creating visualisation

melowsoft:

- Back button navigation on creation of visualisation/dashboard should return user to previous collection


iperdomo:
@janagombitova do we need to show that the visualization was added to a collection? It was a *surprise* to me. See the video.

https://user-images.githubusercontent.com/178474/123138382-3ca9c600-d455-11eb-9f0e-18f98d0918da.mp4

melowsoft:
It’s actually a feature we didn’t have before, so I am not sure if we need it or not

janagombitova:
@iperdomo and @melowsoft this is ok. It can remain as it is. We actually have that notification in production already (from day1). When a user, while being in the Library, adds a visualisation/dataset/dashboard/combination to a collection we show the notification. With this implementation, the user gets informed about something happening behind the scenes, what is a good message to show.

iperdomo (approved):

PR #3149, opened 2021-06-30
[3148] Fix: Removed dataset walkthrough

melowsoft:

- Removed Dataset work through


iperdomo:
If we're not using IntroJS anymore, let's remove it from the dependencies
https://github.com/akvo/akvo-lumen/blob/e110d59b53ece8ff8f9e3220922a24170423fb7a/client/package.json#L79

melowsoft:
Screenshot 2021-06-30 at 18 51 40
Jana said we should keep the library for the next time we need to inform people users about new features in Lumen

iperdomo:
> We should keep the library and underlying implementation for the next time when we need to inform about a new feature in Lumen.

We have a Git history of all changes that have happened in the repository. If we need to do some *walkthrough* in the future, the code is there, in the history. Keeping libraries that we don't use in the list of dependencies make little sense.

melowsoft:
okay, let me take it out

valllllll2000 (approved):

iperdomo (approved):

